Issue #218, opened 2012-05-19
hmset broken


525c1e21-bd67-4735-ac99-b4b0e5262290:
When passed an object of arrays, `HMSET` screws up in strange ways...

``` JavaScript
  var client, redis;

  redis = require('redis');

  client = redis.createClient();

  client.hmset('foo', {
    a: [0, 0, 0]
  });

  client.hmset('foo', {
    a: [0, 0, 0],
    b: [0, 0, 0]
  });

  client.hmset('foo', {
    a: [0, 0, 0],
    b: [0, 0, 0],
    c: 'bar'
  });
```

`redis-cli monitor`

> 1337435337.689485 "hmset" "foo" "a" "0" "0" "0"
> 1337435337.689667 "hmset" "foo" "a" "0,0,0" "b" "0" "0" "0"
> 1337435337.689744 "hmset" "foo" "a" "0,0,0" "b" "0,0,0" "c" "bar"

N.B. It only breaks if the last key of the object is an array.

This was a real PITA to find. I presume something is broken somewhere deep inside all the voodoo of RedisClient.prototype.hmset (https://github.com/mranney/node_redis/blob/9e76387f59b12e2f3590687ac4aba37e0247305a/index.js#L879)

BridgeAR:
This is not supported and returns a warning by now. If you want to provided arrays as keys, please convert them to a string before passing them to the node_redis client.

PR #276, opened 2012-09-15
Consolidate special hash helpers across client + multi client.

### LINKED ISSUE #272: multi(): Update to 0.7.3 changes hmget, hdel command format
Passing an array as second parameter to hmget() and hdel() results in the desired output with version 0.7.2, but just returns one result in 0.7.3. When i concat the array the command's list it works fine in both versions. This is inconsistent with plain `hmget()` and `hdel()`

    var async = require('async');
    var redis = require('redis');
    
    var client;
    async.waterfall([
      function(cb) {
        client = redis.createClient(6379, 'localhost');
    
        client.hmset('hash', 'key1', 1, 'key2', 2, cb);
      },
      function(res, cb) {
        client.multi([['hmget', 'hash', ['key1', 'key2']]]).exec(cb);
      },
      function(result, cb) {
        console.log(result);
        // Output with 0.7.2:
        // [ [ '1', '2' ] ]
        // Output with 0.7.3:
        // [ [ null ] ]
        client.end();
        cb();
      }
    ]);

TheDeveloper:
There were many separate hash methods implemented separately for client and multi, but they were doing very similar things. I have gone through and combined them into methods shared between the normal and multi clients, to ensure consistency, improve maintainability and decrease bug susceptibility.

* Addresses #218: Allow strings and integer in obj for hmset(key, obj), but throw error when given object type
* Fixes #272: Accept array arg consistently across client and multi
* Supplemented with tests and whole suite passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,7 +371,7 @@ channel name as `channel` and the new count of subscriptions for this client as 
 
 ## client.multi([commands])
 
-`MULTI` commands are queued up until an `` is issued, and then all commands are run atomically by
+`MULTI` commands are queued up until an `EXEC` is issued, and then all commands are run atomically by
 Redis.  The interface in `node_redis` is to return an individual `Multi` object by calling `client.multi()`.
 
 ```js
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ Output:
 Multiple values in a hash can be set by supplying an object:
 
     client.HMSET(key2, {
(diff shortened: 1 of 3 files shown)


bcoe:
@TheDeveloper this ticket is pretty ancient, and unfortunately does not merge cleanly. I'm closing it as part of a cleanup we're taking on. Please feel free to rebase and reopen.

TheDeveloper:
@bcoe no problem, it was never fully finished anyway

PR #284, opened 2012-09-19
Fix for subscribing to large messages

macdiesel:
Using pubsub when you are subscribed to a channel and a large message is delivered in multiple packets the message is incorrectly handled. Only the first packet in the stream for the message is parsed.  This patch demonstrates a fix for the issue.


BridgeAR:
@azafty468 @macdiesel is this still an issue as of today? Would you be so kind and check and otherwise try to write a failing test?

azafty468:
Sorry, I rebased this thing once and have long since moved past this.  No clue if this is or is not still an issue, but I'm well past any support here.

BridgeAR:
@azafty468 That sad, but I totally understood your frustration :/

PR #318, opened 2012-10-27
Revert eval caching

frolic:
This reverts my previous attempt at automatically caching `eval` scripts. The current implementation does not support calling with an array of arguments (e.g. `eval(args, callback)`). And after much internal struggle, I decided this automagic probably isn't best suited in `eval`.

I may re-implement this at a later date in some other form and/or under another method name.

@DTrejo I tried to `git revert ...`, but conflicts wouldn't let it happen gracefully, so this is the best I can do. There were no tests or docs added specifically for this change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,6 @@ var net = require("net"),
     Queue = require("./lib/queue"),
     to_array = require("./lib/to_array"),
     events = require("events"),
-    crypto = require("crypto"),
     parsers = [], commands,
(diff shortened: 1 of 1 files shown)


BridgeAR:
@stuartpb thx for pointing this out. node_redis auto pipelines all commands. So I removed the eval caching again as the order of execution was violated and I added some tests that would break otherwise.

PR #338, opened 2012-11-23
call toString() method when the type of the value is number in hmset

luin:
When the second parameter is an object in HMSET, I think it's will be better to convert the value from number to string automatically by call ```toString()```method.

For example, the following code should be worked:

```javascript
client.hmset("key", {
  name: "Tom",
  order: 5
});
```


BridgeAR:
Closing as it seems to be already fixed

PR #347, opened 2012-12-17
Command queue error after pubsub

cxreg:
Don't throw a command queue error if a pubsub message is processed after exiting pubsub mode.  This can happen due to a race and is not a bug, just ignore the message.

#### Ready to merge checklist
- [ ] test(s) in test.js
- [ ] tests will fail without the PR, but succeed once applied
- [ ] merges cleanly
- [x] coding style (4-space indents, looks similar to other code)

### DIFF
--- a/index.js
+++ b/index.js
@@ -628,6 +628,8 @@ RedisClient.prototype.return_reply = function (reply) {
             return elem.replace(/\\"/g, '"');
         });
         this.emit("monitor", timestamp, args);
+    } else if (type === "message" || type === "pmessage") {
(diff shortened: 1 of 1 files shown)


BridgeAR:
@cxreg @humbroll could you try to write a test for this? I would be happy if it fires at least one out of ten times so I'm able to reproduce it at all.

BridgeAR:
This is likely fixed by 837cec36b642cf6ca5eace93bb59378d1c3e477e

@cxreg @humbroll could you verify this? :)

PR #354, opened 2012-12-21
Make return_buffers work with channel names.

saurik:
When using pub/sub, return_buffers is ignored and channel names are forced to be strings, making it impossible to use binary channel names when accessing redis using this library.
It turns out the .toString() on these names is superfluous as the underlying parser is already returning these correctly as either strings or node.js Buffer objects based on the return_buffer setting.

(I am fully willing to believe, however, that a more complex code change is needed to work with different parser backends, or in more circumstances; I know very little about this codebase in general.)

#### Ready to merge checklist
- [ ] test(s) in test.js
- [ ] tests will fail without the PR, but succeed once applied
- [x] docs, if applicable
- [ ] merges cleanly
- [x] coding style (4-space indents, looks similar to other code)

### DIFF
--- a/index.js
+++ b/index.js
@@ -596,9 +596,9 @@ RedisClient.prototype.return_reply = function (reply) {
             type = reply[0].toString();
 
             if (type === "message") {
-                this.emit("message", reply[1].toString(), reply[2]); // channel, message
+                this.emit("message", reply[1], reply[2]); // channel, message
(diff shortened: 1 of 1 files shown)


BridgeAR:
Closed by 55e4a9b847b15f535d1538241c98554de7c14af8

PR #362, opened 2013-01-09
Enough reconnection retry before flushing out offline queue

vedratna-velani:
I came across the scenario where under heavy traffic or due to some connection issue, got redis connection error and redis command just fails with error. Instead I think it should have an option to retry enough before error out the command.


BridgeAR:
Thx a lot for your work but closing this in favor of #829.

PR #384, opened 2013-02-25
Fix for subscribing to large messages

azafty468:
Includes fixes from macdiesel and myself to fix large message
subscription.  Rebased to latest copy (as of 2/23/2013) version of
node_redis.  This is an updated pull request of pull 284.


BridgeAR:
There were a couple of parser fixes since Feb 2013 and no one has ever reported something like that again. The changes as they are in here can't run properly right now, as the try catch is still mandatory. I'll also refactor the pub sub system soon and if I find any issues with large messages I'll have a closer look at this.

PR #428, opened 2013-04-30
Export utils and commands to share (with sentinel client, or others)

benbuckman:
This supports [redis-sentinel-client](https://github.com/DocuSignDev/node-redis-sentinel-client). It should not change or break any node_redis functionality, it simply separates and/or exports pieces that can be used by other modules. (Which should be good architecture regardless of the sentinel client or other modules using it.)

Thanks!


BridgeAR:
Ping @benbuckman

PR #438, opened 2013-05-22
Oscillating reconnection intervals when retry_max_delay option set.

TedTrela:
When client option retry_max_delay set, retry_delay value oscillates
between retry_max_delay and retry_max_delay * retry_backoff.  Caused
inconsistent reconnection attempt intervals.

Adjusted test to not test whether connection actually occurs, but rather that client.retry_delay does not exceed option value.


BridgeAR:
Thx a lot for your work! Closing this as #829 fixed this already.

PR #456, opened 2013-06-23
[fix] remove try catch wrapping around the callback thats passed in by the module user

### LINKED ISSUE #453: Unexpected error thrown from `try_callback`
So [this](https://github.com/mranney/node_redis/blob/master/index.js#L557-L565) piece of code seems to be causing me issues. I return a callback with an error in my code for informational purposes if the hget returns null. This little try catch ends up unexpectedly crashing my code because of it. 

So my main question is why does this code need to exist? I feel like its not a good practice to wrap other people's code in a try catch (but this is just my opinion). Thoughts @DTrejo?

jcrugzz:
fixes #453. I don't see any reason to wrap the user's code in a try
catch as it can have unexpected consequences.

### DIFF
--- a/index.js
+++ b/index.js
@@ -541,18 +541,6 @@ RedisClient.prototype.return_error = function (err) {
     }
 };
 
-// if a callback throws an exception, re-throw it on a new stack so the parser can keep going.
-// put this try/catch in its own function because V8 doesn't optimize this well yet.
-function try_callback(callback, reply) {
-    try {
-        callback(null, reply);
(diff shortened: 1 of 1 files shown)


BridgeAR:
Oh, I forgot about this one and opened #816. If you rebase I'd merge this instead of mine (after we all agreed on removing it overall).

jcrugzz:
@BridgeAR no worries, you can go ahead with your PR since this one has aged and you added some other things :).

PR #457, opened 2013-06-25
Fix TypeError when handling multi exception

kwangsu61:
When errors occurs using multi/exec, avoid "TypeError: Cannot read property 'length' of undefined"


lamujeresponja:
I have found this error again in node redis 0.12

brycebaril:
@lamujeresponja -- can you reproduce the error, and if so can you provide a code snippet that does? Thanks!

PR #506, opened 2013-10-30
fix reconnection error when call client.end()

numbcoder:
When the client is waiting for reconnecting, and then call `client.end()`, it will throw a error.


BridgeAR:
This is already fixed => closed

Issue #510, opened 2013-11-14
node killed when unsubscribe all.


humbroll:
Node process is killed when I unsubscribe ALL. 
Below is code. 
```
socket.on('subsrcribe_car_ids', function(data){    
    redis_sub.unsubscribe(); // <<<<<---- process killed point! 

    for( var i = 0 ; i < data.length ; i++ ){
      car_id = data[i];
      redis_sub.subscribe(car_id);
    }
  })
```

If I assume that the unsubscribe method is working as asynchronous, and I call subscribe sequentially after unsubscribe(), crash will occur.

Any idea about it? 

Error messages are below 

```
events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: node_redis command queue state error. If you can reproduce this, please report it.
    at RedisClient.return_reply (/Users/humbroll/workspace/pangyo/stream/node_modules/redis/index.js:714:15)
    at HiredisReplyParser.<anonymous> (/Users/humbroll/workspace/pangyo/stream/node_modules/redis/index.js:312:14)
    at HiredisReplyParser.EventEmitter.emit (events.js:95:17)
    at HiredisReplyParser.execute (/Users/humbroll/workspace/pangyo/stream/node_modules/redis/lib/parser/hiredis.js:43:18)
    at RedisClient.on_data (/Users/humbroll/workspace/pangyo/stream/node_modules/redis/index.js:535:27)
    at Socket.<anonymous> (/Users/humbroll/workspace/pangyo/stream/node_modules/redis/index.js:91:14)
    at Socket.EventEmitter.emit (events.js:95:17)
    at Socket.<anonymous> (_stream_readable.js:746:14)
    at Socket.EventEmitter.emit (events.js:92:17)
    at emitReadable_ (_stream_readable.js:408:10)
```

BridgeAR:
Duplicate of #577. The reason that this crashes is that if you're subscribed to more than one channel the response is going to be not a single response but one for each channel. Right now the behavior is 1 command => 1 response. That is not true in this case so it breaks while unsubscribing all.

PR #511, opened 2013-11-14
Update index.js

humbroll:
for resolving #510


BridgeAR:
@humbroll would you be so kind and add a test?

BridgeAR:
@humbroll the variable queue_len does not exist in that function? And the issue is that the reply for one command are n replies. So that issue can't be fixed by setting pub_sub_mode to true.

PR #535, opened 2013-12-25
Refactoring for slight performance improvement

k-ori:
Replaced some ECMAScript5 Array method with anonymous function as described in TODO. A few method calls still stay to keep scope clean at the top level. Also refactored places to call Array.length for performance improvement.


BridgeAR:
@k-ori are you till up for the request?

k-ori:
Performance improvement is slight. I'm closing this PR for now.

PR #538, opened 2014-01-09
Add code property to redis errors

skeggse:
The [redis protocol](http://redis.io/topics/protocol) implies all error responses will begin with an all-caps error code, which could be included in the `code` property of the parsed error object.

```js
function require_error(label, code) {
  return function (err, results) {
    assert.notEqual(err, null, label + " err is null, but an error is expected here.");
    assert.strictEqual(err.code, code, label + " expected err with code " + code + ", got code: " + err.code);
    return true;
  };
}

client.evalsha("ffffffffffffffffffffffffffffffffffffffff", 0, require_error("EVALSHA", "NOSCRIPT"));
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -302,7 +302,15 @@ RedisClient.prototype.init_parser = function () {
         if (reply instanceof Error) {
             self.return_error(reply);
         } else {
-            self.return_error(new Error(reply));
+            var err, match = /^([A-Z]+)\s+(.+)$/.exec(reply);
+            if (match) {
+                err = new Error(match[2]);
+                err.code = match[1];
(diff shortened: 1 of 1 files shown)


BridgeAR:
@skeggse what is the benefit of adding the .code? Does it make any difference for you if you just test the error message in your code for the error code instead?

skeggse:
It's just for convenience. It doesn't degrade performance in any meaningful way - this kind of exception path should generally not be optimized.

BridgeAR:
@skeggse would you be so kind and rebase this and add some tests?

PR #546, opened 2014-02-03
update copyright year

gdi2290:


BridgeAR:
As this is already outdated again and there's a new merge request that includes the correct year as well as other small changes I'll close this one.

PR #560, opened 2014-02-26
fix #537 hmset do not save null field

### LINKED ISSUE #537: hmset with object should not save NULL and `undefined` field value
We offen make an mistake that save null value to 'null' string and undefined value to 'undefined' string.

We have 2 choice:

1. save empty '' string as the value
2. don't save anything. (but this will also make a mistake when we want to remove a field value from an exist hash.

guileen:
We offen make an mistake that save null value to 'null' string and undefined value to 'undefined' string.

If you're trying to remove a field value from an existing hash you should likely be using HDEL anyway.

see #537.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1039,6 +1039,7 @@ RedisClient.prototype.hmset = function (args, callback) {
         tmp_keys = Object.keys(args[1]);
         for (i = 0, il = tmp_keys.length; i < il ; i++) {
             key = tmp_keys[i];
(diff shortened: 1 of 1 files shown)


BridgeAR:
Thanks for this PR but to fix this I prefer to use a more strict behavior. From 2.5 on node_redis is going to  emit and log a deprecation message in such a case and from 3.0 on an error will be returned.

PR #572, opened 2014-04-12
Ignore errors after .end() is called instead of causing an uncaught exception

crispy1989:
Currently, it is possible for an error to occur on the stream after .end() is called, which clears event listeners.  This results in an uncaught exception.  In particular, if a connection is pending while .end() is called, then fails (with something like ECONNREFUSED), it triggers an uncaught exception.  This commit causes it to ignore errors after .end() is called.

### DIFF
--- a/index.js
+++ b/index.js
@@ -318,6 +318,7 @@ RedisClient.prototype.on_ready = function () {
     var self = this;
 
     this.ready = true;
+    this.ended = false;
 
     if (this.old_state !== null) {
         this.monitoring = this.old_state.monitoring;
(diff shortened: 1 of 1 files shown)


BridgeAR:
@crispy1989 could you please add a test for this?

BridgeAR:
Seems like this issue was at least partly fixed. We should check if this fix is actually better than the one in place.

crispy1989:
The original issue this pull request fixes was fixed by 1c983a11 and c7dbfec8.  However, it looks like these may actually introduce another issue where callbacks for pending requests may not be called on end (flush_and_error isn't called).

PR #610, opened 2014-06-23
Improving server_info

SinisterLight:
Improving server_info object by parsing the db* keys of redis-info.
earlier it was
server_info = { .. 
 db0: 'keys=1,expires=0,avg_ttl=0',
  db9: 'keys=49,expires=49,avg_ttl=2588996447',
.. }
changed this to
server_info = { ..
  db0: { keys: '1', expires: '0', avg_ttl: '0' },
  db9: { keys: '49', expires: '49', avg_ttl: '2588996447' }
..}

### DIFF
--- a/index.js
+++ b/index.js
@@ -362,7 +362,7 @@ RedisClient.prototype.on_ready = function () {
 };
 
 RedisClient.prototype.on_info_cmd = function (err, res) {
-    var self = this, obj = {}, lines, retry_time;
+    var self = this, obj = {}, lines, retry_time, sub_lines;
 
     if (err) {
         return self.emit("error", new Error("Ready check failed: " + err.message));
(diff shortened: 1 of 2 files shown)


BridgeAR:
@brycebaril I doubt that this is going to impact the performance in a measurable way as there are only very few lines that will be parsed in the special way (most of the time it'll just be one). It might be a very tiny speed improvement to only iterate over the keyspace part e.g.

```js
var i = 0;
var key = 'db' + i;
while (obj[key]) {
  var sublines = obj[key].split(',');
  // do stuff with the keyspace ...
  i++;
  key = 'db' + i;
}
```

And we could improve the performance of the info call further by removing the .forEach. That of course, is independent from this merge request.
So I'd say this is fine if it's going to be rebased, but I do not have a strong opinion about this.

BridgeAR:
Git got fixed by #855 

Thx @SinisterLight

PR #626, opened 2014-07-13
Refactor to_array

alexanderGugel:
See http://debuggable.com/posts/turning-javascript-s-arguments-object-into-an-array:4ac50ef8-3bd0-4a2d-8c2e-535ccbdd56cb for more information on this. This is more elegant and maybe even faster. For sure it makes more sense than iterating manually over args.


BridgeAR:
As already mentioned this is a optimization killer and should be closed.

PR #633, opened 2014-07-23
Allow EVAL KEYS and ARGS to be passed as an array/ Parse ZRANGE WITHSCORES as object (FIXES #97)

### LINKED ISSUE #97: withscores
Hi!

Wonder if you could consider whether it's feasible to support commands which take `withscores` option to return hashes instead of plain arrays.

TIA,
--Vladimir

cPu1:
`redis.eval` requires that the client know the KEYS and ARGS that will be passed to the Lua script.
I am using Lua scripts that manipulate a variable number of KEYS and have to use the following code:

`var args = [script, keys.length].concat(keys);`
`redis.eval(args, function (err, res) {
  //
});
`

redis.eval should allow calls of the forms (callback optional):
`redis.eval(script, keys, args);`
`redis.eval(script, keys);`
`redis.eval(script); //no args`

To give a concrete example, this function increments the KEYS passed to the script and returns the new values in an array. 

```
function incrementAgentCount (groupIds, domain) {
    var groupKeys = groupIds.map(function (groupId) {
        return groupId + '@' + domain;
    }),
        script = "local updated = {};for i = 1, #KEYS do  local k = KEYS[i];  updated[i] = redis.call('hincrby', k, 'agents:online', 1); 
          end;return updated;",
        args = [script, groupKeys.length].concat(groupKeys);

    mRedis.eval(args, function (err, res) {
        //res 
    });

}
```

Although this can be done with a `multi-exec`, the same script also does some comparisons and increments a second key, which I have omitted.

### DIFF
--- a/examples/eval.js
+++ b/examples/eval.js
@@ -12,3 +12,39 @@ client.eval([ "return 100.5", 0 ], function (err, res) {
     console.dir(err);
     console.dir(res);
 });
+
+
+client.eval("return {KEYS[1],KEYS[2],ARGV[1],ARGV[2]}", ["a", "b"], ["c", "d"], function (err, res) {
+	console.dir(err);
+	console.dir(res);
+});
+
+ //test {EVAL - Allow variadic KEYS and ARGS to be passed to script in array format}
+client.eval(["return {KEYS[1],KEYS[2],ARGV[1],ARGV[2]}", ["a", "b"], ["c", "d"]], function (err, res) {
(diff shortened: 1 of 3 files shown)


BridgeAR:
@cPu1 I'd like to add the object notation to v.3.0. Do you mind rebasing your pull request? There is no special handling for eval anymore and all commands can be used with the array notation, so these do not have to be tested anymore.

PR #634, opened 2014-07-27
Sending Arrays in Redis Multi Commands

srinivasiyer:
While chaining multiple commands to client.multi(), the commands only accepted comma separated arguments directly. I made a small change in the index file that allows params to be passed in arrays. The command was earlier being sent to the queue as a multi dimensional array which made the command fail.

Therefore earlier, only client.multi().zadd('foobar', 1, 'foo', 2, 'bar' ) was getting accepted. 
With the new change, the command can be sent as  
client.multi().zadd(['foobar', 1, 'foo', 2, 'bar']) or
client.multi().zadd('foobar',[ 1, 'foo', 2, 'bar'])
and it would work correctly. It would be a very useful fields for commands where multiple arguments need to be passed. I personally thought it was possible until the code returned with nothing.

Hope we can consider this change. I have also written a test MULTI_9 which passed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -994,7 +994,14 @@ commands.forEach(function (fullCommand) {
     RedisClient.prototype[command.toUpperCase()] = RedisClient.prototype[command];
 
     Multi.prototype[command] = function () {
-        this.queue.push([command].concat(to_array(arguments)));
+        if(Array.isArray(arguments[0])){
+            this.queue.push([command].concat(arguments[0]));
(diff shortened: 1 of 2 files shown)


jonathanong:
:+1:

BridgeAR:
Needs a rebase. Otherwise LGTM

bcoe:
Seems like a reasonable patch to get out in version 2.0 to me too.

BridgeAR:
Closing this in favor of #838

PR #647, opened 2014-08-19
Add more chaining to RedisClient

tkazec:
I was annoyed about `#select()` and `#auth()` so I fixed it for almost everything on RedisClient. Avoided a few methods that return something. None of these ever did:

    install_stream_listeners
    initialize_retry_vars
    unref
    flush_and_error
    on_error
    do_auth
    on_connect
    on_ready
    ready_check
    send_offline_queue
    connection_gone
    on_data
    return_error
    return_reply
    pub_sub_command
    select
    auth
    eval

### DIFF
--- a/index.js
+++ b/index.js
@@ -118,6 +118,8 @@ RedisClient.prototype.install_stream_listeners = function() {
         self.should_buffer = false;
         self.emit("drain");
     });
+    
+    return this;
 };
 
 RedisClient.prototype.initialize_retry_vars = function () {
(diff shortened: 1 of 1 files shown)


tkazec:
bump

BridgeAR:
What is the benefit of adding chaining to these methods? And what did you want to do? Please show an example.

bcoe:
my concern with this pull is that I don't really think we've made a huge effort to expose a fluent interface across the board in node_redis. My feeling is that having a partial chaining interface could mislead people a bit, since many commands don't behave this way, e.g.,

```js
> c.get('foo')
true
```

BridgeAR:
Closing as won't fix

PR #649, opened 2014-08-22
Attach command_obj to return error

guileen:
Base on the error reply, we can't know where the error real happens, and what command relative it, if there are multiple command executing at the same time.

So we have to trace what command occurs the error, we need attach the command_obj to the error.


BridgeAR:
@guileen would you be so kind and add a test? Otherwise LGTM

bcoe:
seems like a nice little tweak to release in version 2.0 to me.

BridgeAR:
Closing this in favor of #841

Issue #655, opened 2014-09-04
pls no, my pretty output for my test framework is now muttled with uneeded info.


jjn2009:
https://github.com/mranney/node_redis/blob/master/index.js#L512 
https://github.com/mranney/node_redis/blob/master/index.js#L531

blainsmith:
In order to better support this project and its new group of collaborators under a new org we are trying to clean up and close issues older than 6 months. Your issue may have been fixed since the issue was open, however, if you are still experiencing a problem **please** re-open this issue and we will label it accordingly.

PR #656, opened 2014-09-04
#655: only output errors if I ask

jjn2009:


BridgeAR:
@jjn2009 would you be so kind and use the new debug statement and use it on all console.* entries? That would be great :)

BridgeAR:
Will be solved by #815 => closing

PR #665, opened 2014-09-22
fixed createclient error when args are undefined

deeperx:
When createclient called without arguments and needed environment variables, "unknown type of connection in createclient()" error occurs in some machines. This fixes it.


luin:
+1 for this pr

BridgeAR:
This has been fixed by #786

deeperx:
Ok then.

PR #671, opened 2014-09-30
Fixes: offline_queue and option max_attempts don't work together as expected #280

qdb:
Fiexes:  https://github.com/mranney/node_redis/issues/280 and https://github.com/mranney/node_redis/pull/533

Make sure offline queue does not get flushed on subsequent reconnect attempts.


BridgeAR:
Thx a lot for your work @qdb !

Issue #321, opened 2012-10-30
Optional callback when using Array args


steelThread:
Quick question regarding client's supplying an optional callback when sending commands args via an Array.  The optional callback doesn't appear to be optional anymore for me in this case:

```javascript
client.del(['key1', 'key2']);

// or
client.del(['key1', 'key2'], undefined);
```

Which results in `351566259.101886 "del" "key1,key2"` in the MONITOR.  In this case I was expecting the Array to be expanded.  

This worked at some point but after a recent upgrade I'm noticing it is no longer legit due to the following logic:

```javascript
RedisClient.prototype[command] = function (args, callback) {
      if (Array.isArray(args) && typeof callback === "function") {
          return this.send_command(command, args, callback);
      } else {
          return this.send_command(command, to_array(arguments));
      }
};
```

Which will wrap the Array of args in another Array when the callback is not a function, hence what I'm seeing in the monitor.  Is the function check on the callback necessary here?  Maybe I'm missing something?

Thanks

blainsmith:
In order to better support this project and its new group of collaborators under a new org we are trying to clean up and close issues older than 6 months. Your issue may have been fixed since the issue was open, however, if you are still experiencing a problem **please** re-open this issue and we will label it accordingly.

BridgeAR:
This is fixed in newer versions

PR #694, opened 2014-12-09
hmset can be called without function callback

liorsion:
sometimes I just want to use an array to hmset without the need for callback


liorsion:
@DTrejo commenting with indignation..

index.js
```diff
@@ -1046,7 +1046,7 @@ RedisClient.prototype.HMGET = RedisClient.prototype.hmget;
 RedisClient.prototype.hmset = function (args, callback) {
     var tmp_args, tmp_keys, i, il, key;
 
-    if (Array.isArray(args) && typeof callback === "function") {
+    if (Array.isArray(args) && (typeof callback === "function" || callback === undefined)) {
```
BridgeAR: Removing the check for callback would be better.

BridgeAR:
@liorsion if you remove the check and add a test, this should be fine to merge.

BridgeAR:
Reference #321

PR #701, opened 2015-01-05
Added ability to log redis commands and their replies

alexcarruthers:
This allows for insight into what's going into and out of redis without the performance hit created by monitor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,7 @@ limits total amount of reconnects.
 * `auth_pass` defaults to `null`. By default client will try connecting without auth. If set, client will run redis auth command on connect.
 * `family` defaults to `IPv4`. The client connects in IPv4 if not specified or if the DNS resolution returns an IPv4 address. 
 You can force an IPv6 if you set the family to 'IPv6'. See nodejs net or dns modules how to use the family type. 
+* `log_commands` defaults to false. If true, will log all commands and responses to the console.
 
 ```js
     var redis = require("redis"),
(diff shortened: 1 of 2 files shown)


brycebaril:
There is already a debug mode you can enable in the client -- does that not do what you need?

alexcarruthers:
I'm looking for something that is not used purely for testing purposes. This is more for debugging a production environment after something has gone wrong. This produces one statement that is easy to track in logging tools whereas debug produces a bunch of unconnected statements that don't mean anything on their own and is almost impossible to group them together.

BridgeAR:
Ping @brycebaril I think it does not really hurt to add this.

brycebaril:
I'll leave it up to you @BridgeAR, my biggest issues are:
1. interface bloat (adding options to the constructor when there are already alternatives)
2. this still could be a significant burden to run in production if your values are large
3. it misses the error case

I agree though that this could be more easily used than the existing debug messages if you simply want to see the commands and values being returned.

BridgeAR:
@alexcarruthers if you fix the mentioned issues by @brycebaril I'd be fine to merge this.

Please do so by:
1. Use a env variable instead of adding another option. That way it is also possible to start and end this behavior at any point of the user program.
2. Only print replies that are shorter than an arbitrary small value (e.g. 100 chars) and add dots to the end of the reply if the reply exceeds that value.
3. Add the error case

BridgeAR:
And please add a test too

PR #716, opened 2015-02-24
Fixed missing word in README.md

davemun:
Sentence was missing a word.

Repo:
> client will emit ready a connection is established to the Redis server

PR:
> client will emit ready **once** a connection is established to the Redis server


mranney:
Right you are. Thanks.

PR #719, opened 2015-02-28
FWD_ERRORS_1 keeps listening

kmyr:
Removed side effects of FWD_ERRORS_1 (e.g. printing incoming on pub/subuse of client3).

```javascript
client3.listeners("error").push(originalHandlers);
```
This line is problematic because according to [node/io source](https://github.com/iojs/io.js/blob/v1.x/lib/events.js#L387) `EventEmitter.listeners` returns a copy of the array and also since we are pushing when a concat (and assignment) was probably intended.

In addition `console.log("incoming");` is currently being executed for any code re-using client3 (none currently, but I was going to when I came across this).
As such, in my opinion we should
 - unsubscribe from the FWD_ERRORS_1 channel (in the interest of completeness, not strictly necessary since no other tests publish to it)
 - reset the message listeners to the way we found them.

We do not need to call `client3.removeAllListeners("error");` on teardown because the test should fail if the error event did not fire (only way the listener stays alive).

### DIFF
--- a/test.js
+++ b/test.js
@@ -453,8 +453,12 @@ tests.FWD_ERRORS_1 = function () {
     var toThrow = new Error("Forced exception");
     var recordedError = null;
 
-    var originalHandlers = client3.listeners("error");
+    var originalHandlers = {
+        "error": client3.listeners("error"),
+        "message": client3.listeners("message")
+    };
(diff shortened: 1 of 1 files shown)


brycebaril:
Thanks!

PR #692, opened 2014-12-08
Fix for channel names with spaces

pbihler:
Fix for #691

Issue #691, opened 2014-12-08
Channel names with spaces not properly resubscribed


pbihler:
If you use channel names with spaces, `client.subscribe('foo bar')` works fine, but after a reconnection, only `foo` is resubscribed on redis.

blainsmith:
In order to better support this project and its new group of collaborators under a new org we are trying to clean up and close issues older than 6 months. Your issue may have been fixed since the issue was open, however, if you are still experiencing a problem **please** re-open this issue and we will label it accordingly.

PR #720, opened 2015-03-01
Channel names with spaces + test

kmyr:
#692 #691

Added a test for pbihler/node_redis@a1a7abd as requested in #692.
Fails under the current code because c1 resubscribes on the wrong channel name.


kmyr:
Not sure what happened with the line-break that was inserted, Really should have caught that before opening the pull request. Will fix that and re-open

kmyr:
Should be fixed now

PR #721, opened 2015-03-02
Channel names with spaces + test

kmyr:
(Let's try this again without the extraneous space.)
#692 #691

Added a test for pbihler/node_redis@a1a7abd as requested in #692.
Fails under the current code because c1 resubscribes on the wrong channel name.

I'm happy to squash these two commits into one if you prefer, but I figured I would leave it so pbihler gets credit where it's due.

### DIFF
--- a/index.js
+++ b/index.js
@@ -357,7 +357,8 @@ RedisClient.prototype.on_ready = function () {
             }
         };
         Object.keys(this.subscription_set).forEach(function (key) {
-            var parts = key.split(" ");
+            var space_index = key.indexOf(" ");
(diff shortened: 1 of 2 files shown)


kmyr:
indignation! grr..

kmyr:
Closing since it seems unlikely it will be reviewed or merged at this point.

erinspice:
Sorry @kmyr! We are planning on doing some major cleaning when we have the new test suite in place. We are up to 67.69% coverage. Just need to get back up to where the current test coverage is before we do a lot of reorg that could introduce problems.

PR #615, opened 2014-06-26
Emitting error event when connection permanently goes down

tobek:
Also documenting the "reconnecting" event.

Happy to discuss this further, but either way, when working with sharded Redis instances (for instance using https://github.com/blindsey/redis-shard) it's helpful to have some way to know when a Redis server seems to have totally gone down so that we can send out alerts, remove it from the cluster, spin up another server, etc.

PR #724, opened 2015-03-13
Feature: Emit an event when the maximum number of connection attempts has been reached

jackquack:
I followed the lead of the TODO comment, though I think my preferred solution would be that an error event is raised, similar to when a connection cannot be established in the first place, or when the redis server disappears completely.

Anyway, we wanted this for our own purposes, and it was easy to add. :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -507,8 +507,8 @@ RedisClient.prototype.connection_gone = function (why) {
 
     if (this.max_attempts && this.attempts >= this.max_attempts) {
         this.retry_timer = null;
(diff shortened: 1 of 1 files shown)


Reregistered:
+1

Anubisss:
+1

BridgeAR:
Duplicate of #615 

@jackquack if you're up to fix the comments of 615 before @tobek does (the PR is quite old already) and add some tests, I'll merge this.

PR #729, opened 2015-03-25
Implemented early event handlers

xytis:
Proposal for https://github.com/mranney/node_redis/issues/728


BridgeAR:
@xytis would you be so kind and write a test case?

PR #730, opened 2015-03-25
Add option to return es6 Map from hgetall

b3nj4m:
`hgetall` currently returns a plain object, which can be problematic for certain keys (e.g. `'hasOwnProperty'`, `'__proto__'`, etc.). Adding the option to return a `Map` would avoid these key problems.


b3nj4m:
cc @DTrejo

puzrin:
I think, problem can be partially solved, if `Object(null)` used insted of `{}` to create output.

BridgeAR:
This increases complexity and it seems to be a very theoretical issue. @b3nj4m if you use keys like that I recommend using a prefix. That would avoid any collision.

PR #734, opened 2015-04-02
Support createClient from url

fatelei:
Make user can create redis client from given url.

For example

```
var redis = require('redis');
var cli = redis.createClient('localhost:6379');

// then do something with the client.
...
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .tern-port
+stdout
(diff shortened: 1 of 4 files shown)


puzrin:
https://github.com/ddollar/redis-url

This package has more complete url support (auth, db index, options)

fatelei:
Thanks, I will close this pull-request.

Issue #512, opened 2013-11-16
send_command("monitoring") is doomed to fail


bpytlik:
On line [360 of index.js](https://github.com/mranney/node_redis/blob/master/index.js#L360) the code says
```
this.send_command("monitor")
```

In the send_command function, examine [line 735](https://github.com/mranney/node_redis/blob/master/index.js#L735) and line 756. If `args` is not an array, we throw the error on line 756. Since only the string "monitor" was passed on line 360, we're guaranteed to hit this error.

I think this is preventing me from monitoring more than one redis instance in the same process space, even though I'm using different clients. Regardless of whether I'm doing something else dumb (definitely possible), the code as written can't work I think.

bpytlik:
This seems to fix my issue. Any chance this can be pulled in soon?

bpytlik:
This issue still exists, though it's now at line 372 (in master as I write this). Please reopen this issue. Also, please accept the pull request that fixes this issue and has been sitting there for 3 months.

blainsmith:
I haven't been able to reproduce this error. I am in the middle of writing a test for this command as well.

bpytlik:
The issue exists by inspection. Here is the line in question:
```
this.send_command("monitor");
```
Here is the send_command function:
```
RedisClient.prototype.send_command = function (command, args, callback) {
    var arg, command_obj, i, il, elem_count, buffer_args, stream = this.stream, command_str = "", buffered_writes = 0, last_arg_type, lcaseCommand;

    if (typeof command !== "string") {
        throw new Error("First argument to send_command must be the command name string, not " + typeof command);
    }

    if (Array.isArray(args)) {
        if (typeof callback === "function") {
            // probably the fastest way:
            //     client.command([arg1, arg2], cb);  (straight passthrough)
            //         send_command(command, [arg1, arg2], cb);
        } else if (! callback) {
            // most people find this variable argument length form more convenient, but it uses arguments, which is slower
            //     client.command(arg1, arg2, cb);   (wraps up arguments into an array)
            //       send_command(command, [arg1, arg2, cb]);
            //     client.command(arg1, arg2);   (callback is optional)
            //       send_command(command, [arg1, arg2]);
            //     client.command(arg1, arg2, undefined);   (callback is undefined)
            //       send_command(command, [arg1, arg2, undefined]);
            last_arg_type = typeof args[args.length - 1];
            if (last_arg_type === "function" || last_arg_type === "undefined") {
                callback = args.pop();
            }
[...]

PR #749, opened 2015-05-08
Issue #512 send_command("monitoring") is doomed to fail

danielbprice:
I tried to work out how to add a test case for this condition but it was challenging since the current
tests don't really deal with disconnects/reconnects/etc.  I have run some of my own tests with
this change in place and it seems to work but additional validation would be welcomed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -369,7 +369,7 @@ RedisClient.prototype.on_ready = function () {
         });
         return;
     } else if (this.monitoring) {
-        this.send_command("monitor");
+        this.send_command("monitor", []);
     } else {
         this.send_offline_queue();
     }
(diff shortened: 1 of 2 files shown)


blainsmith:
Since we have updated the testing suite would you mind adding tests for this now? Then we can get this merged.

danielbprice:
I'll try to find some time to work on it soon.

danielbprice:
I added a test case, and confirmed that, without the fix in place, the bug is triggered by the test case.  With the fix in place, the test is passing.

blainsmith:
Thanks for adding tests. @bcoe @erinspice care to take a quick look?

bcoe:
@danielbprice awesome contribution thank you!

erinspice:
Great! Thanks, @danielbprice!

PR #751, opened 2015-05-17
Return Error object for errors in reply to EXEC

dohse:
The javascript reply parser does produces strings instead of `Error`
objects for errors in replies to EXEC. Return `Error` objects to be
consistent with hiredis.


BridgeAR:
Needs a rebase. Otherwise SGTM

bcoe:
@BridgeAR should we close this in favor of #826

BridgeAR:
Closing this in favor of #826 
@dohse I cherry-picked your commit so you'll still get the credit for it.

PR #752, opened 2015-05-17
Allow garbage collection of processed queue items

dohse:
Set processed queue items to `null` to allow garbage collection of these
items. The command queue contains callbacks. They contain the stack of
the current fiber, which can be large. Allow earlier garbage collection
of these stacks by removing the reference to the queue items.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -18,7 +18,10 @@ Queue.prototype.shift = function () {
             return;
         }
     }
-    return this.head[this.offset++]; // sorry, JSLint
+    var item = this.head[this.offset];
+    this.head[this.offset] = null;
+    this.offset++;
+    return item;
 };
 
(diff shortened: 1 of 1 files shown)


mren:
LGTM

bcoe:
@dohse looks good, but would you mind rebasing and adding a test or two to `/test/queue.spec.js`?

BridgeAR:
@bcoe Hm, what tests should be added here? There is already a test for the function itself so I'd say it's ready to merge.

bcoe:
@BridgeAR fine by me, but let's rebase with master so that we force a full test-suite run ... this does change the behavior of the queue slightly :+1:

BridgeAR:
As all tests pass with the patch applied, I'll merge it. Thx @dohse

PR #772, opened 2015-06-30
Small improvements (performance, ensure valid constructor usage)

Ginden:
* Increased perfomance of `to_array` method.
Following good rule "any perfomance should be proven by benchmark" I provide following code: https://gist.github.com/Ginden/4b3ca5bc48634fa9500c
```
toArray1 took 13.677ms on 32768 arrays
toArray2 [current] took 20.659ms on 32768 arrays
toArray3 [new] took 11.943ms on 32768 arrays
toArray4 took 12.269ms on 32768 arrays
```

* `'use strict'` directive was added to enforce valid constructor usage. It's minor "potentially breaking" change - it will make already buggy code fail earlier.

### DIFF
--- a/lib/parser/hiredis.js
+++ b/lib/parser/hiredis.js
@@ -1,3 +1,4 @@
+'use strict';
 var events = require("events"),
     util = require("../util"),
     hiredis = require("hiredis");
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -1,6 +1,8 @@
 // Queue class adapted from Tim Caswell's pattern library
 // http://github.com/creationix/pattern/blob/master/lib/pattern/queue.js
 
+'use strict';
+
 function Queue() {
     this.tail = [];
     this.head = [];
@@ -50,7 +52,8 @@ Queue.prototype.getLength = function () {
 Object.defineProperty(Queue.prototype, "length", {
     get: function () {
         return this.getLength();
-    }
+    },
+    configurable: false
 });
 
 
--- a/lib/to_array.js
+++ b/lib/to_array.js
@@ -1,12 +1,6 @@
-function to_array(args) {
-    var len = args.length,
-        arr = new Array(len), i;
-
-    for (i = 0; i < len; i += 1) {
-        arr[i] = args[i];
-    }
-
-    return arr;
-}
-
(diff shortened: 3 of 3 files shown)


BridgeAR:
@Ginden Your benchmark does not really test what you think it does. toArray1, toArray3 and toArray4 are all using the same functions internal and the first is going to warm the other ones up. In fact your test does not take that into acount. v8 has a jit and has to be warmed up before comparing. So you should have warmed all functions up. Besides that the dummyObject does not return any value ;).

But that's not really my concern. The to_array function is there for slicing arguments. You did not benchmark this for arguments but for simple arrays.
After all, `[].slice.call(arr);` is indeed faster for simple arrays but the function will leak arguments and deopt in v8 if you use this for arguments.

In that case the slice will be drastically slower than the function in place right now.

Ginden:
@BridgeAR: 
> Your benchmark does not really test what you think it does. toArray1, toArray3 and toArray4 are all using the same functions internal and the first is going to warm the other ones up.
This statement isn't true. Functions provided in standard library are treated specially, being [almost] always inlined, [almost] always optimized and so on.

>After all, [].slice.call(arr); is indeed faster for simple arrays but the function will leak arguments and deopt in v8 if you use this for arguments.
> In that case the slice will be drastically slower than the function in place right now.
This statement isn't true, too. If you are calling `func(arguments)` deopt does not depend on `func` internals.
BTW - AFAIR it doesn't causes deopt but prevents optimization.

And even if your argument isn't true (`arguments` materialization isn't benchmarked here), your thesis is true - it's slower. https://gist.github.com/Ginden/4b3ca5bc48634fa9500c
>toArray1 took 29.041ms on 32768 arrays
>toArray2 took 11.748ms on 32768 arrays
>toArray3 took 19.089ms on 32768 arrays
> toArray4 took 19.222ms on 32768 arrays


>Besides that the dummyObject does not return any value 
This statement is true.
>You did not benchmark this for arguments but for simple arrays.
You are right.

mranney:
It wouldn't surprise me if at some point V8 would implement a faster way to do this. But to_array is there because this process kept showing up on the flame graphs.

PR #775, opened 2015-07-10
Added example of a Redis SCAN command

jeremiahlee:
I needed to do this and found the module's documentation missing such an example. Redis SCANs are also a bit tricky. Hopefully this helps someone else out.

### DIFF
--- a/examples/scan.js
+++ b/examples/scan.js
@@ -1,0 +1,34 @@
+var redis  = require("redis"),
+    client = redis.createClient();
+
+var cursor = 0;
+
+function scan() {
+    client.scan(
+        cursor,
+        "MATCH", "q:job:*",
+        "COUNT", "10",
+        function(err, res) {
+            if (err) throw err;
+
+            // Update the cursor position for the next scan
+            cursor = res[0];
+
+            // From <http://redis.io/commands/scan>:
+            // An iteration starts when the cursor is set to 0,
+            // and terminates when the cursor returned by the server is 0.
+            if (cursor === 0) {
+                return console.log('Iteration complete');
+            } else {
+                // Remember, more keys than COUNT or no keys may be returned
+                // See http://redis.io/commands/scan#the-count-option
+                if (res[1].length > 0) {
+                    return console.log('Array of matching keys', res[1]);
+                } else {
+                    // No keys were returned in this scan, but more keys exist.
+                    return scan();
+                }
+            }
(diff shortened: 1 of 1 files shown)


examples/scan.js
```diff
@@ -0,0 +1,34 @@
...
+            // From <http://redis.io/commands/scan>:
+            // An iteration starts when the cursor is set to 0,
+            // and terminates when the cursor returned by the server is 0.
+            if (cursor === 0) {
+                return console.log('Iteration complete');
+            } else {
+                // Remember, more keys than COUNT or no keys may be returned
+                // See http://redis.io/commands/scan#the-count-option
+                if (res[1].length > 0) {
+                    return console.log('Array of matching keys', res[1]);
```
raydog: it seems that in this case, scan() wouldn't be called again, so we'd only get the first batch of items out of Redis, and we wouldn't get the full set of keys that match "q:job:*".

I feel it'd be better to do something like:

```javascript
if (res[1].length > 0) {
  console.log('Array of matching keys', res[1]);
}
return scan();
```

examples/scan.js
```diff
@@ -0,0 +1,34 @@
...
+            // Update the cursor position for the next scan
+            cursor = res[0];
+
+            // From <http://redis.io/commands/scan>:
+            // An iteration starts when the cursor is set to 0,
+            // and terminates when the cursor returned by the server is 0.
+            if (cursor === 0) {
+                return console.log('Iteration complete');
+            } else {
+                // Remember, more keys than COUNT or no keys may be returned
```
raydog: Also, the docs don't say that 'more than COUNT or 0' will be returned. It just says that COUNT should be roughly the number of items returned. Could be more. Could be less.

raydog:
I like the initiative of this pull request. SCAN is a tricky beast, and needs better documentation + examples. Two things, however:
- The docs don't actually guarantee a certain number of results. The COUNT value is actually closer to a target number of items, and Redis will try to hit that number. Sometimes more, sometimes less.
- Somewhere in here, we should point out that using SCAN can sometimes return duplicate keys, especially when the keyspace is changing while doing the full scan.

Thanks for the contribution!

jeremiahlee:
Thanks for the feedback, @raydog. I have updated the pull request.

mranney:
Thanks for the improvements on this, everybody. Great stuff.

PR #776, opened 2015-07-11
various fixes to tests, along with groundwork for adding travis and coveralls support

bcoe:
* upgraded `nyc` to the newest release, it has much nicer test output:

screen shot 2015-07-11 at 4 17 10 pm

* added .travis.yml, so that as soon as soon as (@brycebaril, or @mranney configure Coveralls, and Travis we can add some badges).
* added a test for queue.js (which had poor coverage), in the process refactored the tests so that it's easier to split them into multiple files.
* made it so tests are run with both the `hiredis` and the `javascript` parser -- this raised coverage significantly.
* fixed a major bug with the test suite, once you hit the domains test all tests afterwords did not properly bubble errors.

Doing all this brought test coverage up to `86%` not too shabby.

reviewers: @raydog @erinspice

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,6 @@
 examples/
 benches/
-test.js
+test/
 diff_multi_bench_output.js
 generate_commands.js
 multi_bench.js
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: node_js
+sudo: false
+node_js:
+  - "0.10"
+  - "0.12"
+  - "iojs"
+after_success: npm run coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,8 @@ node_js:
   - "0.10"
   - "0.12"
   - "iojs"
+before_install:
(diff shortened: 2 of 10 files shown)


raydog:
There's a lot of good stuff in here. Do you feel there'd be value in moving node_redis to use a standard test runner, like mocha or similar? I always felt it strange that this library rolls it's own... Who knows what other bugs like the domain one exists in here...

bcoe:
@raydog I was also thinking it would probably be worth moving to mocha, or maybe node-tap which @isaacs' has been putting a ton of work into.

I'd be happy to take that on as round two of this work, I think it can be done separately though?

erinspice:
Agree! Let's convert them to use mocha or similar.

bcoe:
@erinspice @raydog, I've managed to get stuff running on travis. I'm going to go ahead and rebase/merge this, but let's open an issue to discuss moving things to mocha.

mranney:
For sure, let's use something standard.

In case you are wondering why things are the way that they are, I had originally found it very hard to realistically mock redis-server, so I wrote my own "test suite" that used a real redis-server.

That was like 5 years ago, and I'm sure there are much better ways to do things now.

PR #779, opened 2015-07-12
I think it would be cool if we added a logo to this organization

bcoe:
I read the trademark guidelines of [Joyent](https://nodejs.org/about/trademark/) and [Redis](http://redis.io/topics/trademark), and I think it would be fine to add a logo like this to the organization:

node-redis-logo

1. it leaves both logos in their original form.
2. there's a registered trademark symbol on both logos.
3. this project is open-source and non-profit.
4. I've added attribution to the README.md.

thoughts, cc: @erinspice, @raydog, @mranney?

### DIFF
--- a/README.md
+++ b/README.md
@@ -807,3 +807,13 @@ HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
 WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
 FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 OTHER DEALINGS IN THE SOFTWARE.
+
+## Other Legal Stuff
+
+Redis, and the Redis logo are the trademarks of Salvatore Sanfilippo in the U.S. and other countries:
+
+http://redis.io/topics/trademark
+
+Node.js is a trademark of Joyent:
+
+https://nodejs.org/trademark-policy.pdf


bcoe:
@antirez has :-1: using the redis logo -- guess we need another logo for NodeRedis. CC: @substack.

jcblw:
could be cool to use instead of squares in the redis logo hexagons while still retaining the hexagon shape of all the stacks together

node-redis

bcoe:
@jcblw I like it!

simontabor:
I'm no designer but +1 @jcblw - nice, clear and simple. Looks somewhat like an AWS service logo but I don't think that's a problem

jcblw:
@bcoe & @simontabor thanks! If anyone wants to play around with the design here is the svgs.

https://gist.github.com/jcblw/428faaea2faf023acbf4

as well as i made a version with rounded corners, like the node and redis logo both have. Which is also included in the svgs above.

node-redis rounded

stockholmux:
A slight tweak to how the project name text is treated - just more definition between "node" and "redis"

[https://gist.github.com/stockholmux/6de53939170b3815b386](https://gist.github.com/stockholmux/6de53939170b3815b386)

redis-new-text-treatment

bcoe:
@jcblw I like @stockholmux's tweak, what do you think?

simontabor:
Not sure about the black text in @stockholmux's version - any chance we can try it around rgb(55,55,55)?

+1 for rounded corners though, looks great

antirez:
Sorry for saying no to the use of the Redis logo. I use the Redis logo, given that the name itself is used everywhere, in order to identify things that are directly output of the Redis project and directly monitored by myself. Good luck for your project!

stockholmux:
@simontabor Good point. Changed to rgb(55,55,55), gist updated.

redis-new-text-treatment

bcoe:
@antirez no worries, completely understand -- it's forced us to get a little more creative :)

jcblw:
nice @bcoe @stockholmux  i like it! :raised_hands:  @antirez nice call on the the text color

PR #785, opened 2015-07-22
use strict

BridgeAR:
This is going to improve the speed and code safety a tiny bit by using [`use strict`] (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Strict_mode) plus some style changes (adding semicolons). I added the later as I read that most of you seemed to prefer those. Using use strict in node should not break any code if it has not been broken before, since node encapsulates each file and does not apply it global.


benches/stress/codec.js
```diff
@@ -1,16 +1,18 @@
...
-var MsgPack = require('node-msgpack');
-msgpack = {
-	encode: MsgPack.pack,
-	decode: function(str) { return MsgPack.unpack(new Buffer(str)); }
-};
+// var MsgPack = require('node-msgpack');
+// var msgpack = {
+// 	encode: MsgPack.pack,
+// 	decode: function(str) { return MsgPack.unpack(new Buffer(str)); }
+// };
```
erinspice: Can you explain to someone not so familiar with the codebase why you commented this out? If it's not used at all, it should be deleted.

erinspice:
I fully support this! I asked one question about some commented code.

benches/stress/codec.js
BridgeAR: `module.exports = msgpack;` is already commented out and therefor this is a unused variable. I was not sure what this part was for and normally I'd just delete unused code, but I'd prefer to know the reason for the original code before deleting it :).

erinspice: Thanks!

BridgeAR:
Rebased

josephpage:
:+1:

PR #786, opened 2015-07-22
Rewrite createClient. Fixes #651

### LINKED ISSUE #651: Passing 'undefined' to createClient fails with >=0.12
Previously I could pass undefined variables to createClient and they'd be populated as the defaults. Since 0.12 null must be passed due to the === operator. 

Should the old behaviour still work? If so it's an easy fix by changing to == in createClient

BridgeAR:
I'd like to add some tests for it but the test suite right now is not really suitable as I see it. So this should probably wait until the mocha tests got merged as planned.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1228,50 +1228,29 @@ RedisClient.prototype.eval = RedisClient.prototype.EVAL = function () {
     });
 };
 
+exports.createClient = function(arg0, arg1, options) {
+    if (typeof arg0 === 'object' || arg0 === undefined) {
+        options = arg0 || options;
+        return createClient_tcp(default_port, default_host, options);
+    }
+    if (typeof arg0 === 'number' || typeof arg0 === 'string' && arg0.match(/^\d+$/)){
+        return createClient_tcp(arg0, arg1, options);
+    }
+    if (typeof arg0 === 'string') {
+        options = arg1 || {};
 
-exports.createClient = function(arg0, arg1, arg2){
-    if( arguments.length === 0 ){
-
-        // createClient()
-        return createClient_tcp(default_port, default_host, {});
-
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1228,50 +1228,29 @@ RedisClient.prototype.eval = RedisClient.prototype.EVAL = function () {
...
-
-        // createClient( 3000, host, options)
-        // createClient('3000', host, options)
-        return createClient_tcp(arg0, arg1, arg2);
-
-    } else if( typeof arg0 === 'string' ){
-        var parsed = URL.parse(arg0, true, true),
-            options = (arg1 || {});
-
+        var parsed = URL.parse(arg0, true, true);
```
erinspice: It looks like `URL.parse` does not return a `parsed.port` when the string is only a string number like "6379", so this introduces breaking changes. Am I wrong?

```
erin@elisheva:~/node_redis$ node  -e "var URL = require('url'); console.log(URL.parse('8000'));"
{ protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: '8000',
  path: '8000',
  href: '8000' }
```

BridgeAR: I did not test every possibility and would like to add some before this is merged but I'm quite certain the behavior is just the same as it was before. There should not be any BC.

This code is the same as it was before in [line 1245] (https://github.com/NodeRedis/node_redis/pull/786/files?diff=unified#diff-168726dbe96b3ce427e7fedce31bb0bcL1245).

erinspice: Doh, you're right. Looks like this bug predates your PR. Also, all of the new mocha tests pass against this.

PR #799, opened 2015-08-26
Run single command multis without a multi

### LINKED ISSUE #767: multi.exec is slow
multi.exec is slow even if I only call 1 set

set  32000/s
multi (with one set) 12000/s

simontabor:
Extracts multi's that contain just a single command and runs them without the multi/exec.

Should fix #767. 

- Makes a single command multi run ~5-10% faster when running them in series
- When running many in parallel, it runs 2.6x faster
- Has no impact on multi's with more than one command

### DIFF
--- a/index.js
+++ b/index.js
@@ -1109,6 +1109,31 @@ Multi.prototype.exec = function (callback) {
     var self = this;
     var errors = [];
     var wants_buffers = [];
+
+    // check for a queue of just one command (after MULTI)
+    if (this.queue.length === 2) {
+        var args = this.queue[1];
+        var command = args[0];
+        var commandCallback;
+        if (typeof args[args.length - 1] === "function") {
+            commandCallback = args[args.length - 1];
+        }
+        args = args.slice(1, commandCallback ? -1 : args.length);
+        return this._client.send_command(command, args, function(err, resp) {
+            if (err) {
+                if (callback) {
+                    callback([ new Error(err) ]);
+                    return;
+                } else {
+                    throw new Error(err);
+                }
+            }
+
+            if (commandCallback) commandCallback(err, resp);
+            if (callback) callback(err, [ resp ]);
+        });
+    }
+
     // drain queue, callback will catch "QUEUED" or error
     // TODO - get rid of all of these anonymous functions which are elegant but slow
     this.queue.forEach(function (args, index) {


simontabor:
This probably isn't going to be able to pass the tests due to https://github.com/NodeRedis/node_redis/blob/master/test/commands/multi.spec.js#L228 - we can't simulate the EXECABORT error.

Might be best to reject this and note that MULTI adds overhead (is that our job? shouldn't that be clear in redis' docs?)

Even better, we could add `batch` functionality, which is similar to multi but simply queues commands and can execute them all at once with an `exec` function. I reckon this (async control/executing a group of commands together) is a common reason people use `multi` (which without rollbacks isn't that useful in Node imo). I've got this in redis-clustr and it's incredibly useful: https://github.com/gosquared/redis-clustr/blob/master/src/RedisBatch.js.

```js
var batch = redis.batch();
batch.set('key', 'value');
batch.get('key');
batch.exec(function(err, resp) {
  // [ "OK", "value" ]
});
```

bcoe:
@simontabor I think that keeping as much intelligence in redis itself (and out of node-redis) is probably the right approach.

I see no reason why a user shouldn't be able to run a multi command of 1.

Do you think there's anything we could clarify in the node-redis docs for this?

simontabor:
@bcoe the reason to intercept a multi command of 1 is that the multi adds noticeable overhead which is completely unnecessary. I think that should be noted in the Redis docs (there's no mention anywhere of how efficient multi/exec is afaics).

What do you think of adding batch functionality to queue (just in memory) and execute a group of commands all at once? I've found it exceptionally useful and a great feature for handling async + groups of commands. I think one of the main reasons people use `MULTI` is not because they want a transaction, it's because they want to run a batch of commands together.

erinspice:
Also, looks like the tests are failing.

bcoe:
My concern is that `multi` and `exec` are Redis commands with fairly specific expectations and other redis commands that rely on these expectations, e.g., `watch`

For instance see this failing test:

```js
            it('does not execute transaction if watched key was modified prior to execution', function (done) {
                client.watch(watched);
                client.incr(watched);
                multi = client.multi();
                multi.incr(watched);
                multi.exec(helper.isNull(done));
            })
```

One can imagine scenarios where someone enters a transaction to perform a single operation, perhaps they don't want to increment a value if an other client has swooped in and modified a shared resource.

I feel like people tend to achieve the bulk operations you're describing using a combination of commands like:

`MGET`, `MSET`, `HMGET`, `HMSET`, etc.

Perhaps we could open a new issue discussing a further bulk operation abstraction that node_redis could provide? I'd prefer this to changing the behavior of `multi`/`exec` which in (granted confusing) Redis terminology facilitates transactional operations, not bulk operations.

simontabor:
Agreed, I'll open a new issue to discuss batch/group/bulk operations.

PR #802, opened 2015-08-30
Remove redundant check

jomo:
If the first operand of the disjunction (`!obj.loading`) is false, `obj.loading` is truthy.
Thus there is no need to test it again in second operand.


bcoe:
@jomo your logic is sound :) thanks for the contribution \o/

PR #807, opened 2015-09-01
Add jshint and fix some tests

BridgeAR:
Hey, I added jshint with a config that has no deprecated style rules (I'd add jscs in another PR if you want) and fixed some tests. I had to deactivate a few though, since they did not work as intended and there might be more. But I did not find the time to fix them yet.

I guess you'd all agree on the used jshint rules or are there any that you think should be removed again?

### DIFF
--- a/index.js
+++ b/index.js
@@ -955,7 +955,7 @@ RedisClient.prototype.select = function (db, callback) {
         if (err === null) {
             self.selected_db = db;
         }
-        if (typeof(callback) === 'function') {
+        if (typeof callback === 'function') {
             callback(err, res);
         } else if (err) {
             self.emit('error', err);
--- a/index.js
+++ b/index.js
@@ -1,7 +1,5 @@
 'use strict';
 
-/*global Buffer require exports console setTimeout */
-
 var net = require("net"),
     URL = require("url"),
     util = require("util"),
@@ -616,7 +614,7 @@ RedisClient.prototype.return_reply = function (reply) {
         type = reply[0].toString();
     }
 
-    if (this.pub_sub_mode && (type == 'message' || type == 'pmessage')) {
+    if (this.pub_sub_mode && (type === 'message' || type === 'pmessage')) {
         debug("received pubsub message");
     }
     else {
@@ -1172,30 +1170,6 @@ RedisClient.prototype.eval = RedisClient.prototype.EVAL = function () {
     });
 };
 
-exports.createClient = function(port_arg, host_arg, options) {
-    if (typeof port_arg === 'object' || port_arg === undefined) {
-        options = port_arg || options;
-        return createClient_tcp(default_port, default_host, options);
-    }
-    if (typeof port_arg === 'number' || typeof port_arg === 'string' && /^\d+$/.test(port_arg)) {
-        return createClient_tcp(port_arg, host_arg, options);
-    }
-    if (typeof port_arg === 'string') {
-        options = host_arg || {};
-
-        var parsed = URL.parse(port_arg, true, true);
-        if (parsed.hostname) {
-            if (parsed.auth) {
-                options.auth_pass = parsed.auth.split(':')[1];
-            }
-            return createClient_tcp(parsed.port, parsed.hostname, options);
-        }
-
-        return createClient_unix(port_arg, options);
-    }
-    throw new Error('unknown type of connection in createClient()');
-};
-
 var createClient_unix = function(path, options){
     var cnxOptions = {
         path: path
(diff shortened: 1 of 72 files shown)


test/commands/getset.spec.js
```diff
@@ -59,8 +60,7 @@ describe("The 'getset' method", function () {
                 describe("when the key exists in Redis", function () {
                     beforeEach(function (done) {
                         client.set(key, value, function (err, res) {
-                            helper.isNotError()(err, res);
-                            done();
+                            helper.isNotError(done)(err, res);
```
erinspice: I don't know why these take a callback.... they are synchronous. I think we should remove that ability in the future. If someone adds an assertion after this line, they have to know to move the done down. However, if we leave the done where it was and pass no callback, it will be obvious to place a new assertion above the done. IMO, these should all be changed back so that no callbacks are passed to assertion helpers.

erinspice:
Why did these particular tests begin to fail with this PR?

test/commands/getset.spec.js
BridgeAR: They are sync. I thought it's nicer to read it that way though. I made a separate commit for these style changes and I can change them back to the way they were before if you want :)

BridgeAR:
The failing tests are async and the done was missing

erinspice:
It is a shame these weren't found before the 1.0.0 release. We should evaluate whether these fixes should go in before we promote our 1.0.0 release to latest.

bcoe:
@BridgeAR I'm excited to pull in jshint, two comments:

1. could we make it so that we run the jshint as a `pretest` event in scripts, and fail tests if someone breaks the style? Otherwise my experience has been that people rarely run `npm run jshint`.
2. request the second, could we move the refactoring work that is being done on tests into a separate pull, I don't like that a bunch of tests are being set to `skip` as part of this pull, I'd rather that the tests were updated and fixed in a separate pull request.

bcoe:
@erinspice the `1.0.0` has already been promoted to `latest` from `next`. If we do find any actual issues we can release `1.0.1`, my guess would be it's just tests that need fiddling -- all tests were green prior to this pull, curious how adding jshint outlined problems in these tests? was it complaining about an unused callback or something @BridgeAR?

BridgeAR:
1. I'll add jshint as task before the tests run. That way you're sure everything passes
2. That's fine to me. The issue with them is though, that they did not work at all before (like one of them is a test that tried to check if the parameters are valid and the expectation was, that passing `undefined` as a key would throw. That is not the case though and it'll only throw in case of a wrong number of parameters. We could add a pre check in the code to check that no key may be undefined for those tests).

I did not find any issues in the code itself, so only the tests were flaky. When I added jshint I saw at least one place where the callback was not defined, so I checked other places too and found some broken tests. The only real code change I did was converting a string to a number in one place and that was only for consistency.

bcoe:
@BridgeAR we could also set the tests to `skip` in this pull, but not land this pull until we land a separate pull against this branch that fixes the tests?

BridgeAR:
I incorporated the feedback and changed the behavior of `set`, `setex` and `mset` so that they'll throw on undefined keys / values as suggested in the tests. This is a BC though (do you want to use semver?) but I'm uncertain if anyone would want to insert null or undefined values without converting them to a string beforehand. And there's another issue, since this behavior should likely also be the same for `msetnx`, `setnx` and others that are not yet handled. So a alternative would be to remove those tests and keep the old behavior as it was.

I did not fix the subscribe test yet, since until now I'm uncertain how that really works as I always get an `Connection in subscriber mode, only subscriber commands may be used` error even though the commands used seem to be subscriber commands... would you be so kind and have a look at that one?

erinspice:
@BridgeAR unfortunately, I am leaving today to be out of town until Monday. I wonder if @bcoe has time to check it out?

index.js
```diff
@@ -749,12 +746,22 @@ RedisClient.prototype.send_command = function (command, args, callback) {
...
     }
 
-    // if the value is undefined or null and command is set or setx, need not to send message to redis
-    if (command === 'set' || command === 'setex') {
-        if(args[args.length - 1] === undefined || args[args.length - 1] === null) {
-            var err = new Error('send_command: ' + command + ' value must not be undefined or null');
-            return callback && callback(err);
-        }
+    // if the value or the key is undefined or null and command is set, setex or mset; need not to send message to redis
+    if (command === 'set' || command === 'setex' || command === 'mset') {
```
bcoe: my preference would be to change the tests, rather than making this breaking change -- perhaps we could update the tests to reflect the actual behavior?

bcoe:
@BridgeAR great work. Except for my comment regarding the breaking change, I think this is pretty much ready to land. Could you outline in the issue specifically which tests have a problem, so that we can create better test cases?

index.js
BridgeAR: We might consider the change as a bug fix though. Using `null` and `undefined` as a key for set is not reasonable. And there are a few issues where people complain that they overwrite old values by accident because they did not check that the value was undefined or null.
I decided to remove this part from the jshint PR and make an individual one to fix that issue including the therefor broken tests.

BridgeAR:
@bcoe I'll remove the BC and fix the tests in a separate PR.

BridgeAR:
I reverted the BC and removed all broken tests from the PR. They will be handled in another PR. So this should be good to go.

PR #809, opened 2015-09-01
Remove some old stuff that is not needed anymore

BridgeAR:


index.js
```diff
@@ -211,14 +210,14 @@ RedisClient.prototype.do_auth = function () {
     self.send_anyway = true;
     self.send_command("auth", [this.auth_pass], function (err, res) {
         if (err) {
-            if (err.toString().match("LOADING")) {
+            if (/LOADING/.test(err.toString())) {
```
brycebaril: performance nit: creating these regular expressions as vars and moving them out of the send_command code-path would be a slight improvement.

index.js
```diff
@@ -314,11 +313,7 @@ RedisClient.prototype.init_parser = function () {
 
     // "reply error" is an error sent back by Redis
     this.reply_parser.on("reply error", function (reply) {
-        if (reply instanceof Error) {
-            self.return_error(reply);
-        } else {
-            self.return_error(new Error(reply));
-        }
+        self.return_error(reply);
```
brycebaril: Is this safe? I feel like there were cases that the error was a non-Error and needed to be wrapped.

index.js
```diff
@@ -587,27 +582,6 @@ RedisClient.prototype.return_error = function (err) {
...
-                currDomain.exit();
-            }
-        } else {
-            process.nextTick(function () {
-                throw err;
-            });
-        }
-    }
-}
-
```
brycebaril: Some people will be unhappy if you remove domain support.

index.js
```diff
@@ -17,10 +17,9 @@ var net = require("net"),
 // can set this to true to enable for all connections
 exports.debug_mode = false;
 
-var arraySlice = Array.prototype.slice;
 function trace() {
     if (!exports.debug_mode) return;
-    console.log.apply(null, arraySlice.call(arguments));
+    console.log.apply(null, to_array.call(arguments));
```
brycebaril: honestly for best perf we should just inline the contents of to_array here

index.js
BridgeAR: True :)

index.js
BridgeAR: I fixed the non-Error, so this is safe :)

brycebaril:
Some notes above, this one seems to be a fairly significant PR

index.js
brycebaril: Cool, was that the only one? (we don't have a dedicated test set for the parser)
Also is this true for hiredis?

index.js
brycebaril: however "auth" is pretty rare so... :stuck_out_tongue:

index.js
BridgeAR: hiredis did not have any non-Error and as it was the only spot in the js parser that was not an Error it should be fixed properly. And all tests still pass (they would not have without fixing the non-Error).

brycebaril: :+1:

index.js
simontabor: two thoughts here: can we not use `err.message` rather than `err.toString()`?

also, wouldn't all of these be better off using indexOf? `err.message.indexOf('string') !== -1`

index.js
BridgeAR: Hm, wrapping the user callback into a try catch should not be needed at all - this is actually causing [problems for users] (https://github.com/NodeRedis/node_redis/issues/453). So any user should still be able to handle the errors just as they want and it should be safe to remove this.

index.js
BridgeAR: This part is actually obsolete too as I later realized and will be superseded by debug (or what ever it'll later be).

index.js
brycebaril: great points @simontabor

index.js
brycebaril: Generally I agree that we should not be wrapping in try/catch, though this is a pretty major change that might merit its own PR, especially to make sure we're adequately addressing domains users.

index.js
BridgeAR: @simontabor err.message is very likely better but as far as I know a precompiled regex with .test is faster than .indexOf and it's at least more readable in my opinion. It's difficult to write a proper test to compare the regex against the .indexOf as v8 is very likely inlining everything right away.

brycebaril: I wouldn't worry too much about perf here because it's in the `auth` stanza, which is only used by a subset of Redis users, and even then is only called once per client connection, but not only that it's in the error pathway of the auth stanza, so this is a super rare code path.

I shouldn't have brought perf up ealier :grin: 

Mostly just need to worry about correctness

index.js
BridgeAR: I added the try_callback function back in again

BridgeAR:
Feedback incorporated

brycebaril:
LGTM, thanks for helping to clean up this library!

PR #811, opened 2015-09-02
Add "debug" instead of using different indidivudal styles for debugging

BridgeAR:
There were all sorts of ways to print debug statements and I moved all of those to use [debug] (https://github.com/visionmedia/debug) instead. This is handy, as it won't pollute the coverage with statements that are not necessary and debug is widespread.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,25 +12,15 @@ var net = require("net"),
     parsers = [], commands,
     connection_id = 0,
     default_port = 6379,
-    default_host = "127.0.0.1";
-
-// can set this to true to enable for all connections
-exports.debug_mode = false;
-
-var arraySlice = Array.prototype.slice;
-function trace() {
-    if (!exports.debug_mode) return;
-    console.log.apply(null, arraySlice.call(arguments));
-}
+    default_host = "127.0.0.1",
+    debug = util.debuglog('redis');
 
 // hiredis might not be installed
 try {
     require("./lib/parser/hiredis");
     parsers.push(require("./lib/parser/hiredis"));
 } catch (err) {
-    if (exports.debug_mode) {
-        console.warn("hiredis parser not installed.");
-    }
+    debug("hiredis parser not installed.");
 }
 
 parsers.push(require("./lib/parser/javascript"));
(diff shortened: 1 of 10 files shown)


brycebaril:
Not a huge fan of adding the dependency on debug, are we using any of the extra features it provides?

The alternative would be using the core `debuglog` and feature-testing to keep 0.10 compatibility, but perhaps that is best left to the debug module.

The debug module is fairly small and only has one leaf dependency of its own, so if we want to just use it vs trying to do our then I'd say go for it.

BridgeAR:
I removed debug and switched to util.debuglog, so there's no extra dependency needed.

index.js
```diff
@@ -12,25 +12,15 @@ var net = require("net"),
...
-// can set this to true to enable for all connections
-exports.debug_mode = false;
-
-var arraySlice = Array.prototype.slice;
-function trace() {
-    if (!exports.debug_mode) return;
-    console.log.apply(null, arraySlice.call(arguments));
-}
+    default_host = "127.0.0.1",
+    debug = util.debuglog('redis');
```
brycebaril: This unfortunately won't work on node 0.10 or lower, util.debuglog was added in 0.11

See https://github.com/brycebaril/client-request/blob/master/request.js#L15-L19 for what I did in another library to work around that

BridgeAR: Thx, I did not pay attention to that. I added a fallback mode for 0.10 that does just the same as the debuglog.

BridgeAR: That's actually not totally true as it won't print the process ID and "REDIS" before each output.

index.js
```diff
@@ -12,15 +12,16 @@ var net = require("net"),
...
     connection_id = 0,
     default_port = 6379,
-    default_host = "127.0.0.1";
-
-// can set this to true to enable for all connections
-exports.debug_mode = false;
+    default_host = "127.0.0.1",
+    debugModeEnabled = /\bredis\b/i,
+    debug = function () {
+        if (debugModeEnabled.test(process.env.NODE_DEBUG)) {
```
brycebaril: Probably can do the debug test outside the debug function, we don't need to support the user changing `proces.env.NODE_DEBUG` during execution.

E.g.
```
var debug = function noop() {};
if (debugModeEnabled.test(process.env.NODE_DEBUG)) {
  debug = function debug() {
    console.error.apply(...);
  }
}
if (util.debuglog) {
  debug = util.debuglog('redis');
}

BridgeAR: Good point and done. Now it's possible to redirect the call to console.error without having to convert the arguments.

brycebaril:
Awesome! Looks good!

PR #702, opened 2015-01-06
Fetch new list of commands

qix:

PR #814, opened 2015-09-03
Update commands list and remove unecessary code

BridgeAR:
The old code is not necessary at all and can safely be removed. There were only a few _very_ old redis commands that we should probably not support anymore.

Support is dropped for:
```js
[ 'substr' ]
```

New commands supported are:
```js
[ 'command', 'geoadd', 'geohash', 'geopos', 'geodist', 'georadius', 'georadiusbymember', 'hstrlen', 'role', 'wait', 'zrevrangebylex' ]
```

The new commands have no tests yet, but I'd say this is good to go right away and we should add some tests for this later on.


BridgeAR:
Superseds #702

bcoe:
@BridgeAR good job. It's cool to see those cluster commands coming in, it would be neat to put a test or two around those to get a feel for the API.

BridgeAR:
@bcoe cluster was already supported, but I'm uncertain if it works. So we should definitely add some tests but I guess that could be done in a separate PR?

index.js
```diff
@@ -711,36 +714,25 @@ function Command(command, args, sub_command, buffer_args, callback) {
...
+        //     client.command(arg1, arg2, cb);   (wraps up arguments into an array)
+        //       send_command(command, [arg1, arg2, cb]);
+        //     client.command(arg1, arg2);   (callback is optional)
+        //       send_command(command, [arg1, arg2]);
+        //     client.command(arg1, arg2, undefined);   (callback is undefined)
+        //       send_command(command, [arg1, arg2, undefined]);
+        callback = args.pop();
     }
 
-    if (Array.isArray(args)) {
```
BridgeAR: args is always an array

index.js
```diff
@@ -711,36 +714,25 @@ function Command(command, args, sub_command, buffer_args, callback) {
...
-            //     client.command(arg1, arg2);   (callback is optional)
-            //       send_command(command, [arg1, arg2]);
-            //     client.command(arg1, arg2, undefined);   (callback is undefined)
-            //       send_command(command, [arg1, arg2, undefined]);
-            last_arg_type = typeof args[args.length - 1];
-            if (last_arg_type === "function" || last_arg_type === "undefined") {
-                callback = args.pop();
-            }
-        } else {
-            throw new Error("send_command: last argument must be a callback or undefined");
```
BridgeAR: We do not have to protect users from inserting anything else than functions in the explicit argument

brycebaril: We have to make sure that we always support

`client.send_command()` directly, which was part of this code block

If we change the way the above behaves that's definitely a breaking change

BridgeAR: That's true, I forgot about that users might use this function directly. But actually it's not really a BC: if any user did not use the command properly before, he would have run into the errors. So this is less strict but it's going to fail somewhere else for users that do not use the function properly.

brycebaril: Yeah, I'm all for tightening this up, just raising the warning flag for potential backwards compatibility breakage

BridgeAR: I separated this in another PR

PR #815, opened 2015-09-03
Improve code coverage by adding tests and removing unnecessary code

BridgeAR:
I had a closer look at the coverage as it was and wrote some tests and improved the handling if there were redundant checks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ addons:
     - ubuntu-toolchain-r-test
     packages:
     - g++-4.8
+services:
+  - redis-server
 node_js:
   - "0.10"
   - "0.12"
--- a/index.js
+++ b/index.js
@@ -29,14 +29,17 @@ try {
     require("./lib/parser/hiredis");
     parsers.push(require("./lib/parser/hiredis"));
 } catch (err) {
+    /* istanbul ignore next: won't be reached with tests */
     debug("hiredis parser not installed.");
 }
 
 parsers.push(require("./lib/parser/javascript"));
 
 function RedisClient(stream, options) {
+    options = options || {};
+
     this.stream = stream;
-    this.options = options = options || {};
+    this.options = options;
 
     this.connection_id = ++connection_id;
     this.connected = false;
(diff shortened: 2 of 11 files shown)


index.js
```diff
@@ -536,20 +539,12 @@ RedisClient.prototype.return_error = function (err) {
...
-    } else {
-        console.log("node_redis: no callback to send error: " + err.message);
-        // this will probably not make it anywhere useful, but we might as well throw
+    try {
+        command_obj.callback(err);
+    } catch (callback_err) {
+        // if a callback throws an exception, re-throw it on a new stack so the parser can keep going
         process.nextTick(function () {
-            throw err;
+            throw callback_err;
```
BridgeAR: It is safe to remove the case that the callback is not a function (I could not reproduce that behavior) as it should never happen. But even if it would happen it is safe to remove the special handling as it will also be catched by the try catch and rethrown async.

index.js
```diff
@@ -628,7 +623,7 @@ RedisClient.prototype.return_reply = function (reply) {
         command_obj = this.command_queue.shift();
     }
 
-    queue_len = this.command_queue.getLength();
+    queue_len = this.command_queue.length;
```
BridgeAR: I prefere writing `.length` as most queues use that and it is already implemented to support it.

lib/parser/hiredis.js
```diff
@@ -27,12 +27,7 @@ HiredisReplyParser.prototype.execute = function (data) {
     var reply;
     this.reader.feed(data);
     while (true) {
-        try {
-            reply = this.reader.get();
-        } catch (err) {
-            this.emit("error", err);
-            break;
-        }
+        reply = this.reader.get();
```
BridgeAR: It is safe to remove this try catch as there's a try catch calling the execute function and emitting the error just as it was done here before (besides the fact that this code block was never reached in our tests).

.travis.yml
```diff
@@ -8,6 +8,8 @@ addons:
     - ubuntu-toolchain-r-test
     packages:
     - g++-4.8
+services:
+  - redis-server
```
BridgeAR: We need another redis instance to test the default connection and I thought that's an easy way to handle this.

bcoe: It seems slightly weird that the tests would now require that we have a redis up and running on a specific port, I liked @erinspice's refactor that made the test-suite stand up the testing server for this purpose.

Perhaps we could make the tests stand up a server on `:6379`, but only do so if there's not already a server running on `:6379`? I'm happy to make this change myself when I'm done the work day.

BridgeAR: :+1:

test/node_redis.spec.js
```diff
@@ -28,6 +49,58 @@ describe("The node_redis client", function () {
...
+                    try {
+                        redis.createClient(true);
+                        throw new Error('failed');
+                    } catch (err) {
+                        assert.equal(err.message, 'unknown type of connection in createClient()');
+                    }
+                });
+
+                if (ip === 'IPv4') {
+                    it('allows connecting with the redis url and the default port', function (done) {
```
bcoe: love these tests you've added :100:

bcoe:
@BridgeAR I pulled in your work, and made the changes to how redis spawns on this thread:

https://github.com/NodeRedis/node_redis/pull/819

Issue #543, opened 2014-01-24
try_callback function makes server extremely busy so that all CPU is occupied by node process 


nickhsine:
Code in index.js

```js
function try_callback(client, callback, reply) {
    try {
        callback(null, reply);
    } catch (err) {
        if (process.domain) {
            process.domain.emit('error', err);
            process.domain.exit();
        } else {
            client.emit("error", err);
        }
    }
}
```

If the callback has a error and the process.domain exists, then my server will be extremely busy and it causes CPU to have almost 100% loading. My server will be hung up there, and it cannot process any new requests.

Can this library avoid accessing the global object(process.domain)? just use redis client emit the error event? And let other programs using the redis client to handle the error event?
Or can anyone help me to avoid that problem?

Thanks in advance.

blainsmith:
In order to better support this project and its new group of collaborators under a new org we are trying to clean up and close issues older than 6 months. Your issue may have been fixed since the issue was open, however, if you are still experiencing a problem **please** re-open this issue and we will label it accordingly.

PR #816, opened 2015-09-03
Remove try_callback

BridgeAR:
I can't see that this would harm anyone

### DIFF
--- a/index.js
+++ b/index.js
@@ -280,6 +280,8 @@ RedisClient.prototype.init_parser = function () {
                 return true;
             }
         })) {
+            // Do not emit this error
+            // This should take down the app if anyone made such a huge mistake or should somehow be handled in user code
             throw new Error("Couldn't find named parser " + self.options.parser + " on this system");
         }
     } else {
@@ -629,10 +631,12 @@ RedisClient.prototype.return_reply = function (reply) {
                 }
                 this.emit(type, reply1String, reply[2]); // channel, count
             } else {
-                throw new Error("subscriptions are active but got unknown reply type " + type);
+                this.emit("error", new Error("subscriptions are active but got unknown reply type " + type));
+                return;
             }
-        } else if (! this.closing) {
-            throw new Error("subscriptions are active but got an invalid reply: " + reply);
+        } else if (!this.closing) {
+            this.emit("error", new Error("subscriptions are active but got an invalid reply: " + reply));
+            return;
         }
     } else if (this.monitoring) {
         len = reply.indexOf(" ");
@@ -643,7 +647,7 @@ RedisClient.prototype.return_reply = function (reply) {
         });
         this.emit("monitor", timestamp, args);
     } else {
-        throw new Error("node_redis command queue state error. If you can reproduce this, please report it.");
+        this.emit("error", new Error("node_redis command queue state error. If you can reproduce this, please report it."));
     }
 };
 
@@ -702,9 +706,16 @@ RedisClient.prototype.send_command = function (command, args, callback) {
 
     // if the value is undefined or null and command is set or setx, need not to send message to redis
     if (command === 'set' || command === 'setex') {
-        if(args[args.length - 1] === undefined || args[args.length - 1] === null) {
+        if (args.length === 0) {
+            return;
+        }
+        if (args[args.length - 1] === undefined || args[args.length - 1] === null) {
             var err = new Error('send_command: ' + command + ' value must not be undefined or null');
-            return callback && callback(err);
+            if (callback) {
+                return callback && callback(err);
+            }
+            this.emit("error", err);
+            return;
         }
     }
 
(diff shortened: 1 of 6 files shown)


bcoe:
@brycebaril and/or @mranney, could you give some context regarding the `try_callback` method. Is this change going to break domains?

BridgeAR:
Reference #543

BridgeAR:
The original commit for this is 202df58b932dadf2843ad3b9b6e44f2fc3e0e703 and it had nothing to do with domains. As I see it the only thing with domains was, that they had to be added added in the try_callback as node-redis catched those errors too. By removing this nothing should be broken at all.

BridgeAR:
The original commit is faulty and this should have never been introduced. It tries to fix wrong user code by throwing the <b>user</b> error async instead of sync.

So this is a bug fix overall without any negative impact on domains.

brycebaril:
Judging by the original PR, the reason this exists is for multiplexed replies such as:

`-Error message\r\n+OK\r\n`

Rethrowing on the next stack allows the parser to continue and handle the `+OK\r\n` message.

BridgeAR:
But those do not happen in the user callback?

brycebaril:
@BridgeAR they happen when there is an error in a multiplexed reply, which are `emit` by the parser synchronously https://github.com/NodeRedis/node_redis/blob/master/lib/parser/javascript.js#L174-L180

particularly the rethrow on https://github.com/NodeRedis/node_redis/blob/master/index.js#L539-L547 is designed to protect this

othiym23:
I think that, worst case, this will require people who want calls into redis to play nice with domains will need to go back to `.bind()`ing their callbacks before handing them off to the driver, just like they would with `continuation-local-storage` or anything else that needs to deal with the way Redis bundles callbacks together for multiplexing.

I'm +0 on this change – I think the way that this code deals with errors is sufficiently opaque that relatively few users of the library understand how to handle errors with it correctly, and this change is likely to leave people in a confused state in at least a few cases, but the patch also removes some complexity.

BridgeAR:
@othiym23 the binding is still in place and working. Your original test is passing just fine. 

@brycebaril I wrote different multiplexed examples now and they are all handled properly and just the same as with the try catch. There was an error that the js parser replied errors as strings that I [already fixed] (https://github.com/NodeRedis/node_redis/commit/52db91c7530cbca333fb47a9f8d2019856f15b63).

Overall I'm very certain that this code was a bug not a feature ;)

index.js
```diff
@@ -739,9 +704,16 @@ RedisClient.prototype.send_command = function (command, args, callback) {
 
     // if the value is undefined or null and command is set or setx, need not to send message to redis
     if (command === 'set' || command === 'setex') {
-        if(args[args.length - 1] === undefined || args[args.length - 1] === null) {
+        if (args.length === 0) {
```
BridgeAR: I'm uncertain if we want to support this behavior but this is how the tests behave right now. I'd say it's very unlikely that any user is going to call `set` / `setex` without any arguments. They might be undefined, but that would result in an error...

index.js
```diff
@@ -739,9 +704,16 @@ RedisClient.prototype.send_command = function (command, args, callback) {
...
+            return;
+        }
+        if (args[args.length - 1] === undefined || args[args.length - 1] === null) {
             var err = new Error('send_command: ' + command + ' value must not be undefined or null');
-            return callback && callback(err);
+            if (callback) {
+                return callback && callback(err);
+            }
+            this.emit("error", err);
```
BridgeAR: If an error is returned to the callback the same behavior should be applied if no callback is present. So this is going to emit an error in that case from now on.

index.js
```diff
@@ -529,38 +517,13 @@ RedisClient.prototype.return_error = function (err) {
...
-    try {
+    if (command_obj.callback) {
         command_obj.callback(err);
-    } catch (callback_err) {
-        // if a callback throws an exception, re-throw it on a new stack so the parser can keep going
-        process.nextTick(function () {
-            throw callback_err;
-        });
+    } else {
+        this.emit('error', err);
```
BridgeAR: Instead of throwing an error into Nirvana it should be emitted

BridgeAR:
This is going to fix four things:

1. This is going to fix issues where errors could have taken down user apps by throwing them async into nirvana. Instead, they will be emitted from now on, so the user is going to be able to catch them appropriately.
2. This is also fixing issues with users making errors in their callback and not catching them properly.
3. Errors do not get wrapped into other errors anymore and loose there trace that way.
4. Using set with an undefined or null value is and no callback is going to emit an error from now on. This is to make the behavior of using a callback / no callback consistent.

This is not breaking the domains as I already tried to outline. That is still in place and working as it should.

This might break users code that tried to work around these issues but they should still be fixed in my opinion. So this should go into the next major version.

BridgeAR:
I edited the comment above and removed the error wrapping into another PR

bcoe:
@BridgeAR I like that this simplifies the error handling logic, moving towards always emitting, and that it gets rid of some opaque code. I say that we release this as part of `v2.0.0`, and let the community point us towards any issues.

index.js
BridgeAR: Ping @bcoe @brycebaril

BridgeAR: Since there are still some open issues about `undefined` and `null` values passed into redis we're going to have to have a closer look at this later on anyway. Therefor I'll merge it as is.

Issue #605, opened 2014-06-15
.toLowerCase is slow


dcharbonnier:
According to this http://jsperf.com/regex-vs-tolowercase toLowerCase is slow.
My test consist on doing 1000 multi each containing 100 sadd (real use-case for me).
Replacing toLowerCase with regexp increase performance by 5%, removing possibility of uppercase (using === string with no uppercase alternative) increase performance by 10%. Is there any good reason for allowing different case for functions ? When people use redis most of the time performance is critical.

blainsmith:
In order to better support this project and its new group of collaborators under a new org we are trying to clean up and close issues older than 6 months. Your issue may have been fixed since the issue was open, however, if you are still experiencing a problem **please** re-open this issue and we will label it accordingly.

BridgeAR:
Reopened as the issue is still present but going to be fixed soon

PR #817, opened 2015-09-03
Tiny speedup by removing command.toLowerCase()

BridgeAR:
This is not necessary as the command itself is only used from inside the code and as they are (now) all lower case it is safe to remove the toLowerCase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -551,9 +551,9 @@ the second word as first parameter:
 
 ## client.send_command(command_name, args, callback)
 
-Used internally to send commands to Redis. For convenience, nearly all commands that are published on the Redis
-Wiki have been added to the `client` object. However, if I missed any, or if new commands are introduced before
-this library is updated, you can use `send_command()` to send arbitrary commands to Redis.
+Used internally to send commands to Redis. Nearly all Redis commands have been added to the `client` object.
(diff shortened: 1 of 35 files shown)


brycebaril:
This is breaks backwards compatibility, correct?

BridgeAR:
Why should it be? I can't think of any way that a user could insert commands upper case that would not be handled.

brycebaril:
Ahh, yes -- that's correct. The upperCase versions of the commands close the lowercase ones and call that.

Not sure we should lowercase all the tests, though, as this would reduce coverage, no? (Unless you added/left one to specifically test the uppercase commands?)

BridgeAR:
I checked that there were always both versions and kept them as they were if there was only one version. But this would actually not matter for the coverage as the commands don't count towards the coverage :)

bcoe:
This seems reasonable to me, while you're at it, it might be worth changing:

https://github.com/fintura/node_redis/blob/speedup/index.js#L1143

On this branch, so that it matches the pattern used elsewhere for exposing the upper and lowercase versions of each method.

BridgeAR:
@bcoe I changed that in another PR, should I move that change into this one?

BridgeAR:
Closing #605

PR #818, opened 2015-09-03
[WIP] Fix returning buffers not working

BridgeAR:


index.js
```diff
@@ -686,6 +686,7 @@ RedisClient.prototype.return_reply = function (reply) {
             throw new Error("subscriptions are active but got an invalid reply: " + reply);
         }
     } else if (this.monitoring) {
+        reply = reply && reply.toString && reply.toString();
         len = reply.indexOf(" ");
```
bcoe: if reply is no t a string type, won't the following lines result in an exception?

BridgeAR: That's why I added the coercion to string if the return type is buffer :) But I'm still uncertain if that's the right thing to do, as the request was to only return buffers instead of strings.

PR #819, opened 2015-09-04
minor tweaks to how we spawn tests in #815

bcoe:
slight modification to https://github.com/NodeRedis/node_redis/pull/815

* we now spawn redis on the default port so that we don't need to run a second copy of redis for tests that exercise node_redis' default behavior.
* we stop the tests and warn the user if they already have a copy of redis up and running on `:6379`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,14 +29,17 @@ try {
     require("./lib/parser/hiredis");
     parsers.push(require("./lib/parser/hiredis"));
 } catch (err) {
+    /* istanbul ignore next: won't be reached with tests */
     debug("hiredis parser not installed.");
 }
 
 parsers.push(require("./lib/parser/javascript"));
 
 function RedisClient(stream, options) {
+    options = options || {};
+
     this.stream = stream;
-    this.options = options = options || {};
+    this.options = options;
 
     this.connection_id = ++connection_id;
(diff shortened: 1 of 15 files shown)


index.js
```diff
@@ -51,26 +54,23 @@ function RedisClient(stream, options) {
...
-    if (typeof this.options.enable_offline_queue === "boolean") {
-        this.enable_offline_queue = this.options.enable_offline_queue;
+    if (this.options.enable_offline_queue === false) {
+        this.enable_offline_queue = false;
     }
     this.retry_max_delay = null;
-    if (options.retry_max_delay !== undefined && !isNaN(options.retry_max_delay) && options.retry_max_delay > 0) {
-        this.retry_max_delay = options.retry_max_delay;
+    if (options.retry_max_delay && options.retry_max_delay > 0) {
+        this.retry_max_delay = +options.retry_max_delay;
```
bcoe: @BridgeAR this change concerns me a bit, why the pre-increment? and should we still confirm that `options.retry_max_delay` is a #?

BridgeAR: The + is only a type coercion to number :)
`a > 0` is also doing a type coercion, so it'll be the same as it was before

PR #823, opened 2015-09-04
Revert to old behavior

BridgeAR:
I thought it is probably best to revert the debug behavior to be in line with the old one until be release a new major version. It is still possible to pass the NODE_DEBUG env though, so the readme is up to date. Just no code is going to break at all (even though it is unlikely that someone read our stderr stream and relied on our debug messages in real code).

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,16 +13,13 @@ var net = require("net"),
     connection_id = 0,
     default_port = 6379,
     default_host = "127.0.0.1",
-    debug;
+    debug = function(msg) {
+        if (exports.debug_mode) {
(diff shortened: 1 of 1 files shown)


bcoe:
personally I like the new debug logic, did any issues float in that make you want to revert this code?

BridgeAR:
@bcoe There was no issue about it but I thought let's not break any code that we do not have to break. It is still possible to activate the debug mode by using the env variable. It's just not possible to activate / deactivate the debug mode by changing the env variable while running the program. Would you like to do this with the env variable as well?

PR #824, opened 2015-09-04
Remove async dependency

BridgeAR:
It's only a devDependency but I did not really see the benefit of it :)

### DIFF
--- a/test/node_redis.spec.js
+++ b/test/node_redis.spec.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var async = require("async");
 var assert = require("assert");
 var config = require("./lib/config");
 var helper = require('./helper');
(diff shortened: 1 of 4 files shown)


test/node_redis.spec.js
```diff
@@ -133,40 +116,40 @@ describe("The node_redis client", function () {
...
-                                });
-                            });
-
-                            client.subscribe("recon channel", function (err, res) {
-                                // Do not do this in normal programs. This is to simulate the server closing on us.
-                                // For orderly shutdown in normal programs, do client.quit()
-                                client.stream.destroy();
-                            });
-                        });
+                        // it("remains subscribed", function () {
```
BridgeAR: This is a broken test

PR #825, opened 2015-09-07
Remove event emitters from the parsers

BridgeAR:
Right now the parser first emit the reply / error and the catched data is then passed to another function.

This is an unneeded overhead as it's possible to call the function without any detour.

### DIFF
--- a/index.js
+++ b/index.js
@@ -302,18 +302,8 @@ RedisClient.prototype.init_parser = function () {
     this.reply_parser = new this.parser_module.Parser({
         return_buffers: self.options.return_buffers || self.options.detect_buffers || false
     });
-
-    // "reply error" is an error sent back by Redis
-    this.reply_parser.on("reply error", function (reply) {
-        self.return_error(reply);
-    });
-    this.reply_parser.on("reply", function (reply) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -302,18 +302,8 @@ RedisClient.prototype.init_parser = function () {
...
-
-    // "reply error" is an error sent back by Redis
-    this.reply_parser.on("reply error", function (reply) {
-        self.return_error(reply);
-    });
-    this.reply_parser.on("reply", function (reply) {
-        self.return_reply(reply);
-    });
-    // "error" is bad.  Somehow the parser got confused.  It'll try to reset and continue.
-    this.reply_parser.on("error", function (err) {
```
BridgeAR: This is dead code

lib/parser/hiredis.js
```diff
@@ -27,21 +22,16 @@ HiredisReplyParser.prototype.execute = function (data) {
     var reply;
     this.reader.feed(data);
     while (true) {
-        try {
-            reply = this.reader.get();
-        } catch (err) {
-            this.emit("error", err);
```
BridgeAR: This is dead code

Issue #689, opened 2014-12-05
MULTI EXEC will not return error when some of the commands fails


bitinn:
It took me a long time to figure out why: when some of queued commands fail to execute, Redis server send reply as Bulk Strings, and in both node_redis javascript parser and hiredis wrapper, they are emitted as `reply`, not `reply error`.

Thus in exec callback, `err` will be null, `res` will be an array containing all replies, both ok and error. eg.

```
[ 'OK',
  'WRONGTYPE Operation against a key holding the wrong kind of value',
  'WRONGTYPE Operation against a key holding the wrong kind of value' ]
```

(note the errors are returned as string, not wrapped in `Error`, so no easy way to determine whether MULTI EXEC are run successfully)

I am not sure the best way to fix this, changing how node_redis parser handle this type of responses will probably cause it to diverge with hiredis behaviour...

But at the very least we should [document it](https://github.com/mranney/node_redis#multiexec-callback-), so issues don't come up again and again, like https://github.com/mranney/node_redis/issues/624 & https://github.com/mranney/node_redis/issues/550.

What do you guys think? @brycebaril @mranney @mocheng @bsnote @guileen (apologize for the mass notice if you are not interested)

Reference:

- http://redis.io/topics/transactions - see the end of *Errors inside a transaction* section, where it explicitly states, *It's up to the client library to find a sensible way to provide the error to the user.*
- http://redis.io/topics/protocol#bulk-string-reply
- https://github.com/mranney/node_redis/blob/master/lib/parser/javascript.js#L190

Guuz:
I ran into the same problem and was about to file an issue when i saw this one.
Any tips on how to code around this? I'm not sure how to do my error handling now and how to assure the command(s) executed successfully.

Semi-related:
If you did actually get an array of errors, what should i do with them? My higher function would probably expect an error object, not an array of errors. This is not standard.

brycebaril:
For now the best advice I can offer is don't have this problem. It's lame, I know, but one nice thing about Redis is the only errors are operations that should be avoidable, e.g. bad command syntax or mismatching commands and types. Any time you do get an Error from Redis should likely be considered a bug in your use of Redis, and an invalid application.

Of course that isn't the long-term solution because we need our applications to be robust in the face of uncertainty. Beyond that it's going to require working around this quirk in the Redis protocol and inspecting the replies.

Guuz:
[...]

PR #826, opened 2015-09-07
Fix multi error handling

BridgeAR:
This is only optimizing the error handling to be consistent as mentioned in #689. It would be great to get some replies about the behavior as I implemented it here, as there might be room for further optimizations.

<s>It seems like the js parser is not in line with the redis parser though (errors are returned as replies) and therefor one test is going to fail with the js parser. I did not dig into fixing this yet.</s>

### DIFF
--- a/index.js
+++ b/index.js
@@ -389,7 +389,6 @@ RedisClient.prototype.ready_check = function () {
 RedisClient.prototype.send_offline_queue = function () {
     var command_obj, buffered_writes = 0;
 
-    // TODO: Implement queue.pop() as it should be faster than shift and evaluate petka antonovs queue
     while (command_obj = this.offline_queue.shift()) {
         debug("Sending offline command: " + command_obj.command);
         buffered_writes += !this.send_command(command_obj.command, command_obj.args, command_obj.callback);
@@ -995,15 +994,15 @@ Multi.prototype.hmset = Multi.prototype.HMSET = function (key, args, callback) {
     return this;
 };
 
-Multi.prototype.send_command = function (command, args, cb) {
+Multi.prototype.send_command = function (command, args, index, cb) {
     var self = this;
     this._client.send_command(command, args, function (err, reply) {
         if (err) {
             if (cb) {
                 cb(err);
-            } else {
-                self.errors.push(err);
             }
+            err.position = index - 1;
+            self.errors.push(err);
         }
     });
 };
@@ -1029,7 +1028,7 @@ Multi.prototype.exec = Multi.prototype.EXEC = function (callback) {
                 break;
             }
         }
-        this.send_command(command, args, cb);
+        this.send_command(command, args, index, cb);
     }
 
     this._client.send_command('exec', [], function(err, replies) {
@@ -1042,9 +1041,10 @@ Multi.prototype.execute_callback = function (err, replies) {
 
     if (err) {
         if (err.code !== 'CONNECTION_BROKEN') {
+            err.code = 'EXECABORT';
+            err.errors = this.errors;
             if (this.callback) {
-                this.errors.push(err);
-                this.callback(this.errors);
+                this.callback(err);
             } else {
                 // Exclude CONNECTION_BROKEN so that error won't be emitted twice
                 this._client.emit('error', err);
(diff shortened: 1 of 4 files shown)


BridgeAR:
The js parser error got fixed by #751, so all tests are passing again.

index.js
```diff
@@ -638,7 +638,7 @@ RedisClient.prototype.return_reply = function (reply) {
                 reply = reply_to_object(reply);
             }
 
-            try_callback(command_obj.callback, reply);
+            command_obj.callback(null, reply);
```
BridgeAR: The try callback is actually preventing us from emitting the error properly, so this has to be removed for it.

index.js
```diff
@@ -1095,10 +1098,15 @@ Multi.prototype.exec = function (callback) {
...
                 }
             }
+        } else if (err) {
+            err.code = 'EXECABORT';
+            err.errors = errors;
         }
 
         if (callback) {
-            callback(null, replies);
+            callback(err, replies);
```
BridgeAR: I'm uncertain if this should stay like that. Likely not. It might be nice to see what commands passed but this might be confusing if both is returned at the same time.

bcoe: On newer redis' is it possible to have a multi response that contains errors and replies, I thought the transaction is aborted -- so you wouldn't get anything.

Is there a scenario where we would get both errors and replies?

bcoe:
This seems like a reasonable patch to land in a `v2.0.0` release to me, but I'd like to better understand if there are scenarios where we could receive both errors and results in a callback, this seems quite weird to me.

index.js
BridgeAR: I read the [transactions] (http://redis.io/topics/transactions) part again and it seems like this is a special behavior by redis.
The transaction is only aborted if any command can't be queued. If the rough syntax is fine and redis queues the commands without any error all commands are going to be executed. If any of these commands fail it'll return an error individually. But this is not going to abort the transaction. So all commands that got executed are indeed executed in the order of the occurrence.

Earlier you got an array of errors if a command could not be queued. This is changed to be a single error with the `EXECABORT` code and the failing command as element of `err.errors`.
If a single command could not be executed this would not have returned an error in the error return value but in the reply. 
That way it's not obvious that commands could not be executed. And since commands will only fail because of programmer errors this should result in an error instead. So I changed it to return both: the err, to indicate that something went wrong with all not executed commands inserted in `err.errors` including the execution position and the reply with all returned execution values that passed and undefined values for errors.
It might be an idea to keep the errors in the reply as well instead of passing undefined here. What do you think @bcoe @brycebaril @bitinn?

test/commands/multi.spec.js
```diff
@@ -225,18 +225,58 @@ describe("The 'multi' method", function () {
...
+
+                it('reports multiple exceptions when they occur (while EXEC is running) and calls cb', function (done) {
+                    if (!helper.serverVersionAtLeast(client, [2, 6, 5])) return done();
+
+                    var multi = client.multi();
+                    multi.config("bar", helper.isError());
+                    multi.debug("foo").exec(function (err, reply) {
+                        assert.equal(err.code, "MULTIERROR");
+                        assert(err.message.match(/^EXEC/), "Error message should begin with EXEC");
+                        assert.equal(err.errors.length, 2, "err.errors should have 2 items");
```
BridgeAR: TODO: Add another command (e.g. set) that is going to be executed properly and check that the value is persistent.

test/commands/multi.spec.js
```diff
@@ -198,18 +198,55 @@ describe("The 'multi' method", function () {
...
+                it('reports multiple exceptions when they occur (while EXEC is running) and calls cb', function (done) {
+                    var multi = client.multi();
+                    multi.config("bar", helper.isError());
+                    multi.set('foo', 'bar', helper.isString('OK'));
+                    multi.debug("foo").exec(function (err, reply) {
+                        assert.equal(err.code, "MULTIERROR");
+                        assert(err.message.match(/^EXEC/), "Error message should begin with EXEC");
+                        assert.equal(err.errors.length, 2, "err.errors should have 2 items");
+                        assert(/^ERR/.test(err.errors[0].message), "Actuall error message should begin with ERR");
+                        client.get('foo', helper.isString('bar', done));
```
BridgeAR: As you can see here, the value is indeed persistent even though an error occured.

index.js
BridgeAR: After thinking about it further I'd say it is probably the best to return either err or res. Not both. Therefor if the result is returned it might contain errors as well. But at least they are indeed errors and not strings as they were before.
@bcoe what would you say?

BridgeAR:
@bcoe there will only be either a error (transaction aborted) or an array of results including errors occuring while executing the commands (all passing commands will be persistent). I described the new implementation in more detail in #689

Issue #587, opened 2014-05-11
Doesn't timeout properly when server is unreachable


celesteking:
```
 var r = require('redis').createClient(null, '1.2.3.4', {connect_timeout: 1000}); r.on('ready', function(){ console.log('ready!'); }).on('error', function(err){ console.log('error: ' + err)});
```
It doesn't timeout after 1 second. It does timeout after unknown amount of time with:
> error: Error: Redis connection to 1.2.3.4:6379 failed - connect ETIMEDOUT

I think it should follow the docs and error out after specified timeout.

michelsalib:
I also got the same issue, my config is the following:
```js
var client = redis.createClient(conf.port, conf.host, {
    connect_timeout: 100,        // abandon connection after 100ms
    retry_max_delay: 100,        // no impact, will not wait more than 100ms between reconnections attemps
    max_attempts: 1,             // only 1 connection attempt
    enable_offline_queue: false, // no offline queue, must wait for online mode
    no_ready_check: true         // we don't check for redis ready state, let's query the redis directly
});
```

But instead of failing after 100ms, it fails after a full second. Anyone found a fix for that?

bgSosh:
Similar issues here.

raviv:
+1

chrisbaldauf:
+1

bgSosh:
Any news on this?

chrisbaldauf:
I wound up switching to [ioredis](https://github.com/luin/ioredis), which appears to handle the connection timeouts properly and aside from the client creation, be a drop in replacement for my use case.

bgSosh:
That looks interesting - I'll take a closer look, thanks.

BenHall:
To reproduce I paused a Redis Docker container but still seeing long timeouts

jbergknoff:
Is this issue resolved? I experience the behavior described above. The connection fails after about 2 minutes, despite `connect_timeout` of 5000. I am using version 2.2.5.

BridgeAR:
@jbergknoff this should definitly be resolved. Do you have a reproducable case? And how do your current options look like?

jbergknoff:
Hm, interesting. A slightly modified code snippet from the original post here reproduces the issue:

```javascript
 var r = require('redis').createClient({host: '1.2.3.4', connect_timeout: 1000}); r.on('ready', function(){ console.log('ready!'); }).on('error', function(err){ console.log('error: ' + err)});
```

[...]

Issue #566, opened 2014-03-14
redis.createClient: max_attempts seems to be ignored


ClaudioAlbertin:
I use the max_attempts option in pretty much the most straight-forward way possible:

```javascript
process.on('uncaughtException', function (err) {
  console.log(err.toString());
});

var client = redis.createClient({
  host: '127.0.0.1'.
  port: 6379,
  max_attempts: 1
});
```

However, if no redis server is running and the client can't connect, it keeps on trying, instead of giving up after the specified number of attempts:

```
$ node app
Error: Redis connection to 127.0.0.1:[object Object] failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:6379 failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:6379 failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:[object Object] failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:[object Object] failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:6379 failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:6379 failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:[object Object] failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:6379 failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:[object Object] failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:[object Object] failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:6379 failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:[object Object] failed - connect ECONNREFUSED
Error: Redis connection to 127.0.0.1:6379 failed - connect ECONNREFUSED
…
```

It might be worth noting that the client was never connected to the server in the first place in this scenario. This is the first connection it attempts to establish.

I'm also somewhat confused by the doubled error messages, once for `127.0.0.1:[object Object]` and once for `127.0.0.1:6379` that seem to be thrown at about the same time, for each failed connection attempt.

blainsmith:
In order to better support this project and its new group of collaborators under a new org we are trying to clean up and close issues older than 6 months. Your issue may have been fixed since the issue was open, however, if you are still experiencing a problem **please** re-open this issue and we will label it accordingly.

BridgeAR:
[...]

PR #829, opened 2015-09-10
Implement redis connection broken mode

### LINKED ISSUE #569: poor default for retry_max_delay
doubling without a cap is not reasonable IMO, even a large cap would be a much saner default

BridgeAR:
This is also going to change some small things like adding a maximum timeout.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -17,7 +17,7 @@
     "mocha": true,
 
     // Relaxing options
-    "boss": true, // Accept things like `while (command = keys.shift()) { ... }`
+    "boss": true, // Accept statements like `while (key = keys.pop()) {}`
 
     "overrides": {
         "examples/*.js": {
--- a/README.md
+++ b/README.md
@@ -197,8 +197,8 @@ to maximum value, provided in milliseconds.
 * `connect_timeout` defaults to `86400000`. Setting `connect_timeout` limits total time for client to reconnect.
 Value is provided in milliseconds and is counted once the disconnect occured. The last retry is going to happen exactly at the timeout time.
 That way the default is to try reconnecting until 24h passed.
-* `max_attempts` defaults to `null`. By default client will try reconnecting until connected. Setting `max_attempts`
-limits total amount of reconnects. Setting this to 0 will prevent any reconnect tries.
+* `max_attempts` defaults to `0`. By default client will try reconnecting until connected. Setting `max_attempts`
+limits total amount of connection tries. Setting this to 1 will prevent any reconnect tries.
 * `auth_pass` defaults to `null`. By default client will try connecting without auth. If set, client will run redis auth command on connect.
 * `family` defaults to `IPv4`. The client connects in IPv4 if not specified or if the DNS resolution returns an IPv4 address.
 You can force an IPv6 if you set the family to 'IPv6'. See nodejs net or dns modules how to use the family type.
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ to maximum value, provided in milliseconds.
 Value is provided in milliseconds and is counted once the disconnect occured. The last retry is going to happen once after the connect_timeout value is exceeded.
 That way the default is to try reconnecting until at least 24h passed.
 * `max_attempts` defaults to `null`. By default client will try reconnecting until connected. Setting `max_attempts`
-limits total amount of reconnects.
+limits total amount of reconnects. Setting this to 0 will prevent any reconnect tries.
 * `auth_pass` defaults to `null`. By default client will try connecting without auth. If set, client will run redis auth command on connect.
 * `family` defaults to `IPv4`. The client connects in IPv4 if not specified or if the DNS resolution returns an IPv4 address.
 You can force an IPv6 if you set the family to 'IPv6'. See nodejs net or dns modules how to use the family type.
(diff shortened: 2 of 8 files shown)


BridgeAR:
This branch is going to fix a lot of issues for quite some people.

BridgeAR:
Fixes #569, #587, #566 and #280.
This includes the fix as suggested in #671 and #533
Closes #675, #615, #463 and #724

There will be more closed tickets that we closed without fixing that get fixed by this PR.

index.js
```diff
@@ -1063,7 +1051,7 @@ Multi.prototype.exec = function (callback) {
 
     // TODO - make this callback part of Multi.prototype instead of creating it each time
     return this._client.send_command("exec", [], function (err, replies) {
-        if (err) {
+        if (err && !err.code) {
```
BridgeAR: This is a preparation for the multi fixes.

.jshintrc
```diff
@@ -16,6 +16,9 @@
     "node": true, // Enable globals available when code is running inside of the NodeJS runtime environment.
     "mocha": true,
 
+    // Relaxing options
+    "boss": true, // Accept statements like `while (key = keys.pop()) {}`
```
BridgeAR: I personally like to use this. I could imagine though that not everyone likes it?

README.md
```diff
@@ -188,10 +193,11 @@ with an error, or an error will be thrown if no callback is specified.
...
 reconnection (delay) almost doubles. This delay normally grows infinitely, but setting `retry_max_delay` limits delay
 to maximum value, provided in milliseconds.
-* `connect_timeout` defaults to `false`. By default client will try reconnecting until connected. Setting `connect_timeout`
-limits total time for client to reconnect. Value is provided in milliseconds and is counted once the disconnect occured.
-* `max_attempts` defaults to `null`. By default client will try reconnecting until connected. Setting `max_attempts`
-limits total amount of reconnects.
+* `connect_timeout` defaults to `86400000`. Setting `connect_timeout` limits total time for client to reconnect.
+Value is provided in milliseconds and is counted once the disconnect occured. The last retry is going to happen exactly at the timeout time.
+That way the default is to try reconnecting until 24h passed.
+* `max_attempts` defaults to `0`. By default client will try reconnecting until connected. Setting `max_attempts`
```
BridgeAR: The new default should not break any old code, since `null` would behave just the same as `0`.

README.md
```diff
@@ -188,10 +193,11 @@ with an error, or an error will be thrown if no callback is specified.
 * `retry_max_delay`: defaults to `null`. By default every time the client tries to connect and fails time before
 reconnection (delay) almost doubles. This delay normally grows infinitely, but setting `retry_max_delay` limits delay
 to maximum value, provided in milliseconds.
-* `connect_timeout` defaults to `false`. By default client will try reconnecting until connected. Setting `connect_timeout`
-limits total time for client to reconnect. Value is provided in milliseconds and is counted once the disconnect occured.
-* `max_attempts` defaults to `null`. By default client will try reconnecting until connected. Setting `max_attempts`
-limits total amount of reconnects.
+* `connect_timeout` defaults to `86400000`. Setting `connect_timeout` limits total time for client to reconnect.
```
BridgeAR: This could be split into another PR but I'd like to keep it all together

README.md
```diff
@@ -575,12 +581,12 @@ some kind of maximum queue depth for pre-connection commands.
 
 ## client.retry_delay
 
-Current delay in milliseconds before a connection retry will be attempted. This starts at `250`.
+Current delay in milliseconds before a connection retry will be attempted. This starts at `200`.
```
BridgeAR: I think starting with a lower delay might be good and we might actually also consider lowering the multiplier. What do you all think?

test/auth.spec.js
```diff
@@ -12,10 +12,11 @@ describe("client authentication", function () {
         });
     });
 
-    helper.allTests(function(parser, ip, args) {
+    helper.allTests({
+        allConnections: true
+    }, function(parser, ip, args) {
```
BridgeAR: The test suite has a new default to only run each test with each parser once instead of each parser * each connection type (IPv4, IPv6 and socket). This reduces unnecessary overhead. To run each connection type this optional option may be passed to the helper.

README.md
BridgeAR: A whole day as a maximum delay might already be to much.

BridgeAR:
This is going to do the following:

1. [FIX?] Change the default connection timeout to be one day instead of never.
2. [FIX] There won't be any connection retries after the connection timeout anymore. This was the case earlier because the timeout for the next retry is increasing each retry by a multiplier. If the next retry would be above the connection timeout it is now going to be exactly at the maximum retry connection timeout instead. So the timeout is follows as expected from now on. There
3. [??] The default minimum connection retry time is reduced a tiny bit (200 ms instead of 255).
4. [FIX] The queues won't be flushed anymore if node_redis is trying to reconnect. It will only be flushed if the connection is broken or if the user closes the connection. So offline_queue and max_attempts / connection_timeout is now properly working together.
5. [FEATURE] If no connection could be established and max_attempts / connection_timeout has been exceeded node_redis is going into the "connection is broken" mode and emit an error.

README.md
bcoe: `0` seems a bit weird to me, to me this feels like:

`don't try to reconnect`.

How would I tell the client not to reconnect?

BridgeAR: I thought so too. But this is actually just a stylistic change, since passing 0 earlier behaved just the same, it was just not documented. So I thought it's probably best to change `null` to `0` and document it.

> How would I tell the client not to reconnect?

Have a look at the next line :D (entering 1 does that - it's about the connection attempts in total not the reconnect attempts. While fixing the max_attempts I first switched it to be the number of reconnects and therefor `0` would have been `don't try to reconnect` but this is not represented by the variable name and did not behave like that before. So this seems to be legit).

BridgeAR:
This should be ready to merge

Issue #821, opened 2015-09-04
Handle enabling/changing redis password without application restart


epoberezkin:
It seems to be not possible to enable or change redis password without application restart.

Even in case when I want to enable password and I supply password to the driver (it logs warning that the password is not required), when the password is enabled in redis, the driver returns error "not authenticated"; it doesn't try to authenticate using password that was passed in config.

For changing password there is no option. It would not be required if redis supported user accounts - you'd be able to simply create another account, switch application to use it and then remove the old account. Redis doesn't and apparently will not have accounts.

Of course the issue can be resolved by creating a wrapper around redis driver that would create a new client when authentication fails, but it seems as either too much of a change in applications or too much of an effort to make the wrapper that behaves exactly as the driver...

Maybe there is a simpler solution to this problem, like subscribing to some event, authenticating and replaying the last command when authentication fails? Please let me know if there is.

Otherwise it would be nice to have these two features inside driver:

1. Handle enabling password:
   1) If the password is passed
   2) and it was not required by redis
   3) and at some later time some command was rejected as not authenticated
   4) driver will try authenticating again using the passed password
   5) if authentication is successful, it will resend command
   6) if authentication failed it would return error

2. Support feature of password change using new_password option (or something like this). In this case when authentication fails, the driver will try authenticating using the new password and if it was successful the driver will use it afterwards in case of re-connects.

If you are OK with these features I could implement.

Please let me know.
Thank you.

Nevon:
This feature, or something similar, is required in order to support temporary refreshable credentials, such as [AWS IAM authentication](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/auth-iam.html). In that authentication mechanism, the user creates a token that's valid for 15 minutes, which is used as the password in the AUTH command. As the token is only valid for 15 minutes, the client needs to be able to pick up a new token for then next authentication request. Furthermore, the client needs to periodically reauthenticate by sending a new AUTH request - however, I suspect this can be accomplished on the application layer.

Rather than passing some `new_password` property, I would suggest making the `password` option be:

```ts
type Provider<T> = () => T | Promise<T>
type password = string | Provider<string>
```

And have the driver look up the password on every auth command. This is similar to how it is implemented in [the java redis client Lettuce](https://github.com/lettuce-io/lettuce-core/pull/1916), for example.

leibale:
@Nevon good idea! but maybe instead of doing it for `password` only we should do it for the whole `options` object? i.e.
[...]

Issue #830, opened 2015-09-10
Multiple connection strings for master-slave demotion/promotion.


robblovell:
On ElasticCache, a redis cluster has one primary master read/write node and several secondary slave read-only nodes.  The redis cluster periodically switches roles for nodes, exchanging master and slave nodes. The connection string to the ElasticCache is to the primary node when writes are needed. The createClient should be modified to take more than one connection port and host pair, and try each one in turn when an error like "Can't save packet:Error: READONLY You can't write against a read only slave." occurs so that the client can recover if the primary is switched while in operation.

epoberezkin:
Can be related to #821 - essentially multiple connection settings to support any kind of switching in redis

Mattsi-Jansky:
We have also been bitten by this. Compose's Redis distribution is doing some maintenance soon, and will be failing back to their slave Redis instance during the outage. They recommend using a `host:port,host:port` connection string to maintain availability, but it isn't supported.

samuelt1:
if I made a fork for this would you someone even consider approving it?

stockholmux:
humm.. @Mattsi-Jansky and @samuelt1 could you use the solution outlined #1050 ?

If I'm understanding the solution of just specifying multiple connection strings, I can see cases where not everyone would want this.

samuelt1:
yeah, so that is how I am fixing this on my end. I just have my own function that will connect to all of the servers, and then drop the slaves but not the master.

PR #831, opened 2015-09-10
support enabling password in redis …

epoberezkin:
... after connection is already established without application restart, issue #821

Please let me know if you think this is the right approach.
I will then implement the second part - support changing redis password.

Thank you.


epoberezkin:
I will look into it... It is passing locally though.

epoberezkin:
The test was failing because I had a different error message in my local redis version when authentication failed. It is all fixed and working now.

epoberezkin:
Both parts of issue #821 are now implemented.
It would be great if you could merge it...

epoberezkin:
@BridgeAR Hi Ruben. It would be great if you could merge it and push to npm... All the tests pass, not sure why coveralls shows as pending. We really need it in another package (connect-redis) and using forks is a bad option because we have many apps that use it...

BridgeAR:
@epoberezkin I only had a glimpse at the code earlier. I played around with it and it seems to me that this is a very special case and it should likely be handled in your code. I changed a tiny bit in your test code so it works just as you want it to behave. Have another [look](https://github.com/fintura/node_redis/commit/613bd8560b3c154c9c4a409ca395bc2bec035368#diff-a1fed7aea3031eaf98ad35b11327780eR88) at it.
So a solution as I see it might be that you use a small helper function to do what you wish?

But maybe the others are of other opinion. And I'm way to tired to think properly anymore. It's very early in the morning  here.

epoberezkin:
TL;DR Changing redis password without disrupting the application is a very common need for us and many others. Doing it outside of driver is a real pain, whichever way you do it.

That may be not a situation that happens often, but it's a very common need to change redis password (depending on the security policy you may need to do it on a regular basis) and it is really difficult to manage within software as software can't really control the change of redis password, it can only adapt to it.

The test simply mimics what happens outside of the application. In reality the change of password is done manually, from redis-cli. The reason for that is that multiple apps are using the same redis, so none of them can initialte the change of the password without breaking the other. For example, one app uses connect-redis to store sessions and another app accesses those sessions using this redis client.

The difficulty of handling it in the code is that every command in its callback would need to handle the password change. So the only other solution apart from implementing the feature in the driver is to use the wrapper that would implement all commands and messages that redis driver does. It's much more complex than the change in the driver and also has performance cost.

Given that we have more than 10 redis instances and roughly twice more node apps that use them it is a real pain to manage it without this change.

jasoniangreen:
:+1:

BridgeAR:
After giving this another thought I do see the need for it but I'd prefer a more generic way of handling something like this.

I thought about adding a "callFunctionOnError" option that you'd pass a specific error too as parameter and that function would be called if that error occurs. That way it should be possible to implement this but also a lot of other things. What do you think @epoberezkin ? :)

epoberezkin:
@BridgeAR I think it would do the trick as long as the failed command will be retried after the error is handled. But maybe it's a bit too generic solution. I'd prefer if it was managed via configuration. How do feel about covering both this issue and #830 with supplying array of connection settings (including host, port and password) that would all be tried in case of certain errors (like read-only, not authenticated and maybe some others)?

epoberezkin:
@BridgeAR it would be really good to decide what we do about this features... Not that we are in a great rush but we really need it for several projects. Maybe we could merge this? (I would resolve conflicts then). Or I can implement multiple connection settings from #830 that is a bit more generic. Probably there should still be baseline settings and the array of settings that would extend the baseline.

BridgeAR:
@epoberezkin sry, somehow your reply slipped through. This week is a bit busy for me and if this could still wait a few more days that would be great.

epoberezkin:
Thanks for the reply. I can implement myself, I only need some decision about what to do :) By a few days is not a problem of course.

epoberezkin:
@BridgeAR hi, sorry for chasing, any update on this?

BridgeAR:
I'll invest some time into it later today.

epoberezkin:
thanks!

BridgeAR:
@epoberezkin I'm working on a suggestion and I'll have it done tomorrow. Either we go for that, if you like that suggestion or you could implement the multiple connection settings otherwise. That way we could see what fits bests? :)

epoberezkin:
cool, looking forward for it

BridgeAR:
@epoberezkin it took me a while to get to it, sry. I played around with implementing a generic function and it works quite nice in the way I thought about it (you can provide any commands in that function and if you want to you can repeat the original call when ever you want) but I could not fix all your wishes with it. And since it's really very generic you can also do a lot of stuff that might have side effects that someone did not think about. So you might be right that a more specific solution is the better one here.

Would you give a pull request for multiple connections a shot? :)

epoberezkin:
@BridgeAR No problem. I will, definitely do. Just to confirm, the syntax for options object I suggest is this:

```
failover: {
  connections: [
    host: '...'
    port: '...'
    auth_pass: '...'
  ]
}
```

This way some additional options can be added inside failover in the future. If any of connection parameters are not passed in some of connections, the main connection parameters are used. Also, regardless whether the main connection settings are included inside failover or not, driver should be able to switch back to them. Also if no connection parameters are used on the top level in options it should use parameters from inside failover option... Something like this.

What do you think?

BridgeAR:
I like it. It's likely a the best way to have mutiple alternatives. I'm uncertain about not passing any connection parameters but that's more of a detail in my opinion.

epoberezkin:
Ok, thanks, I'll work on it some time this week

BridgeAR:
@epoberezkin I guess you were busy so far? :)

epoberezkin:
@BridgeAR It's slower than I'd like to, but I hope to finish it this week :) I've just finished options and general failover utils. If you're interested you can see the progress in the branch https://github.com/MailOnline/node_redis/commits/failover (I will squash all commits to master once it's working and make a PR with one commit). But it's far from finished still :)

I actually have a question regarding `client.retry_totaltime`. It seems like every time client tries to reconnect this property is increased. But once reconnection succeeds it is not reset to 0. I think it should be reset, as otherwise after some number of connection losses the driver will potentially not try to reconnect for a long enough time because this property will become equal to the allowed maximum. Of course with the default settings it may not happen soon... But with failover support the reconnection becomes a more normal thing so it will be more likely.

I can fix it within the same PR if you agree (reset `retry_totaltime` to 0 after successful reconnect). Please let me know what you think, I may be missing something.

epoberezkin:
Sorry, disregard the question, I can see it is rest :)

epoberezkin:
closing it now, there is the new PR #895

Issue #629, opened 2014-07-17
Error occured when the internal send_command function was called without the 'args' parameter.


dreamlover:
In the "client.send_command(command_name, args, callback)" section of the read me, it says "
All commands are sent as multi-bulk commands. args can either be an Array of arguments, or omitted. ". But when I call that function with the args really omitted, I got the following error: 
---------------------------------
        throw new Error("send_command: second argument must be an array");
              ^
Error: send_command: second argument must be an array
    at RedisClient.send_command (e:\Work\SGLM\Server\GameServer\SRC\node_modules\redis\index.js:771:15)
    at Object._stopServer [as stopServer] (e:\Work\SGLM\Server\GameServer\SRC\data\redis.js:27:8)
    at Object._stopServer [as stopServer] (e:\Work\SGLM\Server\GameServer\SRC\data\portal.js:12:11)
    at Array.2 (e:\Work\SGLM\Server\GameServer\SRC\sglm.js:49:37)
    at ReadStream.__onData (e:\Work\SGLM\Server\GameServer\SRC\common\CConsoleCommand.js:63:19)
    at ReadStream.EventEmitter.emit (events.js:95:17)
    at ReadStream.<anonymous> (_stream_readable.js:745:14)
    at ReadStream.EventEmitter.emit (events.js:92:17)
    at emitReadable_ (_stream_readable.js:407:10)
    at emitReadable (_stream_readable.js:403:5)

blainsmith:
In order to better support this project and its new group of collaborators under a new org we are trying to clean up and close issues older than 6 months. Your issue may have been fixed since the issue was open, however, if you are still experiencing a problem **please** re-open this issue and we will label it accordingly.

PR #835, opened 2015-09-12
Remove send_command safety checks

BridgeAR:
This checks are only important for users who use send_command directly instead of using the convenience method.
As the readme clearly states how send_command should work and any user would have run into errors if misused, these checks can be removed. If any user might misuse the function anyway, it is very likely that another error will be thrown because of that

### DIFF
--- a/index.js
+++ b/index.js
@@ -661,37 +661,26 @@ function Command(command, args, sub_command, buffer_args, callback) {
 }
 
 RedisClient.prototype.send_command = function (command, args, callback) {
-    var arg, command_obj, i, elem_count, buffer_args, stream = this.stream, command_str = "", buffered_writes = 0, last_arg_type;
+    var arg, command_obj, i, elem_count, buffer_args, stream = this.stream, command_str = "", buffered_writes = 0;
 
-    if (typeof command !== "string") {
-        throw new Error("First argument to send_command must be the command name string, not " + typeof command);
+    // if (typeof callback === "function") {}
+    // probably the fastest way:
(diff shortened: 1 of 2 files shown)


BridgeAR:
This is no BC btw, since if a user would have used the function wrong before, an error would have been thrown. So there should not be any code that relies on this behavior.

BridgeAR:
This could be skipped in the 2.0 release but I think it could be included as well.

bcoe:
I don't know if I agree with this change from a usability standpoint, at npm we're in the habit of asserting at the top of methods that the function signature is correct:

https://github.com/ceejbot/numbat-emitter/blob/master/lib/emitter.js#L16

This is a good way to give immediate feedback to a user that they're using the API in the wrong way. What's the argument for removing the safety checks? I could see how it could benefit performance slightly, did you run any benchmarks?

BridgeAR:
@bcoe using send_command directly should not often be the case but all other users and commands run through these safety checks. It should not impact the speed a lot (I did not run any benchmark as I could not think of a proper way to write a benchmark that would only test these statements. The multi_bench benchmark is not reliable on (the output varies a lot by each run and v8 is probably optimizing a lot away) but it should be a tiny bit.
I tried out all possibilities that we prevent users from doing wrong and all are going to throw errors at some point.
As this makes the code better to read and is a small speedup in most cases I think it's reasonable to merge this.

BridgeAR:
This should fix #629 
TODO: Add a test for a case where omitting is fine

PR #837, opened 2015-09-13
retrofit tests so that they work on Windows

bcoe:
* added appveyor configuration (a Windows CI server).
* hiredis is now installed as an optional dev dependency (since Windows machines have trouble building it with node-gyp).
* auth-spec is now skipped if we fail to spawn a Redis (in our Windows CI we assume a Redis is provisioned by the CI environment, not that we can spawn it using a `redis` bin).
* Windows tests run on an older Redis, I fixed up some of the tests that were conditional based on which Redis version was being used.
* found a single test that behaves weirdly on Windows, and added a  TODO (relates to an error being returned even though the operation should simply be enqueued).


package.json
```diff
@@ -10,20 +10,21 @@
...
-    "test": "./node_modules/.bin/jshint * && nyc ./node_modules/.bin/_mocha ./test/*.js ./test/commands/*.js ./test/parser/*.js --timeout=8000",
-    "jshint": "./node_modules/.bin/jshint *"
+    "coverage": "nyc report --reporter=text-lcov | coveralls",
+    "test": "nyc ./node_modules/.bin/_mocha ./test/*.js ./test/commands/*.js ./test/parser/*.js --timeout=8000",
+    "pretest": "optional-dev-dependency hiredis",
+    "posttest": "jshint ."
   },
   "devDependencies": {
     "coveralls": "^2.11.2",
-    "hiredis": "^0.4.1",
```
BridgeAR: node_hiredis states that it [supports windows since v.0.3.0] (https://github.com/redis/hiredis-node/blame/master/README.md#L77)? Do you have any idea why it does not work?

package.json
```diff
@@ -10,20 +10,21 @@
   "license": "MIT",
   "main": "./index.js",
   "scripts": {
-    "coveralls": "nyc report --reporter=text-lcov | coveralls",
-    "coverage": "nyc report --reporter=html",
```
BridgeAR: I use this quite a lot, so for convenience I'd like to keep this in here :)

test/commands/eval.spec.js
```diff
@@ -31,30 +26,37 @@ describe("The 'eval' method", function () {
             });
 
             it('converts a float to an integer when evaluated', function (done) {
+                helper.serverVersionAtLeast.bind(this)(client, [2, 5, 0]);
```
BridgeAR: .call?

test/commands/expire.spec.js
```diff
@@ -24,7 +24,7 @@ describe("The 'expire' method", function () {
                 client.EXPIRE(["expiry key", "1"], helper.isNumber(1));
                 setTimeout(function () {
                     client.exists(["expiry key"], helper.isNumber(0, done));
-                }, 1100);
+                }, 3000);
```
BridgeAR: Was this an issue on windows? Redis should never take that long to execute any command :/

test/commands/multi.spec.js
```diff
@@ -64,9 +64,9 @@ describe("The 'multi' method", function () {
                     // Provoke an error at queue time
                     multi1 = client.MULTI();
                     multi1.mset("multifoo", "10", "multibar", "20", helper.isString("OK"));
-                    multi1.set("foo2", helper.isError());
-                    multi1.incr("multifoo", helper.isNumber(11));
-                    multi1.incr("multibar", helper.isNumber(21));
+                    multi1.set("foo2");
+                    multi1.incr("multifoo");
+                    multi1.incr("multibar");
```
BridgeAR: Where there different return values depending on the redis version?

test/commands/expire.spec.js
bcoe: This is an issue on the Windows port of Redis.

test/commands/multi.spec.js
bcoe: yeah the `redis@2.4.5` build that has been ported to Windows has significantly different behavior for `multi/exec` -- I'm not too surprised, since there's a bunch of conditional logic surrounding this.

package.json
bcoe: newer versions of iojs and node.js have trouble with node-gyp + hiredis:

https://ci.appveyor.com/project/bcoe/node-redis/build/job/r1lcp1l8iyjfhw37
https://ci.appveyor.com/project/bcoe/node-redis/build/job/6774e4ym9ubmhi7u

bcoe:
@blainsmith @BridgeAR any objections to landing this, it gets us up and running with Windows CI.

test/commands/multi.spec.js
BridgeAR: Hm, if I look at the builds that are installed it says it's redis 2.8.21? ^^ If this is indeed an issue with windows, I'd like to wrap the helper.checks into another helper function to only include the check if it's not windows. That way we're still able to test everything on linux

package.json
BridgeAR: Using `npm install --msvs_version=2013` should do the trick :)

BridgeAR:
@bcoe do you have any idea why coveralls does not run?

test/commands/multi.spec.js
bcoe: I did some further testing, `multi1.incr("multibar")` doesn't ever invoke a callback -- this test was a noop.

`multi1.set("foo2")` does result in an error being returned on both old and new redis', so I'm adding this back.

bcoe:
I think this is ready to go out the door -- my only concern is the slight drop in test-coverage, but I think this is a rounding issue due to shuffling code around a bit, should come out in the wash when we land some of @BridgeAR's pulls that increase coverage.

blainsmith:
This looks good on my end. I think we are safe and let the coverage come back up with the pending PRs.

Issue #422, opened 2013-04-16
LPUSH with array


tp:
When supplying an array as second parameter to `lpush` node-redis creates a list but concatenates the single values by `, ` and saves them all as one element.

Is there any convenience method to push/append an array of strings to a list?

BridgeAR:
This is fixed since v.2.0

lucaswxp:
I'm in version 2.6.3 and this is still an issue: ERR wrong number of arguments for 'lpush' command

I tried

                redis.lpush('queue[reducao_tests]', args) // 1

                args.unshift('queue[reducao_tests]') // 2
                redis.lpush(args)

                redis.apply(redis, args) // 3

                args.unshift('queue[reducao_tests]') // 4
                redis.send_command('lpush', args)

Args is an array. All of them gives the same error.

Issue #390, opened 2013-03-03
Lists get condensed into a comma-separated string on retrieval


stickfigure:
Here's what I see in a node session:

> logger = function(err, results) { console.log(JSON.stringify({err: err, results:results})); }
[Function]
> r.rpush("abc", ["foo", "bar"], logger)
true
> {"err":null,"results":1}
> r.lrange("abc", 0, -1, logger)
true
> {"err":null,"results":["foo,bar"]}

Notice that results are ["foo, bar"] instead of ["foo", "bar"]. This seems like a bug of some sort - certainly not what I expected.

My package.json has:
    "redis": "0.8.x",
    "hiredis": "0.1.x",

vskarich:
Spent a few hours tonight scratching my head over this---almost it to production without realizing it what was happening. +1 to better documentation about this issue.

BridgeAR:
Not fixed => reopening

BridgeAR:
This got closed by #838

Issue #369, opened 2013-02-01
What is the proper way of using HDEL with an array of fields?


da99:
Is `HDEL` supposed to be used as the following?:

    client.hdel( my_key, [ field1, field2, ...] , callback );

or: 

    client.hdel( [ my_key, field1, field2, ...] , callback );

Right now, with `node-redis 0.8.2`, (no `hiredis`), the second approach works perfectly,  The first approach only works with an array of length 1: ` [ field1 ] `. Is this the correct behaviour? Or should it be the first approach? 

Here is a more complete code example: 

    var redis = require("redis"),
    client = redis.createClient();

    client.hmset("my_hash", {a: "a", b: "b"}, function (e, r) {
      redis.print(e,r);
      client.hmset("my_hash", {c: "c", d: "d"}, function (e, r) {
        redis.print(e,r);
        client.hdel('my_hash', ['a', 'c', 'd'], function (e, r) {

          redis.print(e,'This does not work: instead of 3 we get: ' + r);
          client.hdel("my_hash", 'a', 'c', 'd', function (e, r) {
            redis.print(e,'We get what we want using this approach: ' + r);
            client.quit();
          });
        });
      });
    });

BridgeAR:
This got closed by #838

bguiz:
For future reference, the following works now (tested on `redis@2.6.2`):

```
redisClient.hdel(key, [subkey1, subkey2], (err, result) => { /* ... */ });
```

PR #838, opened 2015-09-14
Consistent commands arguments

### LINKED ISSUE #686: arg2 of hmset in multi
Hello. If use argument 2 (arg2) as array in command HMSET without MULTI, all is well. But with MULTI if arg2 is array, query return null. To solve this problem it is necessary to add the code in function Multi.prototype.exec:
<pre><code>
if (command.toLowerCase() === 'hmget' && Array.isArray(args[1])) {
	var arr = args.pop();
	while (arr.length > 0) {
		args.push(arr.pop());
	}
}
</code></pre>
Sorry, i don't know how i can write a code that would look markup...

BridgeAR:
This fixes #686, #369, #422 and #390 and closes #634 and very likely more.

Earlier quite some commands were handled different. Now all commands take the same arguments, no matter if they are on multi or not.

### DIFF
--- a/index.js
+++ b/index.js
@@ -863,8 +863,16 @@ RedisClient.prototype.end = function () {
 function Multi(client, args) {
     this._client = client;
     this.queue = [["multi"]];
+    var command, tmp_args;
     if (Array.isArray(args)) {
-        this.queue = this.queue.concat(args);
+        while (tmp_args = args.shift()) {
+            command = tmp_args.shift();
+            if (Array.isArray(command)) {
+                this[command[0]].apply(this, command.slice(1).concat(tmp_args));
+            } else {
+                this[command].apply(this, tmp_args);
+            }
+        }
     }
 }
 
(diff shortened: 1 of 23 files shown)


index.js
```diff
@@ -948,75 +973,63 @@ RedisClient.prototype.auth = function () {
...
+        if (callback) {
+            args.push(callback);
+        }
+        tmp_args = ['hmset', key].concat(args);
+    } else if (typeof args === "object") {
+        tmp_args = ["hmset", key];
+        if (typeof key !== "string") {
+            key = key.toString();
+        }
+        for (field in args) {
```
BridgeAR: Using for in without checking for hasOwnProperty is faster than using Object.keys and looping over that. I can't think of a reason why someone would add something to the object prototype if it'll be passed to node_redis.

BridgeAR:
I'd like to get this into 2.0 as well. If we have enough tests for all possibilities how to use this I can't see a reason why not to include it into 2.0

BridgeAR:
This should be ready to merge

index.js
```diff
@@ -861,8 +861,16 @@ RedisClient.prototype.end = function () {
...
     this.queue = [["multi"]];
+    var command, tmp_args;
     if (Array.isArray(args)) {
-        this.queue = this.queue.concat(args);
+        while (tmp_args = args.shift()) {
+            command = tmp_args.shift();
+            if (Array.isArray(command)) {
+                this[command[0]].apply(this, command.slice(1).concat(tmp_args));
+            } else {
+                this[command].apply(this, tmp_args);
```
BridgeAR: To find a consistent way to handle the commands everywhere the same I had to use this slightly more complex logic. But since this is a very special case anyway this should not matter.

PR #839, opened 2015-09-14
Let v8 inline instead of doing in manually

BridgeAR:
v8 is inlining those variables since a long time.
This is cleaner to read and I also optimized .auth not to use .to_array.
And the redis version is always there when calling .info without arguments


BridgeAR:
Ready to merge

PR #843, opened 2015-09-14
Reduce the reauthenticate timeout to 333ms

BridgeAR:
It should not take all that long to load everything into redis

### DIFF
--- a/index.js
+++ b/index.js
@@ -193,7 +193,7 @@ RedisClient.prototype.on_error = function (msg) {
     this.ready = false;
 
     this.emit("error", new Error(message));
-    // "error" events get turned into exceptions if they aren't listened for.  If the user handled this error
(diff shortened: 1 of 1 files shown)


BridgeAR:
Ready to merge.

bcoe:
I've noticed you bringing the sleep times of tests down a few times, I just wanted to make sure that you know you can run tests in isolation -- I do this when I don't want to wait for the full suite:

`./node_modules/.bin/mocha ./test/commands/foo-command.js`

^ might be helpful .

BridgeAR:
@bcoe I use .only a lot. But I also run the whole suite quite often (each branch / pr * each rebase). That adds up quite a bit :)
I can't see the connection of lowering the reauthenticate timeout to the test timeouts though?

PR #845, opened 2015-09-15
Flush queue when .end() is called

crispy1989:
Flushes the queue when `.end()` is called to avoid pending callbacks not getting called, as mentioned in #572 .

### DIFF
--- a/index.js
+++ b/index.js
@@ -854,6 +854,9 @@ RedisClient.prototype.end = function () {
     }
     this.stream.on("error", function(){});
 
+    // Flush queue
+    this.flush_and_error("Redis connection ended.");
+
     this.connected = false;
     this.ready = false;
     this.closing = true;


BridgeAR:
I'm not certain if this is something that everyone would want how `.end` should behave. This is likely breaking some existing code. But if you'd just add a optional "flush" parameter and some tests, I'd say this is good to go. 

What do you think @NodeRedis/contributors?

bcoe:
Making it optional seems reasonable, I'd also love to see some tests.

BridgeAR:
Ping @crispy1989

BridgeAR:
Added in 4b100b8b64fd142d40db7c0c0662966980f9dbdb and I changed it to be optional. It'll be in the 2.1 release. Thx for your work @crispy1989

PR #846, opened 2015-09-15
Improve js parser

BridgeAR:


lib/parser/javascript.js
```diff
@@ -219,9 +209,6 @@ ReplyParser.prototype.execute = function (buffer) {
 };
 
 ReplyParser.prototype.append = function (newBuffer) {
-    if (!newBuffer) {
-        return;
```
BridgeAR: This is dead code

lib/parser/javascript.js
```diff
@@ -238,27 +225,13 @@ ReplyParser.prototype.append = function (newBuffer) {
         return;
     }
 
-    // very large packet
-    // check for concat, if we have it, use it
-    if (Buffer.concat !== undefined) {
```
BridgeAR: We do not support node < 0.10

lib/parser/javascript.js
```diff
@@ -270,10 +243,6 @@ ReplyParser.prototype._packetEndOffset = function () {
 
     while (this._buffer[offset] !== 0x0d && this._buffer[offset + 1] !== 0x0a) {
         offset++;
-
-        if (offset >= this._buffer.length) {
```
BridgeAR: This should be dead code. I checked a lot and could not trigger this.

lib/parser/javascript.js
```diff
@@ -152,36 +153,25 @@ ReplyParser.prototype.execute = function (buffer) {
...
+            break;
+        }
 
+        try {
             type = this._buffer[this._offset++];
 
             if (type === 43) { // +
                 ret = this._parseResult(type);
 
-                if (ret === null) {
```
BridgeAR: These four checks are all dead code.

lib/parser/javascript.js
```diff
@@ -76,12 +81,8 @@ ReplyParser.prototype._parseResult = function (type) {
             throw new IncompleteReadBuffer("Wait for more data.");
         }
 
-        if (this.options.return_buffers) {
-            return this._buffer.slice(start, end);
-        }
-
         // return the coerced numeric value
-        return +small_toString(this._buffer, start, end);
+        return +this._buffer.toString('ascii', start, end);
```
BridgeAR: Since we're looking for numbers we're safe to choose the fast ascii parser

lib/parser/javascript.js
```diff
@@ -152,36 +153,25 @@ ReplyParser.prototype.execute = function (buffer) {
 
     while (true) {
         offset = this._offset;
-        try {
-            // at least 4 bytes: :1\r\n
-            if (this._bytesRemaining() < 4) {
-                break;
-            }
+        // at least 4 bytes: :1\r\n
+        if (this._bytesRemaining() < 4) {
```
BridgeAR: This can't throw, so we are able to exclude it from the try block.

lib/parser/javascript.js
```diff
@@ -238,27 +225,13 @@ ReplyParser.prototype.append = function (newBuffer) {
...
-    }
-
+    this._buffer = Buffer.concat([this._buffer.slice(this._offset), newBuffer]);
     this._offset = 0;
 };
 
 ReplyParser.prototype.parseHeader = function () {
     var end   = this._packetEndOffset(),
-        value = small_toString(this._buffer, this._offset, end - 1);
+        value = this._buffer.toString('ascii', this._offset, end - 1);
```
BridgeAR: The headers should all be sent in ascii

lib/parser/javascript.js
```diff
@@ -54,6 +54,11 @@ ReplyParser.prototype._parseResult = function (type) {
             throw new IncompleteReadBuffer("Wait for more data.");
         }
 
+        if (type === 45) {
+            var result = this._buffer.toString(this._encoding, start, end);
+            return new Error(result);
```
BridgeAR: See #751

lib/parser/javascript.js
```diff
@@ -76,12 +81,8 @@ ReplyParser.prototype._parseResult = function (type) {
             throw new IncompleteReadBuffer("Wait for more data.");
         }
 
-        if (this.options.return_buffers) {
-            return this._buffer.slice(start, end);
```
BridgeAR: We want to check for numbers. So we should do that.

BridgeAR:
These changes should slightly improve the js parser speed. I analyzed the parser in more depths and as it currently works I can't see any way to improve it for big values any further :/. Most of the time is spend by concatenating the buffers for big values.

PR #847, opened 2015-09-15
Check that the password is from type string

BridgeAR:
Superseds #613 

@brycebaril I can't see anything bad about checking for the password validation. It won't slow anything down but improve the usability for at least a few people. 
@jcppman I did not ask in your pr earlier if you're still up for the pr, so if you want, I'll close this and you can implement these changes.

### DIFF
--- a/index.js
+++ b/index.js
@@ -252,7 +252,7 @@ RedisClient.prototype.on_connect = function () {
 
     this.init_parser();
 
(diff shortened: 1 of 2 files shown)


jcppman:
Thanks for asking! it's fine, this pr is surely more completed and covers more scenarios, I'd be happy to close mine and make my PR list clearer :smile:

PR #848, opened 2015-09-15
Fix auth callback being called more than once

BridgeAR:
@epoberezkin I guess this would solve #821 or did I miss something? I was not sure if I understood correct what you wanted to do, so I created these two tests.

### DIFF
--- a/index.js
+++ b/index.js
@@ -50,38 +50,30 @@ function RedisClient(stream, options) {
         this.options.socket_keepalive = true;
     }
     this.should_buffer = false;
-    this.command_queue_high_water = this.options.command_queue_high_water || 1000;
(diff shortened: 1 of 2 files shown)


epoberezkin:
It won't I am afraid, sorry. You have to synchronise the change in the application with the change in redis itself which is not really possible without failing some commands. The approach I had in mind is to try re-authenticating when some command fails. I've actually implemented both parts of #821 (enabling and changing the password in redis) in PR #831. I see you're overloaded with them, so easy to miss, I understand...

What the change does is completely transparent for the application, it doesn't have to call any methods to change password, and not a single command is lost. The driver re-authenticates using a supplied password in case some command fails because of authentication and then re-issues the same command to redis. There are tests as well.

PR #849, opened 2015-09-15
Add some convenience redis environments

BridgeAR:
I thought maybe it's nice to provide some default environment variables for the users?


BridgeAR:
I thought about it again and decided not to include further environment variables.

PR #851, opened 2015-09-17
trigger `connect`&`ready` event even the client is ready

isayme:
When we try listen `connect` or `ready` event on a ready client, our `listener` would not triggered.

````
// redisClient.js
module.exports = redis.createClient();
````
````
// clientA.js
var client = require('./redisClient');
client.on('ready', function() {
  // normally, we will go here once the client is ready
});
````
````
// clientB.js
var client = require('./redisClient');
client.on('ready', function() {
  // but there possibility that we cannot go here if the client.ready is `true` before we try to listen `ready`
});
````
I add a option called `always_trigger_ready` (the option name looks not good~~) for above situation.
By default, the option is `false` to be compatible.
If set the option to `true`, when we try to listen a `ready` or `connect` event on a ready client, our `listener` would triggered immediately.
If `no_ready_check` is `true`, only `connect` event will triggered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,7 @@ limits total amount of connection tries. Setting this to 1 will prevent any reco
 * `auth_pass` defaults to `null`. By default client will try connecting without auth. If set, client will run redis auth command on connect.
 * `family` defaults to `IPv4`. The client connects in IPv4 if not specified or if the DNS resolution returns an IPv4 address.
 You can force an IPv6 if you set the family to 'IPv6'. See nodejs net or dns modules how to use the family type.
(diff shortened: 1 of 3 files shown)


BridgeAR:
@isayme I would prefer not to add another option. But this is indeed a behavior that would have side effects otherwise.

But you're able to test if a client is already ready by checking `client.ready`. This is set to true on ready. After all, I suggest not to cehck the on ready in more than one place of your app. I suggest to handle it all in one place and the depending code should rely on that client or might use more than a single client.

PR #852, opened 2015-09-17
Parser fix

BridgeAR:
The returned value is now returned async. This is in important bug fix as otherwise errors thrown in a user callback would be catched by the parser.

This was the original intention behind wrapping the callback into a try catch block.

There is one tests that relies on our broken logic. I removed that one.

### DIFF
--- a/index.js
+++ b/index.js
@@ -86,7 +86,9 @@ RedisClient.prototype.install_stream_listeners = function() {
     });
 
     this.stream.on("data", function (buffer_from_socket) {
-        self.on_data(buffer_from_socket);
+        // The data.toString() has a significant impact on big chunks and therefor this should only be used if necessary
+        // debug("Net read " + this.address + " id " + this.connection_id + ": " + data.toString());
+        self.reply_parser.execute(buffer_from_socket);
     });
 
     this.stream.on("error", function (msg) {
@@ -273,8 +275,18 @@ RedisClient.prototype.init_parser = function () {
     this.reply_parser = new this.parser_module.Parser({
         return_buffers: self.options.return_buffers || self.options.detect_buffers || false
     });
-    this.reply_parser.send_error = this.return_error.bind(self);
-    this.reply_parser.send_reply = this.return_reply.bind(self);
+    // Important: Only send results / errors async.
+    // That way the result / error won't stay in a try catch block and catch user things
+    this.reply_parser.send_error = function (data) {
+        process.nextTick(function() {
+            this.return_error(data);
+        }.bind(this));
+    }.bind(this);
+    this.reply_parser.send_reply = function (data) {
+        process.nextTick(function() {
+            this.return_reply(data);
+        }.bind(this));
+    }.bind(this);
 };
 
 RedisClient.prototype.on_ready = function () {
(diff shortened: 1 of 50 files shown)


BridgeAR:
I also removed code from the js parser that is dead from now on, since we do not have to handle user errors in the parser logic anymore.

test/node_redis.spec.js
```diff
@@ -342,29 +342,6 @@ describe("The node_redis client", function () {
...
-                    var name = 'channel';
-
-                    client2.subscribe(name, function () {
-                        client.publish(name, "some message");
-                    });
-
-                    client2.on("message", function (channel, data) {
-                        if (channel === name) {
-                            assert.equal(data, "some message");
-                            throw Error('forced exception');
```
BridgeAR: This code is written by the user and he should handle an error occuring in his code by himself without node_redis doing anything with that error.

lib/parser/javascript.js
```diff
@@ -199,9 +179,6 @@ ReplyParser.prototype.execute = function (buffer) {
         } catch (err) {
             // catch the error (not enough data), rewind, and wait
             // for the next packet to appear
-            if (! (err instanceof IncompleteReadBuffer)) {
```
BridgeAR: This is dead code after fixing the issue that user errors got catched by the parser.

lib/parser/javascript.js
```diff
@@ -176,18 +166,8 @@ ReplyParser.prototype.execute = function (buffer) {
             } else if (type === 36) { // $
                 ret = this._parseResult(type);
 
-                if (ret === null) {
```
BridgeAR: This was already dead code before and I missed to remove it.

lib/parser/javascript.js
```diff
@@ -136,9 +129,6 @@ ReplyParser.prototype._parseResult = function (type) {
                 throw new IncompleteReadBuffer("Wait for more data.");
             }
             res = this._parseResult(ntype);
-            if (res === undefined) {
```
BridgeAR: This is dead code by changing the [`return undefined;`] (https://github.com/NodeRedis/node_redis/pull/852/files#diff-778060651e24292442eb47a196897e2eL95) to `return null;`

lib/parser/javascript.js
```diff
@@ -176,18 +166,8 @@ ReplyParser.prototype.execute = function (buffer) {
             } else if (type === 36) { // $
                 ret = this._parseResult(type);
 
-                if (ret === null) {
-                    break;
-                }
-
-                // check the state for what is the result of
-                // a -1, set it back up for a null reply
-                if (ret === undefined) {
```
BridgeAR: This is dead code by changing the [`return undefined;`] (https://github.com/NodeRedis/node_redis/pull/852/files#diff-778060651e24292442eb47a196897e2eL95) to `return null;`

index.js
```diff
@@ -480,33 +494,22 @@ RedisClient.prototype.connection_gone = function (why) {
     }, this.retry_delay);
 };
 
-RedisClient.prototype.on_data = function (data) {
-    // The data.toString() has a significant impact on big chunks and therefor this should only be used if necessary
-    // debug("Net read " + this.address + " id " + this.connection_id + ": " + data.toString());
-
-    try {
-        this.reply_parser.execute(data);
```
BridgeAR: We do not need this code anymore as the parser won't throw any non parser errors anymore.

PR #685, opened 2014-11-14
Fix Command Queue Stack Error #577, #603

noansknv:
@DTrejo @mranney @nateps Solving https://github.com/mranney/node_redis/issues/577 and https://github.com/mranney/node_redis/issues/603. 

Probably better solution would be what @rootslab proposes, however rewriting message pipeline is beyond the scope of this fix ;-)

Issue #577, opened 2014-04-23
Subscribing & unsubscribing to multiple channels crashes client 


josephg:
Using `redis@0.10.1`, this code:

```javascript
var redis = require('redis');
var client = redis.createClient();

client.subscribe('one', 'two', 'three', function(err) {});
client.unsubscribe('one', 'two', 'three', function(err) {});
client.subscribe('fsad', 'fdssfd', 'fdssdf', function(err) {});
```

... Causes this crash:

```
$ node test.js

events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: node_redis command queue state error. If you can reproduce this, please report it.
  at RedisClient.return_reply (/Users/josephg/src/rediscrash/node_modules/redis/index.js:705:15)
  at ReplyParser.<anonymous> (/Users/josephg/src/rediscrash/node_modules/redis/index.js:309:14)
  at ReplyParser.EventEmitter.emit (events.js:95:17)
  at ReplyParser.send_reply (/Users/josephg/src/rediscrash/node_modules/redis/lib/parser/javascript.js:300:10)
  at ReplyParser.execute (/Users/josephg/src/rediscrash/node_modules/redis/lib/parser/javascript.js:211:22)
  at RedisClient.on_data (/Users/josephg/src/rediscrash/node_modules/redis/index.js:534:27)
```

It has something to do with redis replying with multiple responses, which confuses the command queue. I can work around it by only subscribing & unsubscribing to one item at a time.

blainsmith:
In order to better support this project and its new group of collaborators under a new org we are trying to clean up and close issues older than 6 months. Your issue may have been fixed since the issue was open, however, if you are still experiencing a problem **please** re-open this issue and we will label it accordingly.

BridgeAR:
@josephg @wbotelhos could you have a look at #857 ? :)

wbotelhos:
I'm sorry @BridgeAR , I can't.
I'm currently in another company. :disappointed:

BridgeAR:
This is finally fixed on master

PR #857, opened 2015-09-18
[WIP] Fix Command Queue Stack Error in PUBSUB mode

### LINKED ISSUE #603: Command Queue Stack Error
I'm trying to `SUBSCRIBE`/`UNSUBSCRIBE` from channels in the following way (minimal working example):

```
var client = require('redis').createClient();

function subscribe(channels) {
    client.unsubscribe(function (error) {
        console.log(error);
    });
    client.subscribe(channels, function (error) {
        console.log(error);
    });
}

subscribe(['1', '3']);
subscribe(['2']);
```

The idea is, at any moment, completely replace this list of subscribed channels with a different one. This immediately causes the following error:

```
events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: node_redis command queue state error. If you can reproduce this, please report it.
    at RedisClient.return_reply (/myproject/node_modules/redis/index.js:705:15)
    at HiredisReplyParser.RedisClient.init_parser (/myproject/node_modules/redis/index.js:309:14)
    at HiredisReplyParser.emit (events.js:95:17)
    at HiredisReplyParser.execute (/myproject/node_modules/redis/lib/parser/hiredis.js:43:18)
    at RedisClient.on_data (/myproject/node_modules/redis/index.js:534:27)
    at Socket.stream.on.self.should_buffer (/myproject/node_modules/redis/index.js:91:14)
    at Socket.emit (events.js:95:17)
    at Socket.stream.pause.paused (_stream_readable.js:748:14)
    at Socket.emit (events.js:92:17)
    at emitReadable_ (_stream_readable.js:410:10)
```

Making the call to `client.subscribe` inside the callback passed to `client.unsubscribe` instead doesn't help. Wrapping the call to `client.subscribe` with `process.nextTick` seems to provide a workaround the problem. Any clues on how to solve this?

BridgeAR:
This is going to fix a couple of open tickets. I'm not really happy with the solution as is but at least the app does not crash anymore and it seems to work as it should. But there are likely still errors with the pubsub mode and this is probably not a good fix. It would be good if we'd find a better solution at some later point.

This superseds #685 (it somehow did not fix the issue in my tests?) and fixes #603 and #577

### DIFF
--- a/index.js
+++ b/index.js
@@ -571,12 +571,30 @@ RedisClient.prototype.return_reply = function (reply) {
     // If the "reply" here is actually a message received asynchronously due to a
     // pubsub subscription, don't pop the command queue as we'll only be consuming
     // the head command prematurely.
-    if (this.pub_sub_mode && Array.isArray(reply) && reply.length > 0 && reply[0]) {
+    if (Array.isArray(reply) && reply[0]) {
         type = reply[0].toString();
     }
 
     if (this.pub_sub_mode && (type === 'message' || type === 'pmessage')) {
         debug("Received pubsub message");
+    } else if (type === "subscribe" || type === "unsubscribe" || type === "psubscribe" || type === "punsubscribe") {
+        command_obj = this.command_queue.peek();
+        // If in pubsub mode and subscribing / unsubscribing to multiple channels at once, we have to return the multiple results somehow
+        //
+        // e.g. subscribe '5', 'test' would result in two replies instead of one.
+        //
+        // So we just keep the command until all replies came through
+        if (command_obj) {
+            command_obj.command = type; // URGS :/
+        }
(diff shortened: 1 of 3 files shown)


BridgeAR:
@NodeRedis/contributors I'd like to have someone look at this, since I'm not so certain if this is a good way to handle it.

blainsmith:
I would say for the sake of 2.0.0 we should get this in, but make an issue for further investigation. I would even throw a "help wanted" tag on the issue to get some community feedback on a possible refactor.

bcoe:
@BridgeAR when you have a moment, mind rebasing this with master; I want to see if the Windows failures stop cropping up.

BridgeAR:
@bcoe I have to rework this again. I think I got a way to fix this more approriate, but I have to find the time to do so.

PR #861, opened 2015-09-18
Remove unnecessary line from example

frewsxcv:
This line doesn't seem necessary nor relevant to this example


BridgeAR:
Thx @frewsxcv

Issue #806, opened 2015-09-01
Batch commands functionality


simontabor:
Related to discussions in https://github.com/NodeRedis/node_redis/pull/799

It's very common to want to run a batch of commands at the same time and have a single callback for when they are completed. A lot of the time, people use `MULTI` for this, which is okay but adds noticeable overhead when pushing Redis hard.

An example of this is at GoSquared, where we add a new visitor to our tracking. This involves many redis commands - setting a hash of the visitor details and adding them to relevant lists what we want to run at the same time, but we don't need a multi.

As mentioned in the previous issue, I've got the functionality in redis clustr (https://github.com/gosquared/redis-clustr/blob/master/src/RedisBatch.js) and it's been incredibly useful. Usage is as follows:

```js
var batch = redis.batch();
batch.set('key', 'value');
batch.get('key', function(err, resp) {
  // "value"
});;
batch.exec(function(err, resp) {
  // [ "OK", "value" ]
});
```

Trying to use something like `async` to group the commands together results in awful code, this is familiar and clear (exactly the same usage as multi).

Thoughts?

stockholmux:
So, the above isn't atomic like running `multi`? 

I'm interested in an optimized `multi`-like block that will efficiently run a single command if that is all that is needed. Here is my use case:

```
var myAtomicBatch = client.atomicBatch();

myAtomicBatch.set('key','value');
if (isSomething === true) {
  myAtomicBatch.set('key-for-is-something','value);
}
myAtomicBatch.exec(function(err, values) {
  //...
});
```

So the second `set` command is only called under certain condition(s) but I need both the first and second commands to run together if it is called. 

We do something similar to this now but just take the performance hit if the conditional is not `true`. If I wanted to refactor for to avoid the multi / single command issue, the code would come out fairly messy.

simontabor:
Exactly - this will handle that for you

stockholmux:
@simontabor Gotcha - the first callback and the part about using `async` made me think you were going for a non-atomic series of commands. Also, the link to RedisBatch.js is 404'ing for me.

simontabor:
@stockholmux edited the link, for some reason it had a full stop at the end.

Think I've misunderstood slightly - you're looking for it to run a multi/exec block if it's more than one command but just run the single command if there's only one?

[...]

Issue #659, opened 2014-09-15
Could node-redis support 'redis pipeline'?


mrFranklin:
Hi:
   Could node-reids support 'redis pipeline' at soon?

   my question is simalr with this one: https://github.com/mranney/node_redis/issues/492 

   for example: when the client request 'ping' 10 times. it will send TEN tcp packages to the redis-server, and receive ONE tcp package form the server. (using wireshark to catpure the network data)
   And I'v tested the 'redis-python' client. it just send ONE tcp package(10 'ping' in the package) and receive ONE package from the server.
  I think it's necessary to support the feature. it can reduce the newwork IO, and improve the performance.
  Thank you.

PR #868, opened 2015-09-24
Batch and atomicBatch functionality

simontabor:
Implements #806 

Needs tests (ahh!) but it functionally works.

### All cases
```js
var b = redis.batch();
b.set('test', '1');
b.get('test', function(err, res) {
  //= null, '1'
});
b.exec(function(err, res) {
  //= null, [ 'OK', '1' ]
});
```
```
"set" "test" "1"
"get" "test"
```

### Multiple commands

```js
var b = redis.atomicBatch();
b.set('test', '1');
b.get('test', function(err, res) {
  //= null, '1'
});
b.exec(function(err, res) {
  //= null, [ 'OK', '1' ]
});
```
```
"multi"
"set" "test" "1"
"get" "test"
"exec"
```

### Single command

```js
var b = redis.atomicBatch();
b.set('test', '1');
b.exec(function(err, res) {
  //= null, [ 'OK' ]
});
```
```
"set" "test" "1"
```
Note: no multi/exec wrapper as it's unnecessary for a single command

### DIFF
--- a/index.js
+++ b/index.js
@@ -870,7 +870,27 @@ function Multi(client, args) {
     }
 }
 
+function Batch(client, atomic, args) {
+    this._client = client;
+    this.queue = [];
+    this.atomic = atomic;
+    var command, tmp_args;
+    if (Array.isArray(args)) {
+        while (tmp_args = args.shift()) {
+            command = tmp_args.shift();
+            if (Array.isArray(command)) {
+                this[command[0]].apply(this, command.slice(1).concat(tmp_args));
+            } else {
+                this[command].apply(this, tmp_args);
+            }
+        }
+    }
+}
+
 exports.Multi = Multi;
+exports.Batch = Batch;
+
+
 
 commands.forEach(function (fullCommand) {
     var command = fullCommand.split(' ')[0];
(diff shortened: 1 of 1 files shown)


BridgeAR:
@simontabor thx for your work! I'd like to see such a feature in combination with [pipelining](http://redis.io/topics/pipelining) commands together to reduce the RTT.
Could you try and find a way to implement pipelining? :) That would be just the same as your current batch functionality but more powerful. You can very likely just overload Multi and include some special conditions there to use pipelining. That way both, multi with transactions and "batch" without transactions could benefit from pipelining and you reduce code duplication and prevent divergent behavior. :)

simontabor:
yeah i'd need to get my head around how pipelining is implemented... 

isn't it just automatically handled in `RedisClient.prototype.send_command`, so sending all of the commands through at once will have them all in a single pipeline?

BridgeAR:
Related: #659

send_command does not handle it automatically. It sends each command individually. So we'd have to collect all querys first.

BridgeAR:
Ping @simontabor

BridgeAR:
Closed in favor of #878

PR #877, opened 2015-10-04
Small fix to allow bundling.

subjectix:
This prevents bundlers from trying to include hiredis since it's a native module.

P.S. I'm bundling for node not browser :grin:

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,7 +24,10 @@ exports.debug_mode = /\bredis\b/i.test(process.env.NODE_DEBUG);
 
 // hiredis might not be installed
 try {
-    parsers.push(require("./lib/parsers/hiredis"));
+    // Confuse bundlers so that they won't try
+    // to include this native module.
+    var hiredisParserPath = "./lib/parsers/hiredis";
+    parsers.push(require(hiredisParserPath));
 } catch (err) {
     /* istanbul ignore next: won't be reached with tests */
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -24,7 +24,10 @@ exports.debug_mode = /\bredis\b/i.test(process.env.NODE_DEBUG);
 
 // hiredis might not be installed
 try {
-    parsers.push(require("./lib/parsers/hiredis"));
+    // Confuse bundlers so that they won't try
+    // to include this native module.
+    var hiredisParserPath = "./lib/parsers/hiredis";
+    parsers.push(require(hiredisParserPath));
```
BridgeAR: I suggest adding a comment like `// Native module` instead. That way you could parse the comments while bundling and do not rely on anything else. 
Is your bundling logic public?

subjectix: It's just browserify :sweat_smile: :sweat_smile: :sweat_smile: 
Adding that code does fix it. but if that somehow hinders readability and you prefer not to add that, I'll work around this issue using browserify's own api.

BridgeAR: It does not really matter to me, but I'd guess that this would change at some point and that browserify will detect that in near future too.

subjectix: You're right. That's likely. I guess the best way would be what I'm doing right now, exclude hiredis from require resolution. Thanks for your great code anyway ^^

Issue #875, opened 2015-10-01
usage with node cluster causes intermittent latency issues


torra:
We run ec2 instances using node cluster, with ~5 workers per instance. We were using an old version (0.10.3), and when we tried to upgrade we started seeing timeouts on the load balancer. Here is what I noticed:
- disabling cluster fixes the issue
- leaving cluster enabled, but only using one worker fixes the issue
- reverting back to 0.10.3 fixes the issue (update: 0.12.1 and 1.0.0 also seem to work fine)

Unfortunately I don't have time to dig much deeper, but something since 0.10 seems to not play nice with node cluster.

Any thoughts on what would cause such behavior?

Update with some more info:
- We talk to several different redis servers, both locally and remote. Locally we talk over a unix socket
- We are running redis 3.0.3
- On our local redis server, we do `get`, `hmget`, `hgetall`, `smembers`, `sunion`, and `sinter` operations
- On our remote servers, we do `zrange` and `zrevrange` operations

- I have been testing a bit with both node 0.10.40, but mostly (and currently only) 4.1.1

BridgeAR:
@torra would you be so kind and also test 0.12.1? And could you give me some insights into your code?

torra:
0.12.1 and 1.0.0 both work fine

Our code does mostly `get`, `hmget`, and `hgetall` operations, along with some `zrange` and `zrevrange`. See update above for more details

BridgeAR:
@torra did you check the breaking changes and the difference how reconnecting is handled now? And could you test v.2.1 too? I just released it about an hour ago.

torra:
@BridgeAR which difference in reconnecting specifically are you referring to? I did review the breaking changes, and while testing everything seems(ed) fine. We log all errors, disconnects, etc. anyways, and I haven't seen a sign of anything like that. 

This only occurs when using node cluster, so it seems like something specific to that. I have also not been able to even reproduce the problem outside of production. I can only see that some requests end up taking abnormally long, hitting our load balancer timeout (60s) - I haven't been able to actually see a request trigger the issue.

I can't test with 2.1 right now, but early next week I can give it a shot (probably Tuesday).

I would love to try and come up with a simple way to reproduce the problem, but unfortunately I haven't had the time. It would be great to hear from others using nodes cluster module in production.

zaggino:
[...]

PR #878, opened 2015-10-07
node_redis on speed

BridgeAR:
This is a bigger pull request that includes a couple of changes that someone might want to look at before merging.

1. I exchanged our built in queue with the one from Petka Antonov after benchmarking them. The later is especially faster if the queue gets bigger.
2. I added a .batch command that executes everything just the same as .multi but without transaction.
3. Both .multi and .batch got a improved pipelining that reduces RTT.
4. General performance improvements.
5. Fixed should_buffer boolean for .exec, .select and .auth commands not being returned.
6. .multi got a .exec_atomic function that won't fire the multi / exec to the db if no command was supplied and that will only use multi if more than one command has to be executed.

Overall the performance change is obvious and using .batch for multiple commands improves the performance by up to 400% for node > 0.10 and up to 100% for node 0.10 (it will lower the performance minimal opposed to our normal pipelining if the memory usage is to high in node 0.10).


BridgeAR:
Ping @simontabor @bcoe @brycebaril 

Reference #806 and #868

index.js
```diff
@@ -1090,6 +1146,59 @@ Multi.prototype.execute_callback = function (err, replies) {
...
+};
+
+Multi.prototype.exec = Multi.prototype.EXEC = function (callback) {
+    var len = this.queue.length;
+    var self = this;
+    var index = 0;
+    var args;
+    if (len === 0) {
+        if (callback) {
+            callback(null, []);
```
simontabor: should we setImmediate here to keep it async?

index.js
```diff
@@ -1008,31 +1053,42 @@ Multi.prototype.send_command = function (command, args, index, cb) {
     });
 };
 
-Multi.prototype.exec = Multi.prototype.EXEC = function (callback) {
+Multi.prototype.exec_transaction = function (callback) {
     var self = this;
+    var len = this.queue.length;
+    var cb;
```
simontabor: could we add a length check like https://github.com/NodeRedis/node_redis/pull/878/files#diff-168726dbe96b3ce427e7fedce31bb0bcR1170

```js
if (len === 1) {
   // no commands, only [ [ 'multi' ] ] 
  if (callback) {
    callback(null, []);
    // setImmediate(function() {
    //   callback(null, []);
    // });
  }
  return false;
}
```

there's never a reason to send `MULTI` `EXEC` with no commands!

bcoe:
This is looking great!

I'm a little concerned that there's a variety of different tests that are now failing on our Windows CI. This combined with https://github.com/NodeRedis/node_redis/issues/875, which appears to be a regression, makes me think that we should step back and make some stability fixes before we land any more performance fixes.

Here's what I suggest:

1. we feature freeze for the time being, while we dig into #875 in particular:
  * this sort of intermittent issue drives a developer away from using a library. I was personally tempted to pin npm at `node_redis@1.0.0` at the cost of some of the awesome new work that's being done on this library.
2. we create a `p0` ticket for digging into some of our Windows CI issues, I believe these to represent a class of real bugs that just happen to occur more on a Windows platform:
  * https://github.com/NodeRedis/node_redis/issues/880
3. we treat #875 as a `p0` ticket:
  * find a way to reproduce this regression (I'm almost 100% sure this regression is our fault, given that the reporters can't create the issue on old versions of node_redis).
  * fix it.
4. Once we've done all this (really it's just a bit of spring cleaning, and the sort of work that's important after any major library update)., we start landing new work again including these amazing performance tweaks.

CC: @brycebaril @BridgeAR @mranney @erinspice @blainsmith

BridgeAR:
@bcoe thx. As I already pointed out in #881 and #880 I do not really agree with waiting for a fix for #875. The bug will stay until it is fixed no matter if this is added or not. This is actually increasing the coverage and adding a couple of new test cases and I'm quite certain that I know where I introduced that regression.

Pinning the version to 1.0 is likely not a good idea, as 2.0 fixed a couple major bugs. I actually went though ~ 300 closed tickets tagged them and checked if they are still relevant and reopened them if the ticket reported a not yet fixed bug. Most of those are already fixed from 2.0 on.

index.js
BridgeAR: That is indeed true. The question here is though: do you want to add another check for a case that will be handled properly anway? The .batch function works differently, as it would otherwise not behave the same as with commands. So I'd say it's fine to not handle that case any special.

index.js
BridgeAR: Good catch

index.js
simontabor: I'm not sure I follow - this case doesn't get handled at the moment and the redis server gets an empty `MULTI` `EXEC`, when we could immediately callback.

I've had to deal with multiple issues where there are empty multi blocks, and being able to call `.exec` without checking that `m.queue.length > 1` would be very useful (ideally, users would never want/need to play around with the internal queue).

BridgeAR: I was not aware that you'd never get a reply in this case. I agree with you that it would be nice not to check for that, so I'll add that.

BridgeAR: I checked and it is already handled ;) Yes, an empty command is fired, but that does not hurt as it'll still return `null, []`

simontabor: Sorry I think we've got wires crossed here! 

Yes, running the empty multi/exec works without any problems and responds with `null, []` (correctly), but if we're talking about speed improvements then preventing the unnecessary round-trip to redis will make things faster! 

If this isn't in the node_redis code, then users (like me) will have to continue checking that `multi.queue.length > 1` all over their application code. It's a simple and easy optimisation that can be put into node_redis, so hopefully it can be added.

Slightly related to this (multi queue lengths!), did you have any thoughts about adding `atomicBatch`? So that'd be extracting the single command from a multi if there's just one in the queue (if `multi.queue.length === 2`)

BridgeAR: I added Multi.exec_atomic (I can't get used to snack case^^). It'll use the batch.exec function if none or one command is used and therefor return without fireing any command to redis.

PR #894, opened 2015-10-19
Add support for TLS connections

paddybyers:
This PR supersedes https://github.com/NodeRedis/node_redis/pull/527, with those changes being rebased to the current `master`.

I have add tests in `tls.spec.js` which runs with the variants of `options`, `ip` of `allTests`. This seemed the most direct way to add tests, although you will see that there is a lot of overlap with the `connection.spec.js` tests. I wondered if instead TLS should be a different set of `options` or a different `protocol` enumerated by `allTests` but that looked like it would generate much greater redundancy.

Currently the TLS tests run unconditionally but we might want a way to skip them if `stunnel` isn't available on the platform running the tests.

I have run the tests locally only, on a Mac, so YMMV on other platforms. Feedback appreciated.


BridgeAR:
Thx a lot! I'll have a closer look at it later on

BridgeAR:
I'm sorry that I was not able to do a proper review yet. I only had a short glimpse yet. I'm really busy at the moment.

Please include stunnel for travis ([example](https://github.com/isotoma/isotoma.recipe.stunnel/blob/master/.travis.yml)). We'll probably have to deactivate the the tests for windows though. You can do this just the same we did with the redis process (`helper.redisProcess().spwanFailed()`).

One style nit: please always use a space between an if and a bracket.

I think it's already looking very promising and I'm excited that after over three years[sic!] this is going to be included in node_redis :) And I'm also very happy that you were still willing to code this.

After looking in the code it might be worth to deprecate .createClient() and just use new Redis(options) instead. What do you think? That way it might become even cleaner, but this should be another pull request.

stockholmux:
@BridgeAR What's the advantage of ditching `.createClient()` and moving to `new Redis()`?

BridgeAR:
@stockholmux It's a two step initiation and I prefer a single init function. Right now we have two places where we have to copy the options to make sure we do not mutate those and instead we could just do that once while creating a new instance.

paddybyers:
> One style nit: please always use a space between an if and a bracket.

I think I've addressed all of these (https://github.com/NodeRedis/node_redis/commit/2d22e455d1d66f350d3070d8e2fac43c1e447575).

> After looking in the code it might be worth to deprecate .createClient() and just use new Redis(options) instead. What do you think? That way it might become even cleaner, but this should be another pull request.

As this would be an incompatible change it definitely needs to be a separate PR IMO. We don't need to create new reasons to delay landing these changes :)

I'd always understood that `createClient()` is the public API and `new Redis()` is an internal API, but I don't know how widely used it is externally.  But there is some value in keeping `new Redis()` as a relatively clean API, with all of the optionality, overloading and backwards compatibility in `createClient()`.

paddybyers:
> Please include stunnel for travis (example).

I'll have a go at this, but is anyone volunteering to try the stunnel changes out on Linux first?

BridgeAR:
@paddybyers I tried the original pr and stunnel worked properly on linux there :)

I have a look at it again tomorrow.

test/lib/stunnel-process.js
```diff
@@ -0,0 +1,83 @@
...
+
+StunnelProcess.prototype.clear = function() {
+    this.stunnel = null;
+    clearTimeout(this.timer);
+};
+
+StunnelProcess.prototype.stop = function(done) {
+	if (this.stunnel) {
+	    this.stunnel.kill();
+	}
```
BridgeAR: Indentation

index.js
```diff
@@ -1214,54 +1267,32 @@ Multi.prototype.exec = Multi.prototype.EXEC = Multi.prototype.exec_batch = funct
...
-    redis_client.address = cnxOptions.host + ':' + cnxOptions.port;
-
-    return redis_client;
-};
-
 var createClient = function (port_arg, host_arg, options) {
     if (typeof port_arg === 'object' || port_arg === undefined) {
         options = port_arg || options || {};
-        return createClient_tcp(+options.port, options.host, options);
+        return new RedisClient(null, options);
```
BridgeAR: I think it would be nicer to only pass options to the constructor and check in there if it's an object / stream / undefined. That way we keep the backwards compatibility but the Redis call looks nicer :)

index.js
```diff
@@ -94,17 +84,65 @@ function RedisClient(stream, options) {
...
+        this.cork = noop;
+        this.once('ready', function () {
+            self.cork = function (len) {
+                self.pipeline = len;
+                self.pipeline_queue = new Queue(len);
+            };
+        });
+        stream.uncork = noop;
+        this.write = this.writeStream;
+    }
```
BridgeAR: Good catch. This is not necessary after #901 has landed though. I fixed the logic so it's independent from the used stream.

index.js
```diff
@@ -94,17 +84,65 @@ function RedisClient(stream, options) {
...
+        if (options.path) {
+            //unix
+            this.address = connection_options.path = options.path;
+        } else {
+            //tcp or tls
+            connection_options.port = options.port || default_port;
+            connection_options.host = options.host || default_host;
+            connection_options.family = (options.family === 'IPv6') ? 6 : 4;
+            if (tls_options) {
+                Object.keys(tls_options).forEach(function(opt) {
```
BridgeAR: You can use a `for ... in` loop here without checking hasOwnProperty, since the options got copied earlier and therefor won't contain anything that does not belong here.

BridgeAR:
Just some minor comments. Otherwise this is already looking really good @paddybyers :+1:

xirzec:
I'm currently having to depend on @paddybyers branch in order to use this package with Redis on Azure. Would love to get this merged so I can depend on a real release.

BridgeAR:
@xirzec I want to add this as soon as travis runs properly.

@paddybyers I rebased your commits in a separate [branch](https://github.com/NodeRedis/node_redis/tree/tls) but I changed your code that way and it would be nice if you could check if you're happy with the way I rebased the commits. I did not change anything in the tests. 

I added stunnel to travis but stunnel can't be started properly and I'm uncertain why.

paddybyers:
@BridgeAR: I've had a look and it all looks ok.

This commit https://github.com/NodeRedis/node_redis/commit/4827ead3cc8de96caf95a70358a46b542a28022f crept in - is that a depedency?

BridgeAR:
Merged. I skipped the tls tests on windows and travis for now. As soon as https://github.com/travis-ci/apt-package-whitelist/issues/403 is resolved, we could try to activate the tests on travis again.

Issue #678, opened 2014-10-30
Very slow gets of big values


qzb:
I noticed that getting big values take more time than it should. Worse, it looks like time of get grows expotentially with size of value.

I wrote following script to test it:
```js
var redis = require('redis');
var fs = require('fs');

var key = Date.now().toString();
var data = fs.readFileSync('./data.bin'); // big file (25-100MB)

var client = redis.createClient(6379, 'localhost');

client.on('connect', function () {
    var time = Date.now();
    client.set(key, data, function (err) {
        console.log('set time: %s', Date.now() - time);
        time = Date.now();
        client.get(key, function (err, data) {
            console.log('get time: %s', Date.now() - time);
            process.exit();
        });
    });
});
```
When I pass to it 25MB of data, it takes about 6 seconds to get value from local redis, for 50MB it takes about 25 seconds and for 100MB it takes near 100 seconds.

ajones513:
My results...

Using Hiredis | Size | Write (ms) | Read (ms)
---|---|---|---
No | 25 MB | 22 | 4114
No | 50 MB | 39 | 15128
Yes | 25 MB | 24 | 1026
Yes | 50 MB | 38 | 2007

Looks pretty linear with Hiredis.

mranney:
Thanks for digging into this. I have confirmed these results, and there is not an easy way to fix this with the JS parser. It's certainly not impossible to fix in the JS parser, but it'll be some delicate surgery to do without slowing down or breaking the common case. I guess this is a good reason to keep hiredis working, at least until we can fix it.

For the curious, here's a flame graph of the test program above:

http://ranney.com/redis.svg

This program spends all of its time copying buffers around. hiredis wins because it calls JS back with a single 50MB result.

brycebaril:
I've done a lot of benchmarking as well and have seen this is an issue with the parser. I've played with some things to try and fix it but it would be an api change for sure.

b-long:
:+1:

blainsmith:
In order to better support this project and its new group of collaborators under a new org we are trying to clean up and close issues older than 6 months. Your issue may have been fixed since the issue was open, however, if you are still experiencing a problem **please** re-open this issue and we will label it accordingly.

BridgeAR:
Reopening as the js parser is still not handling big values well

BridgeAR:
I analyzed the parser in depth and the cause of the problems seems to be the following:

[...]

PR #900, opened 2015-10-26
Fix parsing huge replies by JavascriptReplyParser

qzb:
This pull request fixes issue #678. Instead of parsing whole buffer when new chunk arrives I'm adding it to BufferList and converting it to regular Buffer only when it's necessary.

### DIFF
--- a/lib/parsers/javascript.js
+++ b/lib/parsers/javascript.js
@@ -1,11 +1,12 @@
 'use strict';
 
-var util   = require('util');
+var util = require('util');
+var BufferList = require('bl');
 
 function JavascriptReplyParser() {
     this.name = exports.name;
-    this._buffer            = new Buffer(0);
-    this._offset            = 0;
+    this._bufferList = new BufferList();
+    this._offset = 0;
 }
 
 function IncompleteReadBuffer(message) {
@@ -25,7 +26,7 @@ JavascriptReplyParser.prototype._parseResult = function (type) {
         end = this._packetEndOffset() - 1;
         start = this._offset;
 
-        if (end > this._buffer.length) {
+        if (end > this._bufferList.length) {
             throw new IncompleteReadBuffer('Wait for more data.');
         }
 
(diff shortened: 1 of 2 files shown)


BridgeAR:
@qzb you did indeed fix big chunks (they are not as fast as they should but sufficient). But the downside is that the average use case is way slower now (up to five times slower).

So we have to implement something similar on our own that exactly fits to our needs. How it could be done is to switch the concat and the parsing logic. First the buffer should be checked and if we do not find the end of the file / data, just cache that chunk and wait for the next one until you find the end of the data. Than concat everything and use the already collected meta info to return the appropriate result.

b-long:
@BridgeAR Maybe keep moving forward on @qzb 's branch and implement a flag so that clients can choose to use one implementation or the other?

BridgeAR:
@b-long we could implement this as separate parser. That way everyone is free to choose.

@qzb would you be up to change this pr accordingly? :)

qzb:
@BridgeAR I think separate parser will be overkill, I'll try to get rid of bl, and implement buffers caching as you described.

BridgeAR:
@qzb that would be even better :+1:

BridgeAR:
@qzb btw @luin and me want to split the parser into a separate module (https://github.com/NodeRedis/node-redis-parser). Would you be interested in maintaining the parser after fixing it?

qzb:
Yes, I am :)

basvanmeurs:
I had a big mget with 200+ entries, resulting in a 12MB response. The javascript parser of 2.6.3 took about 6s, while the hiredis one did it in 0.5s. So we used hiredis instead of javascript. What is very annoying is the console.error 'Using hiredis is discouraged. Please use the faster JS parser by removing the name option.'. It triggers our error email system. I think that this should not be an error but a warning (console.warn).

BridgeAR:
@basvanmeurs the js parser should pretty much always be better than the hiredis one. But parsing big chunked arrays is not optimized well yet. The reason is that depending on the data in the array the parser has to parse the same data again over and over. This will be addressed in a future version and overall I still recommend to use the js parser to you even though your specific example is currently slower.

basvanmeurs:
Thanks BridgeAR, I am looking forward to that improvement!

basvanmeurs:
BridgeAR, you recommend to use the js parser. I'm afraid that it is not possible. The JS parser not only takes longer, it also takes 100% CPU. Because we use a lot of mgets, just switching to the JS parser brought down our complete website!

BridgeAR:
@basvanmeurs thanks for bringing this again to my attention. I have a fix for it ready https://github.com/NodeRedis/node-redis-parser/pull/26 and it would be nice if you could try that. In most cases the parsing should now be faster than the hiredis parser.

basvanmeurs:
Hi BridgeAR, I finally had the time to try it out. It looks like the issue has been fixed, thanks! We'll start replacing hiredis with the new parser.

PR #903, opened 2015-10-26
Fix and simplify parser

### LINKED ISSUE #902: client.get hangs and never calls callback
I'm currently experiencing some odd behavior where every so often, some of the calls to `client.get` hang and the callback function is never called. I've turned on `NODE_DEBUG=redis`, but the logs aren't showing me anything more helpful, just that a call to `client.get` was made

```
Send barreleye.redistogo.com:[port] id 1: *2
$3
get
$70
[key]
```

I have a ton of debug logs that would fire if the callback comes back, but instead nothing happens and Heroku logs an H12 for a 30 second timeout.

```
heroku[router] at=error code=H12 desc="Request timeout" method=GET path="/all" host=[app] request_id=[id] dyno=web.2 connect=0ms service=30001ms status=503 bytes=0
```

Here is what my code looks like

```js
debug(key + ": Checking session in redis");
client.get(key, function(err, response) {
  if (err) {
    debug(key + ": Client.get err=" + err);
  }

  // Cache hit
  if(response) {
    debug(key + ": Cache hit=" + response);
    // handle valid cache
  }

  debug(key + ": No response found");
}
```

Any idea what would cause `client.get` to not call the callback? If it helps I'm using Heroku and Redis To Go, and our Redis instance appears to be healthy, using under the limit memory and connections.

BridgeAR:
This simplifies the logic in the parser, removes dead code and most important:
fixes #902 and #875.

This will likely also fix some random tests failing on windows with a timeout.

### DIFF
--- a/lib/parsers/javascript.js
+++ b/lib/parsers/javascript.js
@@ -22,12 +22,8 @@ JavascriptReplyParser.prototype._parseResult = function (type) {
 
     if (type === 43 || type === 58 || type === 45) { // + or : or -
         // up to the delimiter
-        end = this._packetEndOffset() - 1;
+        end = this._packetEndOffset();
         start = this._offset;
-
-        if (end > this._buffer.length) {
-            throw new IncompleteReadBuffer('Wait for more data.');
-        }
 
         // include the delimiter
         this._offset = end + 2;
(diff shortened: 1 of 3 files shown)


lib/parsers/javascript.js
```diff
@@ -22,13 +22,9 @@ JavascriptReplyParser.prototype._parseResult = function (type) {
 
     if (type === 43 || type === 58 || type === 45) { // + or : or -
         // up to the delimiter
-        end = this._packetEndOffset() - 1;
+        end = this._packetEndOffset();
         start = this._offset;
 
-        if (end > this._buffer.length) {
-            throw new IncompleteReadBuffer('Wait for more data.');
-        }
```
BridgeAR: This is dead code, because the packetEndOffset is already checking the same.

PR #909, opened 2015-10-29
ignore if info command is not available on server

ivanB1975:
I added some lines to ignore if the "info" command is not available on the server for security reason.


BridgeAR:
Please add a test and use the same style as used in the rest of the code.

Otherwise LGTM

ivanB1975:
Sure, I will see if I have time today.

BridgeAR:
Thx, I merged your commits and it will be released with 2.3

ivanB1975:
You're welcome :)

PR #912, opened 2015-10-30
Fix return_buffers in pub/sub mode

### LINKED ISSUE #911: return_buffers with pub/sub is broken since 2.2.0, detect_buffers - since 2.0.0
I wrote a simple code to demonstrate the bug. As return_buffers works in 2.1.0, var msg is not a buffer in 2.2.0 and later (it's just a string). Same with detect_buffers, but it's broken since 2.0.0.

```js
var redis = require('redis');
var bufferEquals = require('buffer-equals');
var pub = redis.createClient();
var sub = redis.createClient({return_buffers: true});
var sub2 = redis.createClient({detect_buffers: true});
var buff = new Buffer('testtest');
sub.subscribe('test', function() {
  pub.publish('test', buff);
});
sub.on('message', function(pattern, msg) {
  console.log('return_buffers');
  if (Buffer.isBuffer(msg) && bufferEquals(buff, msg)) {
    console.log('it\'s a buffer and it\'s correct');
  } else if (Buffer.isBuffer(msg)) {
    console.log('it\'s a buffer, but damaged');
  } else {
    console.log('it\'s not a buffer');
  }
});
sub2.subscribe('test2', function() {
  pub.publish('test2', buff);
});
sub2.on('message', function(pattern, msg) {
  console.log('detect_buffers');
  if (Buffer.isBuffer(msg) && bufferEquals(buff, msg)) {
    console.log('it\'s a buffer and it\'s correct');
  } else if (Buffer.isBuffer(msg)) {
    console.log('it\'s a buffer, but damaged');
  } else {
    console.log('it\'s not a buffer');
  }
});
```

It breaks socket.io-redis, as you can see here socketio/socket.io-redis#17

``return_buffers`` and ``detect_buffers`` works as expected with ``get`` and most likely other commands, so the problem is probably related only to pub/sub.

komachi:
Fix #911

### DIFF
--- a/index.js
+++ b/index.js
@@ -645,7 +645,7 @@ RedisClient.prototype.return_reply = function (reply) {
         }
     } else if (this.pub_sub_mode || command_obj && command_obj.sub_command) {
         if (Array.isArray(reply)) {
-            if (!command_obj || command_obj.buffer_args === false) {
+            if ((!command_obj || command_obj.buffer_args === false) && !this.options.return_buffers) {
                 reply = utils.reply_to_strings(reply);
             }
             type = reply[0].toString();
--- a/test/return_buffers.spec.js
+++ b/test/return_buffers.spec.js
@@ -7,7 +7,7 @@ var redis = config.redis;
 
 describe("return_buffers", function () {
 
-    helper.allTests(function(parser, ip) {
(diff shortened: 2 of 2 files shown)


BridgeAR:
Thx! You were faster than me :) I was just working on it :D

PR #925, opened 2015-11-27
emit event "ready" when disable_resubscribing is true

maxgalbu:
I have two connections to redis, one for subscribe() and the second one for other commands. If I set `disable_resubscribing` to true and redis goes down, when it goes back up, the `ready` event is triggered for the query connection but not for the subscribe connection, and I want to do something when the connection is available again.


BridgeAR:
Thanks a lot! :+1:

PR #942, opened 2016-01-03
upgrade to nyc@5.2.0 (which has several fixes for Windows)

bcoe:


BridgeAR:
I guess it's about the cache flag as the version was already covered earlier? Feel free to merge it.

bcoe:
@BridgeAR thought I'd try out the `--cache` flag for performance, the actual important update is getting rid of:

`./node_modules/.bin/_mocha`, the new library we're using for spawning doesn't like absolute paths on Windows.

BridgeAR:
I see :+1:

PR #956, opened 2016-01-20
Adds config variable `enable_offline_queue_until_first_ready`

benjaminpearson:
Configuration variable added that allows you to enable the offline queue only while the connection is being opened, before it first becomes ready. After it is ready, if you have `enable_offline_queue` set to false then it will stop queuing. Useful for when you don’t want an offline queue during normal operation (ie, you are using redis as a cache that has a fallback to a primary data source) but you do want it to queue requests initially when loading the connection without havingto listen to the ‘ready’ event being emitted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,13 @@ connection to the redis server, commands are added to a queue and are executed
 once the connection has been established. Setting `enable_offline_queue` to
 `false` will disable this feature and the callback will be executed immediately
 with an error, or an error will be emitted if no callback is specified.
+* `enable_offline_queue_until_first_ready`: *false*; To be used in conjunction with `enable_offline_queue`,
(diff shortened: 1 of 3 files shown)


benjaminpearson:
Sorry I need to add more tests to this PR before it being ready, I had intended to open the PR on my fork rather than the main repo as it was just an idea I was thinking about for our intended use of the redis module. It may be helpful for others though...

shezarkhani:
:+1:

BridgeAR:
Thanks a lot for your contribution!

Right now I try to reduce the amount of options though. There will be a new option soon to handle reconnections easier and better than right now and it'll replace a couple other options that are used for reconnection purposes. You will be able to handle this in there easily but until than you can achive what you want by doing the following:

'''js
client = redis.createClient ();
client.once ('connect', () => {
    this.max_attempts = 1;
});
'''

PR #995, opened 2016-03-02
add support for json

lext-7:


BridgeAR:
Thanks a lot for your contribution! This was likely a bit of work to fully implement this as it is but that part is strictly limited to the new commands. If something like auto stringify should be implemented I'd like to have that possibility for any command you want to use it for, as there are a couple more that would benefit from that.
Adding non standard commands also have the downside of needing quite a bit of documentation. The overhead of implementing this in your app right now should be fairly low too (you could wrap each command that you want to stringify).
After all, I intend to implement more powerful transformations that would allow something like this to be implemented for each command a user wants it for. So please keep an eye out for new releases.

BridgeAR:
Closed as of the already mentioned reasons

PR #1031, opened 2016-04-08
Noting in documentation when client.duplicate() might be used

BrianRossmajer:
Explicitly pointing out that when blocking and non-blocking functions are used on the same connection, unexpected results can occur

### DIFF
--- a/README.md
+++ b/README.md
@@ -618,6 +618,38 @@ the second word as first parameter:
 
 Duplicate all current options and return a new redisClient instance. All options passed to the duplicate function are going to replace the original option.
 
+An example of when to use duplicate() would be to accomodate the connection- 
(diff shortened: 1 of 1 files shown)


BridgeAR:
@BrianRossmajer this looks mainly good but I don't like to provide negative examples. 

Would you be so kind and rewrite the example accordingly?

BridgeAR:
Ping @BrianRossmajer

BrianRossmajer:
Hi Ruben,

Sorry for the delay; I intend to do this but haven't had time.  Should be
in the next few days.

Thanks,
Brian

On Mon, Apr 18, 2016 at 3:50 PM, Ruben Bridgewater <notifications@github.com

PR #1032, opened 2016-04-10
Fixed function name case

benlowry:
### Pull Request check-list

### Description of change

Function name was written in all caps.


BridgeAR:
@benlowry node_redis officially supports all caps and this is the reason why some commands in the tests are actually written that way.

Thanks a lot for your contribution though.

PR #1038, opened 2016-04-19
Add backticks and quotes where appropriate in README.md

jonathanp:


BridgeAR:
@NodeRedis/contributors please weight in for this.

stockholmux:
It's a stylistic issue. We should be consistent in use of backticks when documentation, but I doubt that it creates any confusion as-is. I'm not crazy about using backticks in headings - it makes for strange reading. 

The connection events section needs better documentation in general (something I've been thinking about tackling). Perhaps moving to the table style documentation for the client connection section (as we just did for the `options` object.

BridgeAR:
@jonathanp I agree with @stockholmux that the backticks in headings make it more difficult to read and that table style would likely be the best to go for. Would you be up to change this accordingly? :smiley:

PR #1047, opened 2016-04-27
Adding send_command_buf function sending reply as Buffers instead of Strings

vitaliylag:
Adding send_command_buf function. The same as send_command but reply will be sent to callbacks as Buffers instead of Strings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,12 @@ client.get("missingkey", function(err, reply) {
 });
 ```
 
+If you want receive reply as Buffers instead of Strings just add "b_" prefix before command name:
+
+```js
+client.b_get("some key", function(err, buf) {});
+```
+
 For a list of Redis commands, see [Redis Command Reference](http://redis.io/commands)
 
 Minimal parsing is done on the replies. Commands that return a integer return JavaScript Numbers, arrays return JavaScript Array. `HGETALL` returns an Object keyed by the hash keys. All strings will either be returned as string or as buffer depending on your setting.
@@ -665,6 +671,10 @@ you can use `send_command()` to send arbitrary commands to Redis.
 The command_name has to be lower case.
 
 All commands are sent as multi-bulk commands. `args` can either be an Array of arguments, or omitted / set to undefined.
+
+## client.send_command_buf(command_name[, [args][, callback]])
+
+The same as client.send_command but is sending reply as Buffers instead of Strings.
 
 ## client.connected
 
--- a/index.js
+++ b/index.js
@@ -294,10 +294,7 @@ RedisClient.prototype.create_stream = function () {
 
 RedisClient.prototype.handle_reply = function (reply, command, buffer_reply) {
     if (!buffer_reply && this.using_buffer_parser) {
-        // Reply from parser will be Buffer if:
-        // 1) return_buffers option set to true
-        // 2) or detect_buffers option set to true and command used Buffer arguments
-        // 3) or buffer_reply argument was set to true when calling internal_send_command
+        // If nobody wants buffers for any reasons then we must convert them to strings
         reply = utils.reply_to_strings(reply);
     }
 
@@ -870,9 +867,8 @@ RedisClient.prototype.internal_send_command = function (command_obj) {
     var big_data = false;
     var args_copy = new Array(len);
 
-    if (this.options.return_buffers || this.cur_command_ret_buf) {
-        // this.cur_command_ret_buf check is needed for send_command_buf and "b_" prefixed individual commands
-        // (but not for standart "b_" prefixed command)
+    if (this.cur_command_ret_buf) {
+        // Needed for send_command_buf and overwritten individual commands with "b_" prefix
         command_obj.buffer_reply = true;
     }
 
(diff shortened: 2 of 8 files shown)


BridgeAR:
The general idea is fine, but I would not add a new function, but instead add a new parameter to the already existing function and more important: add a buffer function for each command.

It won't work though as it's currently suggested. The function would have to make sure that the parser returns buffers from that point on without changing the results of any other call no matter what option is set.

And using a try catch block around send_command destroys the purpose of the parameter validation.

vitaliylag:
> And using a try catch block around send_command destroys the purpose of the parameter validation.

I don't use try catch, I use try finally.

> but instead add a new parameter

Calling command not added to node_redis yet will be uncomfortably because parameters will be too difficult.

> add a buffer function for each command

I think great idea. And delete uppercase commands. If somebody use them he can just copy them by himself. Or you just can add Redis.createUpperCaseCommands() function that will copy all commands. Actually you can also add Redis.createBufferCommands() function. If need I can implement them.

> It won't work though as it's currently suggested. The function would have to make sure that the parser returns buffers from that point on without changing the results of any other call no matter what option is set.

Sorry, I didn't understand (I guess my English is bad). «It won't work» — my function (send_command_buf)? «without changing the results of any other call» — my function doesn't change other calls replies.

BridgeAR:
> I don't use try catch, I use try finally.

True indeed, that works and my comment should be ignored. One thing about the try finally though: it might be worth thinking about not having any rescue plan in case of a misuse. The only way how this would not end up in a crash is by listening for unhandled exceptions and in that case the app should at least restart and recovering is not intended.
So guarding against the misuse is a performance hit that we do not have to take.

> Calling command not added to node_redis yet will be uncomfortably because parameters will be too difficult. If you mean internal_send_command only then it's OK (and of course wrapper function can be also added).

I agree that it's uncomfortable for users, so yes it should only be used for internal_send_command (using a new state can be prevented therefore).
This pull request is about having a nice API after all, as there's already the `detect_buffer` option that allows users to return buffers individually.

> I think great idea. And delete uppercase commands. If somebody use them he can just copy them by himself. Or you just can add Redis.createUpperCaseCommands() function that will copy all commands. Actually you can also add Redis.createBufferCommands() function. If need I can implement them.

I'm +1 on removing the uppercase commands. That's already plannend, but this can only happen in the next major version. Adding the extra functions is not necessary in that case, as any user can just write an alias for the original to use it as uppercase command.

> > It won't work though as it's currently suggested. The function would have to make sure that the parser returns buffers from that point on without changing the results of any other call no matter what option is set.

> Sorry, I didn't understand (I guess my English is bad). «It won't work» — my function (send_command_buf)? «without changing the results of any other call» — my function doesn't change other calls replies.

It's not about your function changing another call directly. But the default return value of node_redis is of type String. In that case, the parser setting has to be changed to return buffers. And then the handle_detect_buffers_reply has to be set as function for `handle_reply` to return strings in case of the non buffer functions. Otherwise all "normal" string functions would return buffers too. 

```js
const Redis = require('redis')
const assert = require('assert')
const client = Redis.createClient()

assert(!client.buffers) // Default is to return strings, therefore buffers is false

client.set('foo', 'bar')

// This has to trigger setting `client.buffers` to true and to reset the parser to return buffers
client.getBuffer('foo', function (err, res) {
  assert.strictEqual(res.inspect(), '<Buffer 62 61 72>')
})
// This has to keep on working even though the parser returns buffers
client.get('foo', function (err, res) {
  assert.strictEqual(res, 'bar')
})
```

vitaliylag:
sorry, I don't know how to use github, not pay attention

vitaliylag:
New code:

1. Working propertly even if detect_buffers and return_buffers was set to false.
2. Added support for OfflineCommand constructor.
3. Added "b_" prefixed commands. For example redis.b_het("h", "a", callback);

Code changing:

1. this.buffers property is renamed to this.using_buffer_parser.
2. argument buffer_args is renamed to buffer_reply. It sets to true if:
 - this.options.return_buffers
 - or buffer_args && this.options.detect_buffers
 - or buffer_reply
 - (see internal_send_command function)
 - That means that we don't need to check options on reply anymore, we just need check buffer_reply only.

vitaliylag:
Now it is support multi and batch. Also I have added tests.

vitaliylag:
Now it supports individual commands.

BridgeAR:
I refactored passing the command_obj to be created in a earlier stage in #1074. If you rebase, this should allow to simplify some of the code.

vitaliylag:
How to rebase?

BridgeAR:
I just released v.2.6 and you can rebase to the master instead of the other branch.

Just write something like:
```
git remote add upstream https://github.com/NodeRedis/node_redis.git
git fetch upstream
git rebase -i upstream/master
```
You'll get all the commits from the master and can rebase your against them to solve the issues individually. This is nicer than using a merge. And when you're done you can push your changed history to this branch.

vitaliylag:
rebased and improved tests

vitaliylag:
@BridgeAR, what about merge?

vitaliylag:
is the project alive?

stockholmux:
@vitaliylag Yep. Although I haven't seen Ruben around in a while. I'm assuming this commit is no longer valid, right? It's 18 months old.

Any particular way you'd like to help out?

vitaliylag:
I didn't see any major changes last 18 months so I guess nobody maintain the project.

> I haven't seen Ruben around in a while
So the project is not alive I guess... :)

> Any particular way you'd like to help out?
No. If I were offered to become second admin I would accept it to merge my pull requests without any problems (in my server I use version with promises support so it is second not published PR) but I wouldn't maintain the project.

BridgeAR:
I am going to look into this at some point soonish. I am aware that I did not invest enough time in here but I will get to that soon again and finally get some stuff moving again.

PR #1055, opened 2016-05-04
Do you mean snake_case instead of snack_case?

kevinresol:
### Description of change

Fix typo in README


BridgeAR:
Thanks a lot. Seems like I thought of having a snack while writing that : laughing:

PR #1076, opened 2016-06-01
Ensure synthetic function names conform to naming requirements

danmactough:
### Pull Request check-list

### Description of change

The "restore-asking" function name is not valid and was causing co-redis (by way of its usage of thenify) to throw because thenify uses the function name to rewrite async functions with promises.

This PR will change the name of the "restore-asking" function to "restore_asking", which is valid.

This sanitation is a bit stricter than necessary, since it also sanitizes valid unicode characters, but it covers this module's potential use cases just fine.

### DIFF
--- a/lib/commands.js
+++ b/lib/commands.js
@@ -21,6 +21,7 @@ var changeFunctionName = (function () {
 // that provided a functionality to add new commands to the client
 
 commands.list.forEach(function (command) {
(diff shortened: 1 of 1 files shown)


BridgeAR:
@danmactough Thanks a lot!

danmactough:
my pleasure 😸 @BridgeAR

PR #1087, opened 2016-06-14
Noting in docs that retry_max_delay and connect_timeout are deprecated

Austintacious:
Pretty straightforward. The current README doesn't denote that 'retry_max_delay' and 'connect_timeout' are deprecated, while it does make this distinction for 'max_attempts'.


BridgeAR:
Thanks a lot

PR #1101, opened 2016-07-09
Handle hscan reply as an object

izeau:
### Pull Request check-list
### Description of change

Handle the results of `HSCAN` as an object, as node_redis does for `HGETALL`

### DIFF
--- a/test/commands/hscan.spec.js
+++ b/test/commands/hscan.spec.js
@@ -1,0 +1,45 @@
+'use strict';
+
+var config = require('../lib/config');
+var helper = require('../helper');
+var assert = require('assert');
+var redis = config.redis;
+
+describe("The 'hscan' method", function () {
+
+    helper.allTests(function (parser, ip, args) {
+
+        describe('using ' + parser + ' and ' + ip, function () {
+            var client;
+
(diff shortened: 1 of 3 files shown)


BridgeAR:
I totally understand that this would be convenient for you, but this is a BC and I would like to have a more general way of handling something like this.

test/commands/hscan.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+            it('return values', function (done) {
+                if (helper.redisProcess().spawnFailed()) this.skip();
+                helper.serverVersionAtLeast.call(this, client, [2, 8, 0]);
+                var hash = {};
+                for (var i = 0; i < 500; i++) {
+                    hash['key_' + i] = 'value_' + i;
+                }
+                client.hmset('hash:1', hash);
+                client.zscan('hash:1', 0, 'MATCH', 'key_*', 'COUNT', 500, function (err, res) {
```
fabiocarneiro: zscan?

fabiocarneiro:
I wish this had gone through. It's really not cool to receive keys and values alternated in a list.

PR #1104, opened 2016-07-14
Convert snake_case options to camelCase before sending into tls || net (FIXES #1103)

### LINKED ISSUE #1103: TLS self signed certificate error on version>2.6.0-1
```
const path = require('path');
const fs = require('fs');
redis.createClient('redis_url', {
  tls: {
    requestCert: true,
    rejectUnauthorized: false,
    cert: fs.readFileSync(path.join(__dirname, 'cert.crt')),
    key: fs.readFileSync(path.join(__dirname, 'key.key'))
  },
  password: 'password'
})
  .once('error', (err) => { throw err; })
  .once('ready', function () {
    console.log('bingo');
    this.quit();
  });
```
- **Version**: >2.6.0-1
- **Platform**: Node.js 6.2.1 on Windows 8
- **Description**: Throws _Error: Redis connection to redis_url failed - self signed certificate_, bingo on 2.6.0-1

dab00:
Convert snake_case options to camelCase before sending into tls || net (FIXES #1103)

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,9 +24,9 @@ if (typeof EventEmitter !== 'function') {
     EventEmitter = EventEmitter.EventEmitter;
 }
 
-function noop() { }
+function noop () {}
 
-function handle_detect_buffers_reply(reply, command, buffer_args) {
+function handle_detect_buffers_reply (reply, command, buffer_args) {
     if (buffer_args === false || this.message_buffers) {
         // If detect_buffers option was specified, then the reply from the parser will be a buffer.
         // If this command did not use Buffer arguments, then convert the reply to Strings here.
@@ -43,7 +43,7 @@ exports.debug_mode = /\bredis\b/i.test(process.env.NODE_DEBUG);
 
 // Attention: The second parameter might be removed at will and is not officially supported.
 // Do not rely on this
-function RedisClient(options, stream) {
+function RedisClient (options, stream) {
     // Copy the options so they are not mutated
     options = utils.clone(options);
     EventEmitter.call(this);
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -237,11 +237,19 @@ RedisClient.prototype.create_stream = function () {
             this.stream.destroy();
         }
 
+        // Accept snake_case options and convert them to camelCase
+        var opts = {};
+        var keys = Object.keys(this.connection_options);
+        var key;
+        while (key = keys.pop()) {
+            opts[key.replace(/_([a-z])/g, function (g) { return g[1].toUpperCase(); })] = this.connection_options[key];
+        }
```
BridgeAR: This is going to run on every reconnect attempt. Instead, the fix could be placed where the options are converted to camel_case. That would only run once for each client. The clone function could receive a extra argument to skip the converting tls options to camel_case and special handle that by checking if the key name is `tls`.

dab00: I don't understand the sacred meaning of conversion to snake_case so I wrote my code right before sending connection options into ```tls``` or ```net```, the same way it was converted. Sorry, this code is not in my style. May be this quest is for the person who has implemented the conversion. Do it as you see fit please.

BridgeAR: The reason for the conversion is that node_redis was written in snake_case and camelCase support was added by converting these values to snake_case and therefore both is supported.

dab00: In my opinion the solution is to leave ```connection_options``` as is, I mean do not ```clone``` it.
Instead of:
```
function RedisClient (options, stream) {
    // Copy the options so they are not mutated
    options = utils.clone(options);
    EventEmitter.call(this);
    var cnx_options = {};
    ...
}
```
Something like:
```
function RedisClient (options, stream) {    
    EventEmitter.call(this);
    var cnx_options = {};
    ...
    // Copy the options so they are not mutated
    options = utils.clone(options);
    ...
}
```

BridgeAR: The options are already cloned before they are passed to the constructor. This is done in the createClient function.

BridgeAR:
@dab00 thanks a lot for your contribution! And sorry it took so long...

Issue #1159, opened 2016-10-15
Uncaught exception:  TypeError: Cannot read property 'code' of null     at Object.redis.createClient.retry_strategy 


huaconghub:
```js
var client = redis.createClient({
    retry_strategy: function (options) {
        if (options.error.code === 'ECONNREFUSED') {
            // End reconnecting on a specific error and flush all commands with a individual error
            return new Error('The server refused the connection');
        }
        if (options.total_retry_time > 1000 * 60 * 60) {
            // End reconnecting after a specific timeout and flush all commands with a individual error
            return new Error('Retry time exhausted');
        }
        if (options.times_connected > 10) {
            // End reconnecting with built in error
            return undefined;
        }
        // reconnect after
        return Math.max(options.attempt * 100, 3000);
    }
});
```
throw exception:
```
Uncaught exception:  TypeError: Cannot read property 'code' of null     at Object.redis.createClient.retry_strategy 
```
log out options:
```json
{ 
  attempt: 1,
  error: null,
  total_retry_time: 0,
  times_connected: 1
 }
```

fruch:
I'm also trying to use the retry_strategy, even as minimal as 
```
        retry_strategy: function (options) {
                console.log(options);
                // reconnect after
                return 10000;
            }
```
seems not to work, my process is crashing like that:
```
 AbortError: Redis connection lost and command aborted. It might have been processed.
     at RedisClient.flush_and_error (/app/node_modules/redis/index.js:350:23)
     at RedisClient.connection_gone (/app/node_modules/redis/index.js:652:14)
     at Socket.<anonymous> (/app/node_modules/redis/index.js:286:14)
     at Socket.g (events.js:291:16)
     at emitNone (events.js:91:20)
     at Socket.emit (events.js:185:7)
     at endReadableNT (_stream_readable.js:974:12)
     at _combinedTickCallback (internal/process/next_tick.js:74:11)
     at process._tickDomainCallback (internal/process/next_tick.js:122:9)
```

jdegger:
This has just been committed: https://github.com/NodeRedis/node_redis/commit/4b27f798d887a2b6b1db5cb01463660594a76cc3

PR #1168, opened 2016-11-04
Check options.error

katoken-0215:
### Pull Request check-list

### Description of change

options.error it sometimes null.
Example code should check it before reference options.error.code.
(See also #１１５９ )


katoken-0215:
Two items are not checked, but this change is just README.md.
Let me omit them ?

PR #1171, opened 2016-11-10
Update retry_strategy example

novemberborn:
### Pull Request check-list

### Description of change

Presumably `Math.min()` was intended, otherwise the `options.attempt * 100` result would be ignored for the first 30 attempts. IMHO the initial retries should be quick, but then the delay should be capped at a reasonable interval.


BridgeAR:
Thanks a lot! It was indeed a mistake

PR #1173, opened 2016-11-17
enabled adding abritary commands

stockholmux:
### Pull Request check-list

- [ X ] Does `npm test` pass with this change (including linting)? _(test results have same 5 failures as master branch)_

### Description of change

I was working a bit with Redis modules and noticed that we don't have a way to add commands to the library at runtime. I just had to do a little refactoring to enable new commands. To add a command:

`redis.addCommand("nr.run");`

(using neural redis as an example)

I haven't written a test for this - I'm a bit stumped on how to test.


stockholmux:
@BridgeAR Thoughts on this one? I'd like to push it forward.

index.js
```diff
@@ -1090,4 +1090,6 @@ exports.AggregateError = errorClasses.AggregateError;
 // Add all redis commands / node_redis api to the client
 require('./lib/individualCommands');
 require('./lib/extendedApi');
-require('./lib/commands');
+
+//enables adding new commands (for modules and new commands)
+exports.addCommand = require('./lib/commands')
```
BridgeAR: Currently the default is still to use underscores. But we support both, so please add a alias for it with underscore.

lib/commands.js
```diff
@@ -17,18 +17,14 @@ var changeFunctionName = (function () {
...
+var addCommand = function (command) {
     // Some rare Redis commands use special characters in their command name
     // Convert those to a underscore to prevent using invalid function names
     var commandName = command.replace(/(?:^([0-9])|[^a-zA-Z0-9_$])/g, '_$1');
 
     // Do not override existing functions
-    if (!RedisClient.prototype[command]) {
-        RedisClient.prototype[command.toUpperCase()] = RedisClient.prototype[command] = function () {
+    if (!RedisClient.prototype[commandName]) {
+        RedisClient.prototype[commandName.toUpperCase()] = RedisClient.prototype[commandName] = function () {
```
BridgeAR: This is a minimal changed behavior. The `command` is the true redis name and the `commandName` would fail in that case.

BridgeAR:
@stockholmux two small nits and please add some tests. Otherwise LGTM

lib/commands.js
stockholmux: Ok - this bit of code tripped me up. I think the way it is currently implemented isn't correct. In the current master the `commands.js` has the following bits (save for the `console.log` I inserted and noted below):
```
    // Some rare Redis commands use special characters in their command name
    // Convert those to a underscore to prevent using invalid function names
    var commandName = command.replace(/(?:^([0-9])|[^a-zA-Z0-9_$])/g, '_$1');
    console.log(command); //added to show all commands
    // Do not override existing functions
    if (!RedisClient.prototype[command]) { ...
```
The will show all the commands as they are added to the client - including two with strange characters (`host:` and `restore-asking`) which yields two functions with illegal characters `RedisClient.prototype['host:']` and `RedisClient.prototype['restore-asking']`. So shouldn't it actually be `RedisClient.prototype[commandName]` to yield `RedisClient.prototype.host_` and `RedisClient.prototype['restore_asking']`?

BridgeAR: So the reason is: Redis itself has a command with the name `restore-asking`. And if you want to use this command anyone should be able to call the function with exactly this name, not a different one. But to have a function name in a possible stack trace the name is added to the function itself. And if it contains illegal function name characters those characters are converted to a underscore.
Another solution would have been to special handle this command and to document it but this is something I'm not in favor of.

test/commands/addCommand.spec.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+var config = require('../lib/config');
+var helper = require('../helper');
+var redis = config.redis;
+var assert = require('assert');
+
+describe("The 'addCommand/add_command' method", function () {
+
+    helper.allTests(function (parser, ip, args) {
```
BridgeAR: You do not have to execute these tests for different kinds of options passed to the redis instance. This is only necessary in case the client might behave differently due to the different options (e.g. comparing the JS parser to the hiredis parser).

test/commands/addCommand.spec.js
```diff
@@ -0,0 +1,44 @@
...
+            var client;
+
+            beforeEach(function (done) {
+                client = redis.createClient.apply(null, args);
+                done();
+            });
+
+            it('camel case version exists', function (done) {
+                assert.strictEqual(typeof redis.addCommand, 'function');
+                return done(null);
```
BridgeAR: These tests only executing sync code, so you do not need a `done`.

test/commands/addCommand.spec.js
```diff
@@ -0,0 +1,44 @@
...
+            });
+            it('snake version exists', function (done) {
+                assert.strictEqual(typeof redis.add_command, 'function');
+                return done(null);
+            });
+            it('generates a new method for an added command', function (done) {
+                redis.addCommand('newcommand');
```
BridgeAR: Please check first that the command does not exist before adding it.

stockholmux:
Alright - restored the original `command` thing but added `commandName` aliases if they differ. It's important for Redis modules to be able to be accessed with JS-safe command names.

stockholmux:
@BridgeAR Are we good to go on this?

stockholmux:
@BridgeAR 4.0 is out (with modules) - are we gtg on merging this?

lib/commands.js
```diff
@@ -61,8 +57,12 @@ commands.list.forEach(function (command) {
             }
             return this.internal_send_command(new Command(command, arr, callback));
         };
+        //alias commands with illegal function names (e.g. NR.RUN becomes NR_RUN and nr_run)
+        if (commandName !== command) {
+            RedisClient.prototype[commandName.toUpperCase()] = RedisClient.prototype[commandName] = RedisClient.prototype[command];
+        }
```
stockholmux: I've read through your request a few times and I can't square what you're trying to say aside from reverting and removing the entire `if` block. However, removing the `if (commandName !== command) { ... }` block renders the code unusable. For example:
```js
redis.addCommand('ft.info');
client = redis.createClient(require('../localredis.json'));
console.log(client.multi()['FT.INFO']); //-> [Function: ft_info]
console.log(client.multi()['ft.info']); //-> [Function: ft_info]
console.log(client.multi().ft_info) //-> undefined
```

*All* modules are supposed to have a dot namespacing, so this is rather big deal. No one will want to use bracket notation.

BridgeAR: 1) It is not a illegal function name as it is also property and those may contain anything.
2) I would indeed prefer not to add the alternative writing but I now understand your intention behind it. If this should be added, it must also be documented. That should be the case for `addCommand` in general.

lib/commands.js
```diff
@@ -104,10 +104,18 @@ commands.list.forEach(function (command) {
             this.queue.push(new Command(command, arr, callback));
             return this;
         };
+        //alias commands with illegal function names (e.g. NR.RUN becomes NR_RUN and nr_run)
+        if (commandName !== command) {
+            Multi.prototype[commandName.toUpperCase()] = Multi.prototype[commandName] = Multi.prototype[command];
+        }
         if (changeFunctionName) {
-            Object.defineProperty(Multi.prototype[command], 'name', {
+            Object.defineProperty(Multi.prototype[commandName], 'name', {
```
BridgeAR: This should be kept as `command` in both cases.

BridgeAR:
Merged in 0437aa49851365bdf82711b7ff6ac7dc369ec7bb

stockholmux:
Thanks for the merge! I was going to respond to your requests but I've been pulling 12 hour days recently.

PR #1194, opened 2017-01-26
fix #1193

### LINKED ISSUE #1193: duplicate code in commands.js
There is duplicate code in `commands.js` that convert call to client (`RedisClient` or `Multi`) to a `Command` ([line 32](https://github.com/leibale/node_redis/blob/master/lib/commands.js#L32), [line 74](https://github.com/leibale/node_redis/blob/master/lib/commands.js#L74)).

leibale:
Delete duplicate code and export `callToCommand` function

### DIFF
--- a/lib/commands.js
+++ b/lib/commands.js
@@ -29,37 +29,7 @@ commands.list.forEach(function (command) {
     // Do not override existing functions
     if (!RedisClient.prototype[command]) {
         RedisClient.prototype[command.toUpperCase()] = RedisClient.prototype[command] = function () {
-            var arr;
-            var len = arguments.length;
-            var callback;
-            var i = 0;
-            if (Array.isArray(arguments[0])) {
-                arr = arguments[0];
-                if (len === 2) {
-                    callback = arguments[1];
-                }
-            } else if (len > 1 && Array.isArray(arguments[1])) {
-                if (len === 3) {
-                    callback = arguments[2];
-                }
-                len = arguments[1].length;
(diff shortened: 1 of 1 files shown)


BridgeAR:
The arguments would leak and this is a significant performance decrease.

leibale:
What? Can you elaborate?

BridgeAR:
```js
var Benchmark = require('benchmark')
var suite = new Benchmark.Suite()

function clone (args) {
  var arr = new Array(args.length)
  for (var i = 0; i < args.length; i++) {
    arr[i] = args[i]
  }
  return arr
}

function noLeak () {
  var arr = new Array(arguments.length)
  for (var i = 0; i < arguments.length; i++) {
    arr[i] = arguments[i]
  }
  return arr
}

function leaks () {
  return clone(arguments)
}

suite.add('Does not leak', function () {
  noLeak(1, 2, 3)
})

suite.add('Leaks', function () {
  leaks(1, 2, 3)
})

suite.on('cycle', function (event) {
  console.log(String(event.target))
})

suite.run({ delay: 1, minSamples: 150 })
``` 
```
Does not leak x 50,351,674 ops/sec ±0.72% (90 runs sampled)
Leaks x 19,793,341 ops/sec ±1.94% (90 runs sampled)
```
[More info](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers) about this topic.

BridgeAR:
Thanks for your effort though!

Issue #1198, opened 2017-02-18
retry_strategy doesn't reset after flushing the error


Janpot:
* **Version**: node_redis 2.6.5, redis 3.2.7
* **Platform**: What platform / version? Node.js 7.5.0 on Mac OS 10.12.3
* **Description**: 

I'm using the following code 

```js
const redis = require('redis');

const client = redis.createClient({
  url: 'redis://localhost:6379',
  retry_strategy: options => new Error('Fail')
});

setInterval(() => client.incr('some-key', console.log), 5000);
```

running against a local redis (`docker run --rm -p 6379:6379 redis`)

After I see the first output, I kill the redis to simulate disconnect. When I've seen the error come in I restart the redis. The connection doesn't come up again. I would expect that after flushing the error to the offline handlers, the client would try to reconnect on the next command. instead it stays in closed state. Also it's returning with `AbortError` instead of `new Error('Fail')`.

viglucci:
Via the documentation for retry_strategy, it appears that node_redis will only attempt to reconnect if the function returns a number. 

>If you return a number from this function, the retry will happen exactly after that time in milliseconds.

In your case you are returning a non number, which does not satisfy the requirements for attempting to reconnect.

> If you return a non-number, no further retry will happen and all offline commands are flushed with errors.

Janpot:
the flow I'm looking for is: on disconnect, retry connecting 3 times with 1 second interval. If at the third time it's still disconnect, it should return all outstanding commands with an error. I don't want it to stop trying to reconnect.

viglucci:
With a modified version of the example given in the README.md, you should be able to achieve what you are looking for. However, once the max attempts of 3 is exhausted, the client will not automatically attempt to reconnect on the next command. I believe you would need to manually call createClient() again.

```javascript
const client = redis.createClient({
    url: 'redis://localhost:6379',
    retry_strategy: (options) => {
        if (options.times_connected >= 3) {
            // End reconnecting after a specific timeout and flush all commands with a individual error
            return new Error('Retry attempts exhausted');
        }
        // reconnect after
        return 1000;
    }
});
```

Janpot:
Yes I left out the attempts since that doesn't make a difference here.

So basically you mean:
```js
let client = createClient();
function createClient () {
  return redis.createClient({
    url: 'redis://localhost:6379',
    retry_strategy: (options) => {
      client = createClient();
      return new Error('Retry attempts exhausted');
    }
  });
}
```
[...]

PR #1200, opened 2017-02-20
Support retry after flush

viglucci:
### Pull Request check-list

### Description of change

Allows for the retry_strategy to return a object which results in the command queue being flushed but also for a connection retry attempt to occur. This change is in response to #1198  .

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,10 +83,6 @@ function RedisClient (options, stream) {
             // Do not print deprecation warnings twice
             delete options.retry_max_delay;
         }
-        this.retry_on_flush = false;
-        if (options.retry_on_flush) {
-            this.retry_on_flush = options.retry_on_flush;
-        }
     }
 
     this.connection_options = cnx_options;
--- a/index.js
+++ b/index.js
@@ -83,6 +83,10 @@ function RedisClient (options, stream) {
             // Do not print deprecation warnings twice
             delete options.retry_max_delay;
         }
+        this.retry_on_flush = false;
+        if (options.retry_on_flush) {
+            this.retry_on_flush = options.retry_on_flush;
+        }
     }
 
     this.connection_options = cnx_options;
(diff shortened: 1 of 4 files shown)


viglucci:
I think there is some edge case with the AppVeyor tests that cause a slight miss timing and cause one of the tests to fail, I same a similar or the same error in a few of the other pull requests that are also marked as not successful.

test/connection.spec.js
```diff
@@ -322,8 +342,8 @@ describe('connection tests', function () {
                 });
 
                 it('retryStrategy used to reconnect with defaults', function (done) {
-                    var unhookIntercept = intercept(function () {
-                        return '';
+                    var unhookIntercept = intercept(function (text) {
+                        return "";
```
BridgeAR: This should be kept as it was before.

test/connection.spec.js
```diff
@@ -303,6 +303,26 @@ describe('connection tests', function () {
...
+                                client.set('foo', 'bar', function (err, res) {
+                                    assert.strictEqual(err.message, 'Stream connection ended and command aborted.');
+                                    assert.strictEqual(err.code, 'NR_CLOSED');
+                                    assert.strictEqual(err.origin.code, 'ECONNREFUSED');
+                                    done();
+                                });
+                            }
+                            return {
+                                retry_delay: Math.min(options.attempt * 25, 200),
+                                error: options.error
```
BridgeAR: It would be better to return a individual error. That way we'd know that the calling function indeed received that error.

index.js
```diff
@@ -616,14 +616,25 @@ RedisClient.prototype.connection_gone = function (why, error) {
             if (this.retry_delay instanceof Error) {
                 error = this.retry_delay;
             }
+            // Handle object pattern
+            if (typeof this.retry_delay === 'object' && !(this.retry_delay instanceof Error) && this.retry_delay !== null) {
```
BridgeAR: The `&& !(this.retry_delay instanceof Error)` is a redundant check. Please just use a `else if` instead.

test/connection.spec.js
viglucci: I'm not sure I follow. Do you mean just in the context of this test case, or in the implementation? With this change we have to return the error and the retry_delay, otherwise the reconnect is never attempted.

viglucci:
I adressed your comments above @BridgeAR ,  however that seems to have upset the CI :(.

PR #1201, opened 2017-02-21
Repurpose retry_strategy + add connection_strategy

viglucci:
### Pull Request check-list

### Description of change

Addresses #1198 as an alternative to #1200 .

This change adds `options.connection_strategy` which intends to replace the existing `options.retry_strategy`, as the existing `retry_strategy` is mostly concerned with handling re-connection efforts, and this change aims to add additional control regarding queue'd commands. In that aim, this change changes the purpose of `options.retry_strategy` to handle allow for control of queue'd commands. The `retry_strategy` now accepts the same arguments as it did before, however if it returns an instance of an `Error`, then the queues will be flushed with that error. This allows for queue flushing and connection attempt handling to be decoupled.

I can understand why re-purposing the existing `retry_strategy` and renaming to `connection_strategy` may not be preferred and would cause a breaking change / major version bump, so if that is an obstacle please let me know and I would be happy to rename `connection_strategy` back to `retry_strategy` and come-up with a new name for `retry_strategy` in the context of this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ coverage
 stunnel.conf
 stunnel.pid
 *.out
+.idea/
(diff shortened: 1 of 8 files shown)


BridgeAR:
I am in favor of the other solution

Janpot:
Personally I'm in favour of this solution. It's a simpler API. But it can be implemented as a non-breaking change too.

viglucci:
Closed in favor of pursuing #1200

Janpot:
Too bad. This is the simpler solution, he just shouldn't have changed the functionality of `retry_strategy`.

PR #1258, opened 2017-07-15
Don't assume port_arg is a string

lukechilds:
Fixes https://github.com/NodeRedis/node_redis/issues/1257

Don't assume `port_arg` is a string, the url string could of been passed in as a string or as the `url` property in an object.


BridgeAR:
Thanks a lot for the PR

PR #1259, opened 2017-07-17
Fix "therefor" vs. "therefore" typo in README.md

mattaudesse:
Hi there,

This is just a single character change to the documentation so I'll skip the PR template.

I've changed "therefor" in `README.md` to "therefor**e**" since I believe that must've been the author's intention and it seems more appropriate given the context. Both are valid English words, but their meanings differ. [See this stackexchange answer for a nice, succinct explanation of how](https://english.stackexchange.com/a/58617).

Thanks for `node_redis`!


stockholmux:
Good catch. Therefore, I will merge it ;)

PR #1277, opened 2017-10-09
document camel-case commands

dcharbonnier:
### Pull Request check-list

### Description of change
Documentation only, the convention for javascript functions and properties is camel case, those changes does not require any change, the documentation should just suggest the correct command alias


BridgeAR:
In general I would like to change the documentation as well but camelCase was added in v.2.6 and some people might still use older versions, so I am hesitant to merge this and would like to change the documentation in a major instead.

gkorland (approved):

gkorland (approved):

PR #1283, opened 2017-10-30
Add promisify example for node.js 8

1999:
### Pull Request check-list

### Description of change

_Please provide a description of the change here._

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,15 @@ return client.multi().get('foo').execAsync().then(function(res) {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


1999:
I'm pretty sure build fails because of smth different.

LiranBri:
oops, I just sent PR for the same purpose (https://github.com/NodeRedis/node_redis/pull/1284)
however, I think mine is more detailed a bit

1999:
Good catch. I think it's also worth it to have async/await example in your PR as well @LiranBri

LiranBri:
sure i'll add one

PR #1311, opened 2018-02-23
fix: client duplicate function now allows db param to be passed

### LINKED ISSUE #1281: client.duplicate(options) with options.db does not work
* **Version**: node_redis: 2.8.0, redis-server: 3.2.9
* **Platform**: Node.js 6.11.1
* **Description**: When duplicating a client object via existing one with different DB option, the new one's DB is same to previous one's DB.

```javascript
const redis1 = redis.createClient({ db: 1 }); // use DB 1. OK
const redis2 = redis1.duplicate({ db: 2 }); // use DB 1, not DB 2
```

I think that this code line is related to this issue.
https://github.com/NodeRedis/node_redis/blob/009479537eb920d2c34045026a55d31febd1edd7/lib/extendedApi.js#L98

markmcnelis:
### Pull Request check-list

### Description of change

When invoking the redis client duplicate function the new client will respect a new selected_db target if the db option param is passed.

Thoughts?

fixes #1281

### DIFF
--- a/lib/extendedApi.js
+++ b/lib/extendedApi.js
@@ -95,7 +95,7 @@ RedisClient.prototype.duplicate = function (options, callback) {
         existing_options[elem] = options[elem];
     }
     var client = new RedisClient(existing_options);
-    client.selected_db = this.selected_db;
+    client.selected_db = options.db || this.selected_db;
     if (typeof callback === 'function') {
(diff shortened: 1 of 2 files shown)


BridgeAR:
Thanks a lot for your contribution!

Issue #1313, opened 2018-03-04
LREM


TomPed:
Hello, Redis now supports `LREM`: https://redis.io/commands/lrem

How can we go about implementing this?

stockholmux:
Already supported. We support all Redis commands.

TomPed:
Okay cool! So sorry about that 😄.

stockholmux:
@TomPed Out of curiosity, how did you come to the conclusion that it wasn't supported?

TomPed:
1. I looked in `README.md` and did not see any usage of `lrem`
2. I searched the repo for `lrem`, and found this issue: https://github.com/NodeRedis/node_redis/issues/888
3. Noticed it was an old issue, and took a quick look at http://redis.js.org, found nothing quick about `lrem`
4. Did some Googling, found nothing
5. Then I came to the conclusion that it wasn't supported.

I think that's about it 😃.

TomPed:
@stockholmux

1. I looked in `README.md` and did not see any usage of `lrem`
2. I searched the repo for `lrem`, and found this issue: https://github.com/NodeRedis/node_redis/issues/888
3. Noticed it was an old issue, and took a quick look at http://redis.js.org, found nothing quick about `lrem`
4. Did some Googling, found nothing
5. Then I came to the conclusion that it wasn't supported.

I think that's about it 😃.

stockholmux:
Thanks, @TomPed.

The first line of the readme is:

"This is a complete and feature rich Redis client for node.js. **It supports all Redis commands** and focuses on high performance."

I don't say this because I'm trying to rub in that you didn't see it, but this is not the first issue of this type that we've received and I'm wondering if our documentation isn't effective because we aren't explicitly saying every command we support.

TomPed:
Yeah I gotcha, no worries. I did read that, but to be honest it did not give me the confidence that it was support, even though you state that it support *all* Redis commands. I think a Wiki on the API for the client would be useful, just so it shows what is supported and the expected commands. I did not know the commands to provide (even though it's in the exact same order as Redis' command) so I ended up Googling that too.

dirkbonhomme:
Maybe we can add all Redis commands in very small letters at the bottom of the readme?

e.g. "`Yes, we support all commands: set, lrem, hget, ...`"

TomPed:
[...]

PR #1321, opened 2018-03-16
docs: generate list of all commands into readme #1313

dirkbonhomme:
### Pull Request check-list

### Description of change

Includes a list of all Redis commands into the README and provides an npm command to re-generate this list in case there's an update.

Done as a "fun" response to #1313 so no hard feelings if this doesn't get merged. The situation there was painfully recognizable though because my first experience using this library was by going to the Github homepage and searching for the command I wanted to use.


stockholmux:
I had been thinking about this too. Great mind, eh?

I don't think, however, this is the answer, execution wise. Not that SEO should be our goal but this looks like a keyword link dump and Google hates those. So, here are my thoughts:

*  We don't really need to link to each command.
* I think a table might be more usable/readable. Perhaps 3 columns: command, supported (a green emoji check, which will be every row) and a third column for any notes, relevant for commands like HMSET. 
* The position on the page could be at the bottom right before the license/contributors. We could do an internal link at multiple parts inside the document.

As you might imagine this is not as simple as your solution, but I think it would yield good results.

DanielSeehausen:
using this in some capacity (@stockholmux suggestion looks great) would be a solid improvement to the docs. 

I imagine a lot of developers find themselves keyword searching for a package that implements a specific redis query. i.e., the only reason I found myself here is I was looking for the implementation of `setrange`

please don't let this die!

dirkbonhomme:
@DanielSeehausen Thanks for bringing this back to my attention, kinda forgot about it.

I've reviewed the tool based on @stockholmux' feedback:
https://github.com/NodeRedis/node_redis/blob/de18e0b9f941138b9ebd85f7920500337076e7a6/README.md#list-of-supported-commands

Some commands have notes with internal links to their explanation. Note sure if there are more commands that can or should have more explanation..

Any feedback?

BridgeAR:
This is a pretty huge list and takes a lot of space. I think moving this into a separate doc and adding a link to that would be best.

I feel that in general the main README should become smaller and not bigger. As soon as it's indeed small people will probably also find everything :-)

DanielSeehausen:
> @DanielSeehausen Thanks for bringing this back to my attention, kinda forgot about it.
> 
> I've reviewed the tool based on @stockholmux' feedback:
> https://github.com/NodeRedis/node_redis/blob/de18e0b9f941138b9ebd85f7920500337076e7a6/README.md#list-of-supported-commands
> 
> Some commands have notes with internal links to their explanation. Note sure if there are more commands that can or should have more explanation..
> 
> Any feedback?

I think it looks great -- the external docs are a good idea as well

DanielSeehausen:
@BridgeAR @dirkbonhomme wanted to ping to see if we were good with closing this, understanding external docs may be the way the wind is blowing!

dirkbonhomme:
@DanielSeehausen do you think it's useful if I update the tool to generate a new file `COMMANDS.md` (or other) and just add a link to that file from the README? Or abandon the idea altogether?

DanielSeehausen:
I think that is a great idea @dirkbonhomme ! +1 from me

dirkbonhomme:
It's now generating a separate COMMANDS.md file instead of listing all supported commands in the README: https://github.com/dirkbonhomme/node_redis/blob/master/COMMANDS.md

misamoo:
3

PR #1323, opened 2018-03-28
Update readme to reflect usage of .add_command

kraftman:
### Pull Request check-list

- [y] Does `npm test` pass with this change (including linting)?
- [y] Is the new or changed code fully tested?
- [y] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

### Description of change

I tried running client.add_command but it doesn't exist. It does exist on the redis import, and works, so I've updated the documentation to hopefully help people in the future.


stockholmux:
Thanks and good catch!

PR #1325, opened 2018-04-03
Fix travis ci testing

Apollon77:
### Pull Request check-list

- [n/a] Is the new or changed code fully tested?
- [n/a] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

### Description of change

Travis-CI has disabled IPv6 in their trusty images and thats why your tests failed since some time. because I rely on your work I thought it is a good idea to fix the testing. So I did :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
-sudo: false
+sudo: required
 env:
   - CXX=g++-4.8 TRAVIS=true
 addons:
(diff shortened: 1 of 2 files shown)


BridgeAR:
@Apollon77 thanks a lot! :+1: about the changed error message: I am not sure why the message is different on Windows and on Linux. I am going to investigate that.

Apollon77:
Oh, oversaw the windows testing (I also have appveyor) ... Interesting ... shpuld I change to allow tests to accept both?

BridgeAR:
@Apollon77 that would probably be the best. It does not matter if it is a specific message as long as we know that it returns the "appropriate" message.

Apollon77:
Done. Appveyor also green (in general, only a nodejs 0.10 timing issue happened)

BridgeAR:
Thabks a lot for your contribution!

PR #1352, opened 2018-06-10
Added documentation for REDIS optimistic lock system.

Fl4m3Ph03n1x:
### Description of change

Documentation update for optimistic locks using WATCH.

### DIFF
--- a/README.md
+++ b/README.md
@@ -704,6 +704,99 @@ significantly compared to firing the same commands in a loop without waiting for
 the result! See the benchmarks for further comparison. Please remember that all
 commands are kept in memory until they are fired.
 
+## Optimistic Locks
+
+Using `multi` you can make sure your modifications run as a transaction, but you 
+can't be sure you got there first. What if another client modified a key while 
+you were working with it's data?
+
+To solve this, Redis supports the [WATCH](https://redis.io/topics/transactions) 
+command, which is meant to be used with MULTI:
+
+```js
+var redis  = require("redis"),
+    client = redis.createClient();
+
+client.watch("foo", function( err ){
+    if(err) throw err;
+
+    client.get("foo", function(err, result) {
+        if(err) throw err;
+  
+        // Process result
+        // Heavy and time consuming operation here
+
+        client.multi()
+            .set("foo", "some heavy computation")
+            .exec(function(err, results) {
+                
+                /**
+                 * If err is null, it means Redis successfully attempted 
+                 * the operation.
+                 */ 
+                if(err) throw err;
+                
+                /**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -704,6 +704,99 @@ significantly compared to firing the same commands in a loop without waiting for
...
+Using `multi` you can make sure your modifications run as a transaction, but you 
+can't be sure you got there first. What if another client modified a key while 
+you were working with it's data?
+
+To solve this, Redis supports the [WATCH](https://redis.io/topics/transactions) 
+command, which is meant to be used with MULTI:
+
+```js
+var redis  = require("redis"),
+    client = redis.createClient();
```
stockholmux: Please add something to the arguments of `redis.createClient`, even if it's just an empty object indicating that something _should_ be there. No arguments means that you have no password on your Redis instance and it's a huge security issue, so we always try to illustrate best practices.

README.md
```diff
@@ -704,6 +704,99 @@ significantly compared to firing the same commands in a loop without waiting for
...
+        clients.alterer.quit();
+      });
+  },1000);
+});
+```
+
+**NOTE**: Redis WATCH does not work on fields of hashes and other objects. You can watch a hash
+to see if anything inside it was modified, but you cannot watch a specific hash field for a 
+modification.
+
```
stockholmux: A better way to put this is that Redis WATCH works on whole _keys_ - you can also mention the hash fields as it's a common misconception by beginners.

Maybe also mention that you can't watch individual set members, that way it is a bit more general and drives home the point that it's the whole _key's value_ being watched, no matter the type of the key.

stockholmux:
A few small nits but otherwise awesome. Make the changes and I'll merge. Thanks for your contribution!

Fl4m3Ph03n1x:
Alright, did the requested improvements. Let me know if everything is to your liking!

README.md
Fl4m3Ph03n1x: Done :D

stockholmux:
Thanks for your contribution!

Fl4m3Ph03n1x:
My pleasure, 
Only wish I could contribute more with more examples and snippets. 

All my projects have a project page and I think it would benefit this one also a lot, but I also understand your desire to keep documentation as minimal and necessary as possible to not replicate the already good documentation Redis has. 

Oh well, time to contribute to MDN and update my SO answer :P

funny-falcon-at-joomcode:
https://redis.io/topics/transactions
> Redis scripting and transactions
> A Redis script is transactional by definition, so everything you can do with a Redis transaction, you can also do with a script, and usually the script will be both simpler and faster.

PR #1353, opened 2018-06-13
fixed typo in readme example

kaija:
### Description of change

Fixed typo for protocol of redis_url


stockholmux:
No, that's correct. Both `redis://` and `rediss://` are valid (the latter being similar to `https://`

PR #1361, opened 2018-07-18
Remove clientError handler

bengl:
This event is never emitted on `net.Socket` or `tls.TLSSocket` streams in
any of the supported versions of Node.js.

### Pull Request check-list

  - Not applicable.
  - Not applicable.

### Description of change

It seems that this handler was added in eae5596, as part of adding support for TLS. While `tls.TLSServer` instances can emit `'clientError'` (prior to Node 6), `tls.TLSSocket` cannot.

### DIFF
--- a/index.js
+++ b/index.js
@@ -279,12 +279,6 @@ RedisClient.prototype.create_stream = function () {
         self.on_error(err);
     });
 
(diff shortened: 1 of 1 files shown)


BridgeAR:
@bengl thanks a lot!

PR #1380, opened 2018-10-10
Update README.md

rainabba:
Updated README to reflect the new way to define an expiration via the client (previously documented method threw an unrecognized command error).


rainabba:
Scratch this PR. Method doesn't work and I hadn't tested.

PR #1396, opened 2018-11-29
Add delay option for socket.setKeepAlive

betimer:
### Pull Request check-list

### Description of change

Add option for socket_initialdelay when calling socket.setKeepAlive from node API


betimer:
This is critical as because of this issue: https://github.com/nodejs/node/issues/24689

stockholmux:
I'm going to go ahead and merge, but I think we should investigate the root cause here as we might need a default that isn't 0.

betimer:
@stockholmux I strongly agree to have a value which is not the default 0.
Created this PR to add the option but does not change the existing behaviour.
I think something like 5000~60000 milliseconds would be fine.

I can make the change if this sounds OK, also may i know when we might get the new version?

stockholmux:
I consulted with some colleagues on this. I don't think there is actually a rational/universal initial delay default. I'm ok with this as-is.

betimer:
Nice. no worries, as long as this already provides an interface to whom, like us, facing the issue, to change the setting. Thank you @stockholmux

RidgeA:
Hi @betimer !
Is it possible to the latest version of the library from NPM?
As I can see the library version was updated on NPM more than a year ago...

Fundebug:
@anhzhi I found that idle redis connection will be disconnected every 2 hours, it is strange. It is solved by using your code:

```javascript
client.on('connect', function () {
    var socket = client.stream
    socket.setKeepAlive(true, 30 * 1000)
})
```

RidgeA:
This would work, but it is an ugly workaround. There is a special option for this, the package just need to be updated on npm. We have switched to ioredis package.

thameezb:
This contains a typo on the option. `socket_initialdelay` vs `socket_initial_delay` as used in other locations

PR #1404, opened 2019-01-03
Fix build failure by updating test case assertion

josnidhin:
Commit 34eac0db40967b231907a5835453da75de93a91f updated the error
message. This commit will update the test case to reflect that
change.

### Pull Request check-list

### Description of change

_Please provide a description of the change here._

### DIFF
--- a/test/unify_options.spec.js
+++ b/test/unify_options.spec.js
@@ -209,7 +209,7 @@ describe('createClient options', function () {
                 }, undefined);
                 throw new Error('failed');
             } catch (err) {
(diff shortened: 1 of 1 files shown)


stockholmux:
Erm.. Could you please separate out the linting (which largely seems to be indenting issues) from the test assertions? I'm glad to merge both of them but being under one PR makes it impossible to make a good judgement on what's changed and what is just style.

josnidhin:
I have updated the PR as you suggested

PR #1431, opened 2019-04-25
removing use of function.prototype.apply which internally uses queuin…

arpan-jain:
…g causing stack limit to be reached on large number of arguments

### Pull Request check-list

### Description of change

Earlier MULTI function was calling the commands using function.prototype.apply which has a hardcoded argument limit of [65536](https://bugs.webkit.org/show_bug.cgi?id=80797). 
for eg. If I wanted to add 1 mil keys in a set using a SADD command running REDIS transaction it throws a Maximum call stack size exceeded error.
```
client.multi(
['sadd',keyName,1,2,3,4,.......999999,1000000]
['expire',keyName,30000]
)
```
So, I have changed the function invocation which is not using function.prototype.apply.

### DIFF
--- a/lib/multi.js
+++ b/lib/multi.js
@@ -13,9 +13,9 @@ function Multi (client, args) {
             command = args[i][0];
             tmp_args = args[i].slice(1);
             if (Array.isArray(command)) {
-                this[command[0]].apply(this, command.slice(1).concat(tmp_args));
+                this[command[0]](command.slice(1).concat(tmp_args));
             } else {
-                this[command].apply(this, tmp_args);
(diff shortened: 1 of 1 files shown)


stockholmux:
This breaks callbacks (that's why it uses `apply`). I believe you can pass very high number of arguments by using the commands directly rather than arrays e.g.:
```
client.multi()
.sadd('foo',{ high number of arguments} )
.exec()
```

From a perspective of Redis itself, it's probably better to send multiple SADDs anyway.

I'm going to close this as an edge case where the fix breaks other functionality (and the limit is hard coded).

PR #1470, opened 2019-10-29
feat(parser): update to latest Redis parser & errors

WarmVissarutRonaldoDude:
### Pull Request check-list

### Description of change

- update `redis-parser` to `3.0.0` for removing `hiredis` to support node 12+.
- fix eslint errors.
- fix some of existing errors.
- add nodejs 12 to travis ci.
- add nodejs 12 to appveyor ci.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ node_js:
   - "4"
   - "6"
   - "8"
+  - "12"
 after_success: npm run coveralls
 before_script:
   # Add an IPv6 config - see the corresponding Travis issue
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,7 @@ environment:
     - nodejs_version: "4"
     - nodejs_version: "6"
     - nodejs_version: "8"
+    - nodejs_version: "12"
 
 pull_requests:
   do_not_increment_build_number: true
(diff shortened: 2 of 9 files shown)


WarmVissarutRonaldoDude:
@DanielSeehausen Hi!!

about removal `call` in `lib/customErrors` is because

`call` is not working with ES6 class constructor without `new keyword` and after look into the code for both `old` and `modern` in `redis-error` I saw it's only set `message` to `this.message` which we also set it in any `customErrors` here so I think we can just remove it instead of add complexity here to solving old and modern compat. :)

Reference: https://github.com/NodeRedis/redis-errors/blob/master/lib/modern.js

DanielSeehausen:
> @DanielSeehausen Hi!!
> 
> about removal `call` in `lib/customErrors` is because
> 
> `call` is not working with ES6 class constructor without `new keyword` and after look into the code for both `old` and `modern` in `redis-error` I saw it's only set `message` to `this.message` which we also set it in any `customErrors` here so I think we can just remove it instead of add complexity here to solving old and modern compat. :)
> 
> Reference: https://github.com/NodeRedis/redis-errors/blob/master/lib/modern.js

thank you for the explanation! sounds good

adylevy:
Can someone provide an update on this one? really need to get rid of hiredis...

zhangshirly:
Can someone provide an update on this one? really need to get rid of hiredis... please，i don't want to change to ioredis

PR #1492, opened 2020-02-06
[WIP] CI/Tests fixes

Salakar:


.gitignore
```diff
@@ -7,3 +7,14 @@ coverage
...
+
+# IntelliJ IDEs
+.idea
+
+# VisualStudioCode IDEs
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
```
BridgeAR: Nit: adding all IDE parts is not possible. Thus, it's normally ideal to let users ignore these globally on their local machine.

test/pubsub.spec.js
```diff
@@ -515,19 +515,15 @@ describe('publish/subscribe', function () {
...
+                                if (typeof pattern === 'string') {
                                     pattern = new Buffer(pattern);
                                     channel = new Buffer(channel);
                                 }
                                 assert.strictEqual(pattern.inspect(), new Buffer('*').inspect());
                                 assert.strictEqual(channel.inspect(), new Buffer('/foo').inspect());
                                 sub.quit(end);
                             });
-                            if (parser === 'javascript') {
-                                assert.notStrictEqual(sub.message_buffers, sub.buffers);
```
BridgeAR: Should this assertion not be kept?

.gitignore
Salakar: It's fairly common practice to add these in on OSS repositories, though I'll get rid of the `!.vscode*` lines as they're not required

test/pubsub.spec.js
Salakar: Good catch, would have given my code a review before merging don't worry 😅 

I've put this back locally however it causes a failure;

image


This line though doesn't make sense as to what it's testing, do you have any idea what it's trying to test?

Salakar: Figured out the issue

PR #1503, opened 2020-03-11
docs: clearify tls options behavior

brannondorsey:
### Description

Clearify that TLS will be attempted even if the `tls` option config is an empty object. 

The docs are somewhat ambiguous about this. We were looking to enable a TLS connection with the default options set, without having to build out a URL that includes `rediss://`. 

After looking at the code below it was apparent that we could acheive this behavior by specifying `tls: {}` in the client config options.

https://github.com/NodeRedis/node-redis/blob/bdc7356b379ab2b0b5bd62baefa4b4f1ec087891/lib/createClient.js#L36

https://github.com/NodeRedis/node-redis/blob/da31ade348f9b98ebc5a3164d59a93576a865906/index.js#L192-L193

This small PR adds a sentence to the description of the `tls` option indicating this default behavior.

---

### Checklist


brannondorsey:
Not sure there's anything I can do about that ESLint failure. Does anything else need to be done to make this PR merge ready?

leibale:
These files do not exist anymore...

Issue #1444, opened 2019-06-05
"total_retry_time" not include connection timeout


qishibo:
* **Version**: 2.8.0
* **Platform**: node v8.11.2 on ubuntu
* **Description**: Connection to an IP which can't be reached, emit a retry event after a very very long time during every retrying... What I meas the `total_retry_time` not include connecting time, and retry event will be emited after another connection timeout, if the `connection timeout` not setted, default is quite a long time.

```javascript
// connect to an IP not exists
redis.createClient(6379, '999.999.999.999', options);
```

I solved this with a `connect_timeout` param which is abandoned by official...this will solve the first time connecting timeout problems...

![image](https://user-images.githubusercontent.com/5136418/58927994-5a7b3400-8783-11e9-9b4d-2405beab4070.png)

[gitter]: https://gitter.im/NodeRedis/node_redis?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge

ljluestc:
```

const redis = require('redis');

// Redis connection options
const options = {
  // Set the socket timeout to 2 seconds
  socket_timeout: 2000, // Timeout in milliseconds
};

// Create Redis client with options
const client = redis.createClient({
  port: 6379,
  host: '999.999.999.999',
  ...options
});

// Event handler for connection error
client.on('error', (err) => {
  console.error('Redis connection error:', err);
});

// Event handler for retry
client.on('reconnecting', (details) => {
  console.log('Retrying connection...', details);
});

// Event handler for successful connection
client.on('ready', () => {
  console.log('Redis client connected and ready.');
});

// Event handler for end of connection
client.on('end', () => {
  console.log('Redis client connection closed.');
});
```

PR #1531, opened 2020-07-18
docs: improve docs on connect_timeout vs options.retry_strategy and some testing help

marnikvde:
### Description
It took me quite some time to figure out why the auto-reconnect stopped after 1 hour, even when trying all possible different retry_strategy options, which turned out to be undocumented connect_timeout parameter. The code comments mentions that this behaviour should be changed in v3.0, but apparently that has not been implemented yet, so my addition to the docs seems correct. Not sure if related to #1444.

I also added a bit more info to be able to run the tests, and fixed some whitespaces.

### Checklist


marnikvde:
Not sure what is wrong with the actions, I tried on my fork and the "Run ataylorme/eslint-annotate-action@1.0.4" works just fine there (I changed the "on pull request" to "push", but that's probably not the problem. Here it errors with "##[error]Resource not accessible by integration".

gkorland (approved):

PR #1554, opened 2020-12-16
Add a note about prefix not being used in pubsub

mcfedr:
Just a small docs update, I'm sure I'm not the first person to be surprised by this behaviour and its not documented.

For example, https://github.com/NodeRedis/node-redis/issues/1286


leibale:
These files do not exist anymore...

PR #1578, opened 2021-03-08
Upgrade node and dependencies

leibale:
Set node minimum version to >= 10 (currently the oldest active version),
Upgrade dependencies (including redis-commands),
Fix some tests


leibale:
ref #1562

gkorland (approved):

PR #1580, opened 2021-03-08
submit only master's coverage reports

leibale:


gkorland:
I don't think you need this PR, it should know maintain coverage history per branch

PR #1595, opened 2021-04-08
fix #1569 - improve monitor_regex

### LINKED ISSUE #1569: GHSL-2021-026
Hi,

GitHub Security Lab has found a potential vulnerability in node-redis. Please could you let us know how to contact you privately? We recommend adding a [SECURITY.md](https://docs.github.com/en/github/managing-security-vulnerabilities/adding-a-security-policy-to-your-repository) to your repository with a contact email address, which you can do by going to the [Security tab](https://github.com/NodeRedis/node-redis/security) and clicking the "Set up a security policy" button.

Kind regards,

Kevin Backhouse
GitHub Security Lab

leibale:


OnlineCop (commented):
The original, ending in `( ".+?")+$`, would fail to match:

    1234567890.0 [1 2]  "test"

(There are two spaces between the closing square bracket and the double quote.)

The new, ending in `.*"$`, will match that string (or any other string, so long as the last character on the line is a double quote).

Should this text match, or fail to match?

leibale:
@OnlineCop Basically you're right, but since this regex is used only to check if a message from Redis is a monitor message or a reply to a command, and the only message that starts with a decimal number is a reply from the monitor command, even the new regex is an overkill...

Issue #1593, opened 2021-04-05
On node versions higher than 10.19.0, a socket disconnect can permanently break the internal command_queue state


jakepruitt:
### Issue

The node-redis client has a race condition in the handling of write error events from the TCP socket if the TCP socket disconnects during the execution of the `internal_send_command()` function, specifically on versions of Node.js >= 10.23.2.

The issue stems from the way error events are emitted from net sockets between 10.19.0 and 10.23.2. Whereas socket `EPIPE` error events are emitted **asynchronously** on 10.19.0, they are emitted **synchronously** higher versions, which means the `internal_send_command` function may wind up executing the `this.flush_and_error()` function [within `connection_gone`](https://github.com/NodeRedis/node-redis/blob/master/index.js#L603) _before_ the `this.command_queue.push()` call [here](https://github.com/NodeRedis/node-redis/blob/master/index.js#L911). This means the command queue gets polluted with a stray command that throws off the synchronization between the command queue and the commands sent to redis once redis reconnects.

To be a bit more concrete, here is an example of the same socket disconnect in the same part of the code being handled on 10.19.0 (where the `.flush_and_error()` is called after `.command_queue.push()`) vs 10.23.2, where the `.flush_and_error` happens midway through the code in `internal_send_command` sending the write to redis.

On 10.19.0:

![Screen_Shot_2021-04-02_at_10_14_30_AM](https://user-images.githubusercontent.com/5084263/113437993-8924be80-939c-11eb-9c04-3d89c6f38c7f.png)

On node 10.23.2:

![Screen_Shot_2021-04-02_at_10_09_51_AM](https://user-images.githubusercontent.com/5084263/113437754-1287c100-939c-11eb-84f6-7687d93cd373.png)

I was not able to pinpoint the exact change in libuv that is causing the problem, but it was definitely introduced in libuv [between versions 1.25.0 and 1.34.2](https://github.com/libuv/libuv/compare/v1.25.0..v1.34.2#diff-4bb95a510382f076f33909e13714f15d52db24b57c3d11926bc8b27cc0378a18), since that is the difference between node 10.19.0 and 10.23.2. Without in-depth knowledge, I believe https://github.com/libuv/libuv/pull/2425 may have something to do with it.

### Reproducing

[...]

PR #1603, opened 2021-04-27
Wait until next tick before handling socket errors

jakepruitt:
### Description

Per #1593, there is a currently a race condition where an error in a `socket.write` call can cause the error handler (which clears the command queue) to be fired before the write command is added to the write queue, causing a permanent break in the command queue.

While this situation is rare (it occurs once out of 20 tries to reproduce it) it causes significant pain on downstream consumers, which expects node-redis to return the correct responses. Currently, unless the client implements extra verification that the value returned from node-redis is correct, the client can be working with the completely wrong data in their application.

> Description your pull request here

This change adds as little intrusion as possible - wrapping the `on_error` call in a `process.nextTick()` to ensure the assumption that the code makes about the socket error handling control flow will always be true. (The code currently assumes that the code inside `internalSendCommand` will execute to completion before any error handlers are called.)

---

### Checklist

  - [ ] Since this issue is so intermittent, I am not sure how to write an appropriate test. If anyone has advice, I would appreciate it.
- [x] ~~Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?~~ N/A

### DIFF
--- a/index.js
+++ b/index.js
@@ -220,7 +220,9 @@ RedisClient.prototype.create_stream = function () {
     });
 
     this.stream.on('error', function (err) {
-        self.on_error(err);
+        process.nextTick(function () {
+            self.on_error(err);
+        });
     });
 
     this.stream.once('close', function (hadError) {


misamoo (commented):
x

PR #1630, opened 2021-07-17
commands-queue.ts: String length / byte length counting issue

### LINKED ISSUE #1628: v4: Should send bytes instead of UTF-16 length
### Issue

> Describe your issue here

For Node Redis v4, the code seems to be using `String.prototype.length` to calculate the byte length of an argument to load on the payload
https://github.com/NodeRedis/node-redis/blob/c72aab2fc29a9cc994f6b810aeb10a4fe1863f29/lib/commands-queue.ts#L52-L53
It should load the byte length instead. `Buffer.byteLength(String.prototype)` is one of its ways.

---

### Environment

 - **Node.js Version**: `16.0.0`
 

 - **Redis Version**: `6.2.4`
 

 - **Platform**: `Linux MCPEPC-DESKTOP 5.8.0-59-generic #66~20.04.1-Ubuntu SMP Thu Jun 17 11:14:10 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

NovaAtWarren:
Hopefully fixes #1628

### Description

Fixes using string length instead of byte length when encoding command
### Checklist

- [x] ~~Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?~~ Nothing but bugfix, not needed

### DIFF
--- a/lib/commands-queue.ts
+++ b/lib/commands-queue.ts
@@ -50,7 +50,7 @@ export default class RedisCommandsQueue {
 
         for (let i = 1; i < args.length; i++) {
             const str = args[i].toString();
-            encoded.push(`$${Buffer.byteLength(str)}`, str);
+            encoded.push(`$${str.length}`, str);
         }
 
         return encoded.join('\r\n') + '\r\n';
--- a/lib/commands-queue.ts
+++ b/lib/commands-queue.ts
@@ -49,8 +49,7 @@ export default class RedisCommandsQueue {
         ];
 
         for (let i = 1; i < args.length; i++) {
-            const str = args[i].toString();
-            encoded.push(`$${Buffer.byteLength(str)}`, str);
+            encoded.push(`$${Buffer.byteLength(args[i])}`, args[i]);
         }
 
         return encoded.join('\r\n') + '\r\n';
--- a/lib/commands-queue.ts
+++ b/lib/commands-queue.ts
@@ -50,7 +50,7 @@ export default class RedisCommandsQueue {
 
         for (let i = 1; i < args.length; i++) {
             const str = args[i].toString();
-            encoded.push(`$${str.length}`, str);
+            encoded.push(`$${Buffer.byteLength(str)}`, str);
         }
 
         return encoded.join('\r\n') + '\r\n';
--- a/lib/commands-queue.ts
+++ b/lib/commands-queue.ts
@@ -44,7 +44,7 @@ export default class RedisCommandsQueue {
     static encodeCommand(args: Array<string>): string {
         const encoded = [
             `*${args.length}`,
-            `$${args[0].length}`,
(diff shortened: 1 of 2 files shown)


NovaAtWarren:
Just got my laptop on, did some testing with `\u{1f91e}`, and it seems it's already picking it up as 2 bytes with the old method.
As I've already written the test I'll leave this open, but I'll revert 2fa5ef6.

MCPE-PC (changes requested):
str.length is UTF-16 characters count

## Check 1.

Runkit

 - `'\u{1f91e}'.length !== Buffer.byteLength('\u{1f91e}');`
 - `본` is 3 bytes, `'본'.length` is 1

## Check 2.

UnicodePad

U+1F91E is an emoji and is 4 bytes

NovaAtWarren:
@MCPE-PC I've changed back to `Buffer.byteLength()`, and set the tests to look for a bytecount of 4, but it's now failing the test on my machine:
```#  1) Command Queue
       UTF-16 Byte length check (see #1628):

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ '*1\r\n$2\r\n🤞\r\n'
- '*1\r\n$4\r\n🤞\r\n'
          ^
```
I'm not quite sure yet, but it seems as if the different machines might be using different encoding standards (e.g `utf-8` and `utf-16`)
Not sure how much time I'll have today, but I'll look into this further.
Sorry for dismissing you earlier.

NovaAtWarren:
Turns out there were two places that calculated the length of a string
https://github.com/NodeRedis/node-redis/blob/c72aab2fc29a9cc994f6b810aeb10a4fe1863f29/lib/commands-queue.ts#L47
and
https://github.com/NodeRedis/node-redis/blob/c72aab2fc29a9cc994f6b810aeb10a4fe1863f29/lib/commands-queue.ts#L53

The fix you wrote worked perfectly fine, I just didn't encounter it because I put the special character in the wrong place to encounter the specific issue you did.

Thank you for the help, re-opening PR for review and merging.

lib/commands-queue.ts
```diff
@@ -44,13 +44,13 @@ export default class RedisCommandsQueue {
     static encodeCommand(args: Array<string>): string {
         const encoded = [
             `*${args.length}`,
-            `$${args[0].length}`,
+            `$${Buffer.byteLength(args[0])}`,
             args[0]
         ];
 
         for (let i = 1; i < args.length; i++) {
             const str = args[i].toString();
```
MCPE-PC, line 52:
```suggestion
```
As `args` is `string[]`, this line also seems to need to be reviewed. Is it casting `string` to a `string`?

MCPE-PC (approved):

leibale (approved):

leibale:
@NovaAtWarren @MCPE-PC thank you very much! :)

PR #1708, opened 2021-11-04
Typescript: Add rejectedUnauthorized and other TLS options

dajinchu:
### Description

I need to add `rejectUnauthorized: false` to the socket options but TypeScript will not allow it. 

Added `tls.CommonConnectionOptions` into the socket options type which allows the keys:

```
/**
* An optional TLS context object from tls.createSecureContext()
*/
 secureContext?: SecureContext | undefined;
/**
 * When enabled, TLS packet trace information is written to `stderr`. This can be
 * used to debug TLS connection problems.
 * @default false
 */
enableTrace?: boolean | undefined;
/**
 * If true the server will request a certificate from clients that
 * connect and attempt to verify that certificate. Defaults to
 * false.
 */
requestCert?: boolean | undefined;
/**
 * An array of strings or a Buffer naming possible ALPN protocols.
 * (Protocols should be ordered by their priority.)
 */
ALPNProtocols?: string[] | Uint8Array[] | Uint8Array | undefined;
/**
 * SNICallback(servername, cb) <Function> A function that will be
 * called if the client supports SNI TLS extension. Two arguments
 * will be passed when called: servername and cb. SNICallback should
 * invoke cb(null, ctx), where ctx is a SecureContext instance.
 * (tls.createSecureContext(...) can be used to get a proper
 * SecureContext.) If SNICallback wasn't provided the default callback
 * with high-level API will be used (see below).
 */
SNICallback?: ((servername: string, cb: (err: Error | null, ctx?: SecureContext) => void) => void) | undefined;
/**
 * If true the server will reject any connection which is not
 * authorized with the list of supplied CAs. This option only has an
 * effect if requestCert is true.
 * @default true
 */
rejectUnauthorized?: boolean | undefined;
```

### Checklist


leibale:
The file was moved to [~/packages/client/lib/client/socket.ts](https://github.com/redis/node-redis/blob/master/packages/client/lib/client/socket.ts), can you please update your branch?

dajinchu:
@leibale done

PR #1717, opened 2021-11-11
Support esModuleInterop set to false.

evert:
## Description

When testing the upcoming 4.x release, we got a bunch of typescript
errors emitted from this project.

We quickly realized this is because the library uses the esModuleInterop
flag. This makes some imports _slightly_ easier to write, but it comes
at a cost: it forces any application or library using this library to
*also* have esModuleInterop on.

The `esModuleInterop` flag is a bit of a holdover from an earlier time,
and I would not recommend using it in libraries. The main issue is that
if it's set to true, you are forcing any users of the library to also
have `esModuleInterop`, where if you keep have it set to `false` (the
default), you leave the decision to the user.

This change should have no ramifications to users with
`esModuleInterop` on, but it will enable support for those that have it
off, so it's the least opinionated option =)

This is especially good for library authors such as myself, because I
would also like to keep this flag off to not force *my* users into this
feature.

---

### Checklist


evert:
Note that this PR is not working yet. I'm having issues with the redis-parser type definitions, but would still be nice to hear if there's an interest for this so it's worth for me to keep working on it =)

leibale:
@evert sounds good to me :) Thanks for contributing!
Util it's ready, please convert this PR into a draft PR.

package.json
```diff
@@ -23,6 +23,7 @@
   },
   "devDependencies": {
     "@tsconfig/node12": "^1.0.9",
+    "@types/redis-parser": "^3.0.0",
```
evert, line 26: Note: I took the liberty to take the types from DefinitelyTyped, and remove the local ones. I figured maybe the local types were added because the DefinitelyTyped package didn't exist yet or it was inaccurate

But if you want me to revert this change, let me know! It will work either way

packages/client/lib/client/commands-queue.ts
```diff
@@ -1,8 +1,12 @@
-import LinkedList from 'yallist';
-import RedisParser from 'redis-parser';
+import * as LinkedList from 'yallist';
 import { AbortError } from '../errors';
 import { RedisCommandArguments, RedisCommandRawReply } from '../commands';
 
+// We need to use 'require', because it's not possible with Typescript to import
+// classes that are exported as 'module.exports = class`, without esModuleInterop
+// set to true.
+const RedisParser = require('redis-parser');
```
evert, line 8: I wish there was a way to avoid `require` but I had no luck with this =(

evert:
I think this is ready for review! Thanks for considering this, we're very excited for v4 with promises and typescript everywhere =)

package.json
leibale: Using the DefinitelyTyped version is a good idea, I haven't noticed it exists... Just please move it into ./packages/client/package.json (and not in the root package.json)

packages/client/lib/client/commands-queue.ts
leibale: Yea, I have found a solution too

tsconfig.base.json
```diff
@@ -3,11 +3,11 @@
   "compilerOptions": {
     "declaration": true,
     "allowJs": true,
-    "useDefineForClassFields": true
+    "useDefineForClassFields": true,
+    "esModuleInterop": false
   },
   "files": [
     "./packages/client/lib/ts-declarations/cluster-key-slot.d.ts",
```
leibale, line 10: Please remove the trailing comma

leibale (changes requested):

evert:
I addressed both issues!

leibale (approved):

PR #1727, opened 2021-11-21
Allow passing CURSOR to async scan iterators

ekvedaras:
### Description

At the moment iterators always start from `0`. I'm rebuilding my [redis-gui](https://github.com/ekvedaras/redis-gui) app using the new version and I have a use case where I want to start scanning from the specified cursor.
The app has a keys list on the left which by default loads only 100 keys matching the given pattern. If 100 keys were found during the scan but the cursor is still not `0`, that means there are more to scan. In such a case, it shows a _load more_ button, which should scan for another 100 keys starting from where it left the last time. I would really like to use iterators but I need to be able to pass the starting cursor.

This PR adds a new `ScanCursor` interface which is used to intersect with the `ScanOptions` interface on those iterator functions. If the optional `options.CURSOR` is not provided it defaults to `0` as before so this is all backwards compatible.

---

### Checklist

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -187,7 +187,8 @@ You can override the default options by providing a configuration object:
 client.scanIterator({
   TYPE: 'string', // `SCAN` only
   MATCH: 'patter*',
-  COUNT: 100
+  COUNT: 100,
+  CURSOR: 123,
 });
 ```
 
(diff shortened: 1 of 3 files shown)


ekvedaras:
I'm not able to properly run full tests locally (also it's incredibly slow). I see GitHub runs them quite fast, so I'd like it to do the job 😀

ekvedaras:
OK, I see I have some tests to fix. I will check tomorrow when I find some time 🙂

ekvedaras:
OK, just realized that this won't allow me to use iterators the way I want since I can't get the cursor back after iteration 😕

PR #1730, opened 2021-11-25
v4.0.0

leibale:


aardvarkk:
Just wanted to say thanks for this! The v4 API with TypeScript support feels fantastic -- what an improvement!

PR #1769, opened 2021-12-07
byteLength fix

kidGodzilla:
byteLength = string.length if string, Buffer.byteLength(arg) otherwise

### Description

Ran into a bug when an argument other than a string was passed (number), cannot read length of number.

Realized the evaluation of the ternary was flipped.

byteLength would be the length of a string, or the measured byteLength, for all other types. 

Opposite doesn't make sense (unless I'm missing something)

---

### Checklist

### DIFF
--- a/packages/client/lib/commander.ts
+++ b/packages/client/lib/commander.ts
@@ -96,7 +96,7 @@ export function* encodeCommand(args: RedisCommandArguments): IterableIterator<st
         stringsLength = 0;
     for (const arg of args) {
         const isString = typeof arg === 'string',
-            byteLength = Buffer.byteLength(arg);
+            byteLength = Buffer.byteLength(isString ? arg : '');
         strings += `$${byteLength}${DELIMITER}`;
 
         if (isString) {
--- a/packages/client/lib/commander.ts
+++ b/packages/client/lib/commander.ts
@@ -96,7 +96,7 @@ export function* encodeCommand(args: RedisCommandArguments): IterableIterator<st
         stringsLength = 0;
     for (const arg of args) {
         const isString = typeof arg === 'string',
-            byteLength = isString ? arg.length : Buffer.byteLength(arg);
+            byteLength = Buffer.byteLength(arg);
         strings += `$${byteLength}${DELIMITER}`;
 
         if (isString) {
--- a/packages/client/lib/commander.ts
+++ b/packages/client/lib/commander.ts
@@ -96,7 +96,7 @@ export function* encodeCommand(args: RedisCommandArguments): IterableIterator<st
         stringsLength = 0;
     for (const arg of args) {
         const isString = typeof arg === 'string',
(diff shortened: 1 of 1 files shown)


leibale:
`string.length` is the number of characters in a string, while `Buffer.byteLength('string')` returns the size in bytes (for example: `'א'.length` is `1`, but `Buffer.byteSize('א')` returns `2`.  [Check out the spec file](https://github.com/redis/node-redis/blob/master/packages/client/lib/commander.spec.ts) for more examples). Redis needs the size in bytes of the argument, not the number of characters in it.

So let's read it again:
```javascript
isString ?
  Buffer.byteLength(arg) : // if the argument is string - calculate its size in bytes using `Buffer.byteSize()`
  arg.length // the argument should be a `Buffer` - use the `.length` property.
```

unless you are using `.sendCommand` directly, each command should convert the arguments for you, take [HSET](https://github.com/redis/node-redis/blob/master/packages/client/lib/commands/HSET.ts) as an example:
```javascript
// plain strings
await client.hSet('key', 'field', 'value');

// plain numbers
await client.hSet('key', 1, 1);

// objects with strings and numbers
await client.hSet('key', {
  field1: '1',
  field2: 2'
});

// maps with strings and numbers
const map = new Map();
map.set(1, '1');
map.set(2, 2);
await client.hSet('key', map);
```

Which command are you trying to use?

kidGodzilla:
Thanks for the prompt reply!

I just read a couple of the related issues, I see the issue you're working around.

Unless I'm mistaken, I don't see the point of ever reading arg.length in the case that it is not a string -- should probably read `Buffer.byteLength` (or `Buffer.byteSize`?) in every case?

I ran into this by sending:

`client.HSET("b:" + remoteAddress, "loginTime", curTime);` 

which encounters an error, type number is passed to encodeCommand:

```
ERROR uncaughtException: TypeError: Cannot read properties of undefined (reading 'toString')
    at encodeCommand (@node-redis/client/dist/lib/commander.js:73:30)
    at encodeCommand.next (<anonymous>)
    at RedisSocket.writeCommand (@node-redis/client/dist/lib/client/socket.js:56:20)
    ...
```

vs. `client.HSET("b:" + remoteAddress, "loginTime", curTime.toString()); `

which works.

Basically, a number is getting through unless I explicitly convert to string before calling `.HSET`.

This may or may not be fixed by your change 5 days ago (which has not yet been published to NPM).

--

Unless I misunderstand something, arg should always be a string, and if it's not, we need to protect `Buffer.byteLength`, which throws an error if you pass `undefined` or a number (I'm guessing all non-strings).

If I'm just not understanding the issue, let me know.

leibale:
"This may or may not be fixed by your change 5 days ago (which has not yet been published to NPM)." - as you have said, I already pushed a fix to HSET that will handle numbers. I'll release it to npm soon.

https://nodejs.org/api/buffer.html#buflength

PR #1792, opened 2021-12-20
Contributing improvements

dhensby:
### Description

As someone trying to get the test suite running on macOS, the steps in the contributing guide and the test-utils and suite itself could be improved to make contribution easier.

This mainly aims to improve the ease of contributing to the repo on macOS. ~~Whilst Docker on mac does support host networking~~ [Docker on mac doesn't support host networks](https://docs.docker.com/network/host/), docker itself is run in a virtualised environment which means the host network is actually to that VM and not to the local host machine. This means that the ports do not actually get exposed as expected.

This PR does:

1. Documents the need for npm v7+ in the contributing guidelines
2. Adds titles to the `before` hooks that bring up docker containers (now if these fail during the test run it's easier to see what failed)
3. Improve the docker provisioning in the test suite so port forwarding works on macOS. 

Outstanding items:

- [ ] The docker cluster tests do not work locally on a mac still (investigating)

---

### Checklist

- [x] Does `npm test` pass with this change (including linting)? Not locally
- [x] Is the new or changed code fully tested? N/A
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? N/A


dhensby:
The clusters aren't working locally on a mac for me. Presumably because each instance of the redis docker can't communicate with the others when they are using traditional port-forwarding and the host network option doesn't work on mac either.

I'm no Docker expert and I've spent a bit of time looking into this and I'm not sure there's a good way to get it working as currently built. I'd have thought using `docker-compose` would have been a better solution (though admittedly more rigid) but at least it would work cross-platform rather than linux only as is the case.

leibale:
@dhensby It should work now, wanna test it?

dhensby:
Hi @leibale - thanks so much for looking into this. I'm afraid that the `testWithClient` tests are still failing, presumably because the docker containers still can't communicate with each other :(

dhensby:
I've tried adding macos (and windows too) to the build matrix as that might make things easier to see if it's working/fixed

Issue #1816, opened 2022-01-10
Export RediSearch module CreateSchema interface


dubblen:
Hello!

I am just migrating from ioredis to node-redis. Mainly to use the advantage of redis modules (RedisJson, RediSearch). I was using my own extension of ioredis that adds support of those two modules to ioredis but I did not have the full coverage of all commands that those two modules provides. Then I noticed that the 4th version of node-redis is released and it has exactly this feature.

What my extension had and what I was not able to achive with node-redis is a typescript interface that describes a RediSearch create schema.

To be able to do for example this:
![redisearch](https://user-images.githubusercontent.com/21110744/148841626-19239ac5-2dd3-47f9-9a84-ee4f1d908be6.png)

I know that I am able to copy the interfaces from the source code of the module, but it is just a code duplication and if the schema will change in the future then I will have to rewrite my interface again.

There could be more exports in general. I also was used to return Pipeline in some of my methods. That is possible with raw ioredis but unfortunately not with node-redis. Also SearchOptions interface might be very useful if we have access to it...

It is just a small change in the code and it might make life a little easier but maybe it has some reason not to export it. If it has then please tell me what it is.

Thank you very much!

PR #1825, opened 2022-01-12
export RediSearchSchema type

leibale:


leibale:
#1816

@dubblen is that the only type/interface you need?

dubblen:
Hi! Thank you for your response. I expect that there will be more of the interfaces that will be handy to have access to. Until now I am missing these interfaces:

- SearchOptions
- RedisClientMultiCommandType aka Pipeline
- CreateOptions
- CreateSchema
- SearchReply

Thank you very much.

mrbaguvix (approved):

PR #1832, opened 2022-01-13
Add streams XREADGROUP and XACK example.

### LINKED ISSUE #1831: Add example for XREADGROUP
Add example script for consumer groups using the blocking variant of the `XREADGROUP` command.

simonprickett:
### Description

This adds a stream consumer group example that uses `XREADGROUP` and `XACK`.

Closes #1831 

---

### Checklist

- [ y] Does `npm test` pass with this change (including linting)?
- [ y] Is the new or changed code fully tested?
- [ y] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?


examples/stream-producer.js
```diff
@@ -6,7 +6,6 @@ async function streamProducer() {
   const client = createClient();
 
   await client.connect();
-  await client.del('mystream');
```
leibale, line 9: ?

examples/stream-consumer-group.js
```diff
@@ -0,0 +1,104 @@
...
+
+  await client.connect();
+
+  // Create the consumer group (and stream) if needed...
+  try {
+    await client.xGroupCreate('mystream', 'myconsumergroup', '0', {
+      MKSTREAM: true
+    });
+    console.log('Created consumer group.');
+  } catch (e) {
```
leibale, line 41: catch error

examples/stream-producer.js
simonprickett: @leibale I removed the `del` because this makes this example work better with the consumer group one... if this deletes the stream it will delete any associated consumer groups too and then if they're running the consumer group example and it runs out of entries to process then they start the producer again it will blow away the stream rather than just continue to add new entries to it.

leibale (approved):

PR #1899, opened 2022-02-07
A new RESP2 parser

leibale:


packages/client/lib/client/RESP2/composers/buffer.spec.ts
```diff
@@ -0,0 +1,14 @@
+import { strict as assert } from 'assert';
+import BufferComposer from './buffer';
+
+describe('Buffer Composer', () => {
+    const composer = new BufferComposer();
+
+    it('should compose two buffers', () => {
+        composer.write(Buffer.from([0]));
+        assert.deepEqual(
```
chayim, line 9: Question: Are deepequal comparisons expensive in node? They are in Go.

packages/client/lib/client/RESP2/decoder.ts
```diff
@@ -0,0 +1,229 @@
+import { ErrorReply } from '../../errors';
+import { Composer } from './composers/interface';
+import BufferComposer from './composers/buffer';
+import StringComposer from './composers/string';
+
+enum Types {
+    SIMPLE_STRING = 43, // +
```
chayim, line 7: Good commenting on the actual ascii values.
IMHO a link to the RESP2 spec for the reader would be nice.

packages/client/lib/client/RESP2/decoder.ts
```diff
@@ -0,0 +1,229 @@
...
+
+        const toWrite = chunk.slice(this.#cursor);
+        composer.write(toWrite);
+        this.#cursor = chunk.length;
+    }
+
+    #findCRLF(chunk: Buffer): number {
+        for (let i = this.#cursor; i < chunk.length; i++) {
+            if (chunk[i] === CR) {
+                return i;
```
chayim, line 113: Might be a style difference. In python I'm more likely to do something like:
```
f = -1
for thing ... <my loop>:
    if things match:
        f = thing
        break
return f
```

I find it makes it easier to read functions when you're not looking for multiple returns

packages/client/lib/client/RESP2/decoder.ts
```diff
@@ -0,0 +1,229 @@
...
+        this.#cursor = chunk.length;
+    }
+
+    #arraysInProcess: Array<ArrayInProcess> = [];
+
+    #initializeArray = false;
+
+    #arrayItemType?: Types;
+
+    #parseArray(chunk: Buffer, arraysToKeep = 0): ArrayReply | undefined {
```
chayim, line 169: Yes, this feels ugly - you're right.
It also looks right.

packages/client/lib/client/RESP2/decoder.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    toWrite: Buffer;
+    returnStringsAsBuffers: boolean;
+    replies: Replies;
+}
+
+function generateTests({
+    toWrite,
+    returnStringsAsBuffers,
+    replies
+}: TestsOptions): void {
```
mrbaguvix, line 45: No commenting

packages/client/lib/client/socket.ts
```diff
@@ -113,23 +115,22 @@ export default class RedisSocket extends EventEmitter {
...
+            this.#socket = undefined;
 
-                throw err;
+            if (err instanceof AuthError) {
+                this.#isOpen = false;
             }
 
-            if (!this.#isOpen) return;
+            throw err;
         }
```
mrbaguvix, line 130: should we do `console.error(err)`

mrbaguvix (changes requested):

PR #1901, opened 2022-02-08
fix #1891 - fix ft.create type

leibale:

PR #1980, opened 2022-02-09
[SUGGESTION] Cleaned up search index & json module

Didas-git:
### Description

Since i commented about this I decided to take a look and cleanup the file a bit.
This should be taken with a grain of salt bcz its purelly based on my opinion.

Also implemented all the changes in #1901 and the suggestion i commented about, `keyof any` to `PropertyKey`

---

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "redis",
-  "version": "4.6.5",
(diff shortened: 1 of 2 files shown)


Didas-git:
I also thought about changing some of the `if` statements to `&&`

PR #1984, opened 2022-02-14
fix #1865 - enhance "all in one" types

### LINKED ISSUE #1865: RedisClientType hard to use with createClient
The RedisClientType has two generic parameters that make it hard to use to signify the return type of `createClient`. In particular, the first argument `modules` in the declaration of `createClient` is a local variable that is not exported and thus cannot be easily reused in one's own code.

Maybe it's already enough to provide sufficiently general defaults for the generic arguments. If I'm not mistaken these parameters control more the configuration of the client and don't really influence the typing of the "normal" methods; so they are not that important for an user of a factory method.

Below is a similar observation from #1673.

---

The `RedisClientType` exported since 4.0.1 still does not work for me. Could not figure out how to make the exported `RedisClientType`, `RedisClientOptions` and the type returned by `createClient` compatible. Here is an [example that illustrates the issue](https://codesandbox.io/s/node-redis-typescript-fhzck).

The following does not compile unless someone knows how to type `RedisClientOptions` correctly (refer to the code sandbox link above).

```
import { createClient, RedisClientOptions, RedisClientType } from "redis";

const factory = (options: RedisClientOptions<any, any>): RedisClientType => {
  return createClient(options);
};

const client: RedisClientType = factory({...});
```

Instead, I am doing this for now:

```
import { createClient} from "redis";
type RedisClientType = ReturnType<typeof createClient>;
type RedisClientOptions = Parameters<typeof createClient>[0];

const factory = (options: RedisClientOptions): RedisClientType => {
  return createClient(options);
};

const client: RedisClientType = factory({...});
```

_Originally posted by @tbinna in https://github.com/redis/node-redis/issues/1673#issuecomment-993335708_

leibale:
#1865

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,5 +1,13 @@
-import { createClient as _createClient, createCluster as _createCluster, RedisClientOptions, RedisClientType, RedisClusterOptions, RedisClusterType } from '@node-redis/client';
-import { RedisScripts } from '@node-redis/client/dist/lib/commands';
+import {
+    RedisModules,
+    RedisScripts,
+    createClient as _createClient,
+    RedisClientOptions,
+    RedisClientType as _RedisClientType,
+    createCluster as _createCluster,
+    RedisClusterOptions,
+    RedisClusterType as _RedisClusterType
+} from '@node-redis/client';
 import RedisBloomModules from '@node-redis/bloom';
 import RedisGraph from '@node-redis/graph';
 import RedisJSON from '@node-redis/json';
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -13,28 +21,46 @@ export * from '@node-redis/json';
...
     ft: RediSearch,
     ts: RedisTimeSeries
 };
 
-export function createClient<S extends RedisScripts>(
-    options?: Omit<RedisClientOptions<never, S>, 'modules'>
-): RedisClientType<typeof modules, S> {
+export type RedisDefaultModules = typeof modules;
+
+export type RedisClientType<
```
tobiasdiez, line 34: I don't have experience with this, but is this not running into conflicts with `export * from '@node-redis/client';` where `RedisClientType` is also exported?

leibale: No, it'll override it

Issue #1989, opened 2022-02-17
Module 'redis' has no exported member 'Multi' on node-redis version 4.0.3


gutakk:
Hi everyone, I'm trying to migrate node-redis v3 to v4 and this issue occur on `Multi` import as shown below
`Module '"redis"' has no exported member 'Multi'`

Not sure how can I find the type `Multi` in node-redis v4.0.3.

Thanks in advance

**Source code**
```ts
import { Multi } from "redis";

export interface IRedisPromiseMultiClient {
  exec(): Promise<any[]>;
  get(key: string, field: string): RedisPromiseMultiClient;
}

export class RedisPromiseMultiClient implements IRedisPromiseMultiClient {
  protected formatter: ((result: any) => any)[] = [];

  constructor(protected multi: Multi) {}

  public exec(): Promise<any[]> {
    return new Promise((resolve, reject) => {
      this.multi.exec((err, reply: any[]) => {
        if (err) {
          reject(err);
        }
        resolve(reply.map((value, i) => this.formatter[i](value)));
      });
    });
  }

  public get(key: string, field: string): RedisPromiseMultiClient {
    this.multi = this.multi.hget(key, field);
    this.formatter.push((r) => r);
    return this;
  }
}
```

**Updated: Solution by myself**
```ts
// added following line to @node-redis/client/dist/index.d.ts
export { RedisClientMultiCommandType } from './lib/client/multi-command';
```
and then added following lines of code to source code
```ts
import { RedisClientMultiCommandType, RedisModules, RedisScripts } from "redis";

type Multi = RedisClientMultiCommandType<RedisModules, RedisScripts>
```

Is it looking good to make the PR for this?

Didas-git:
About the initial issue:
There is a better and easier way to get arround this current issue this being the following.
```ts
import { createClient } from "redis",

const redisClient = createClient()
type RedisClientMultiCommandType = ReturnType<typeof redisClient.multi>
```
About your pr:
Your pr needs to be basicly remade before being merged because it is exporting it from the wrong places.
`RedisClientMultiCommandType` is defined at client>dist>lib>client>multi-command.ts
And the export should be the the main index.ts

gutakk:
@Didas-git Thanks. It is working like a charm

PR #1992, opened 2022-02-18
export RedisClientMultiCommandType

gutakk:
Issue description is in #1989


mrbaguvix (commented):
Should we try this if it works the problem https://github.com/redis/node-redis/issues/1989#issuecomment-1049314741

gutakk:
> Should we try this if it works the problem [#1989 (comment)](https://github.com/redis/node-redis/issues/1989#issuecomment-1049314741)

This workaround working like a charm. Thanks

PR #2008, opened 2022-02-25
Fixes the time command response.

### LINKED ISSUE #2001: TIME() returning a date in 1970 rather than current timestamp
According to the Redis docs, TIME() should show the current server time, as a timestamp since the Unix epoch, and the number of microseconds that have elapsed: https://redis.io/commands/TIME

However, doing client.time() returns a date in 1970, not the current time.

Sample code:

```
const client = redis.createClient();
client.connect();

 client.on('connect', function() {
    console.log('Connected!');
	client.TIME().then(function(reply){
		console.log(reply);
	});
});
```

Output: 1970-01-20T01:05:44.873Z { microseconds: 880958 }

Linux `date +%s` returns 1645544389, which is the current date/time

**Environment:**
 - **Node.js Version**: 14.19.0
 - **Redis Server Version**: 6.2.6
 - **Node Redis Version**: 4.0.4
 - **Platform**: Ubuntu 20.04.3 LTS

simonprickett:
This should fix #2001 - issue with the `TIME` command returning a date in 1970.  Also adds an example to show it working:

```bash
node-redis $ cd examples
examples $ node get-server-time.js
2022-02-25T12:55:41.000Z { microseconds: 130723 }
```


examples/get-server-time.js
```diff
@@ -0,0 +1,16 @@
...
+
+async function getServerTime() {
+  const client = createClient();
+  await client.connect();
+
+  const serverTime = await client.time();
+  // 2022-02-25T12:57:40.000Z { microseconds: 351346 }
+  console.log(serverTime);
+
+  await client.quit();
```
mrbaguvix, line 13: should we wrap in `try/catch` block?

examples/README.md
```diff
@@ -10,6 +10,7 @@ This folder contains example scripts showing how to use Node Redis in different
 | `connect-as-acl-user.js`    | Connect to Redis 6 using an ACL user                                                                                                                |
 | `count-min-sketch.js`       | Estimate the frequency of a given event using the [RedisBloom](https://redisbloom.io) Count-Min Sketch                                              |
 | `cuckoo-filter.js`          | Space efficient set membership checks with a [Cuckoo Filter](https://en.wikipedia.org/wiki/Cuckoo_filter) using [RedisBloom](https://redisbloom.io) |
+| `get-server-time.js`        | Get the time from the Redis server                                                                                                                  |
```
mrbaguvix, line 13: Is the time a current UNIX time like `now()`? so we can have clearer description.

mrbaguvix (changes requested):

examples/get-server-time.js
simonprickett: Doesn't add value to the point of the example, so I'd say no.

examples/README.md
simonprickett: No need, it's shown as a comment in the code and documented at https://redis.io/commands/time

mrbaguvix (approved):

Issue #2006, opened 2022-02-24
Why is #private used?


euberdeveloper:
I see that `#private` is used, even if the code is written in Typescript (ts already has a `private` keyword). It is true that typescript privates have no reflections in the resulting javascript code, but at least they support es5. What should I do if I wanted to use es5?

**Environment:**
 - **Node.js Version**: v16.3.0
 - **Node Redis Version**:  4.0.4
 - **Platform**: Ubuntu 20

leibale:
["Private class features" available since node 12](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields#browser_compatibility), which is [the oldest "active" node version](https://nodejs.org/en/about/releases/). What are you trying to do?

euberdeveloper:
The problem is not when I execute the ES5 code, but when I transpile it with Typescript targeting ES5

```Private identifiers are only available when targeting ECMAScript 2015 and higher```

euberdeveloper:
The only way I've found to avoid this problem is setting `skipLibCheck` as false. Otherwise I had to change the target to ES6

Didas-git:
I dont get what you mean, the issue should be on your code since the version on npm is already compiled and it shouldnt error.
Maybe with more information we are able to find something (?)

euberdeveloper:
I think that just by installing/using the library and trying to transpile with typescript (target es5, non-skip-lib), the issue would appear.

Didas-git:
> I think that just by installing/using the library and trying to transpile with typescript (target es5, non-skip-lib), the issue would appear (during transpilation, not runtime).

I will take a look then, but i dont see why you would skip library checkings

euberdeveloper:
But I thin that the only way is to use skipLibCheck or Babel with Typescript. The reason why your code transpiles is because you have es2019 as target.

So the only thing I would do on your side is writing in the migration guide that for typescript es5 as target is not supported because of the # fields and that the only way to keep it work is to add the skipLibCheck

euberdeveloper:
> > I think that just by installing/using the library and trying to transpile with typescript (target es5, non-skip-lib), the issue would appear (during transpilation, not runtime).
> 
> I will take a look then, but i dont see why you would skip library checkings

It is because in that way there is not a typescript check in the dependencies (i.e. node_modules)

Didas-git:
[...]

PR #2009, opened 2022-02-25
Mihration guide 

Didas-git:
### Description

Discussed in #2006 targeting a version of ecmascript older than 2015 will cause you errors.
Not a breaking change but as the user suggested im adding it to the migration guide.

### Checklist

- [x] Is the new or changed code fully tested? (No changes were made)

### DIFF
--- a/docs/v3-to-v4.md
+++ b/docs/v3-to-v4.md
@@ -4,6 +4,7 @@ Version 4 of Node Redis is a major refactor. While we have tried to maintain bac
 
 ## All of the Breaking Changes
 
+If using typescript set the target to ES2015 (ES6) or higher, ES5 and bellow are not supported due to the usage of private class functions.
 See the [Change Log](../CHANGELOG.md).
 
 ### Promises


docs/v3-to-v4.md
```diff
@@ -4,6 +4,7 @@ Version 4 of Node Redis is a major refactor. While we have tried to maintain bac
 
 ## All of the Breaking Changes
 
+If using typescript set the target to ES2015 (ES6) or higher, ES5 and bellow are not supported due to the usage of private class functions.
 See the [Change Log](../CHANGELOG.md).
```
mrbaguvix, line 8: Is there also a specific Typescript version (or range) we should expect this change?
```suggestion
If using Typescript set the target to ES2015 (ES6) or higher, ES5 and below are not supported due to the usage of `private` class functions.
See the [Change Log](../CHANGELOG.md).
```

Didas-git: It works on Typescript v4 and should work on v3. Not sure about other versions but i would always use the most recent version if possible.
And thnx for the correction

PR #2106, opened 2022-04-26
Update sorted set example to show ZRANGE with REV modifier

### LINKED ISSUE #2100: Amend the sorted set example script to show `REV` behaviour
Amend the sorted set example script to show `REV`.

Script is: https://github.com/redis/node-redis/blob/master/examples/sorted-set.js

simonprickett:
Updates the sorted set example to show use of the `REV` modifier to the `ZRANGE` command (Redis >= 6.2).  Closes #2100

### DIFF
--- a/examples/sorted-set.js
+++ b/examples/sorted-set.js
@@ -1,5 +1,5 @@
 // Add several values with their scores to a Sorted Set,
-// then retrieve them all using ZSCAN.
+// then retrieve them all using ZSCAN, ZRANGE and ZREVRANGE.
 
 import { createClient } from 'redis';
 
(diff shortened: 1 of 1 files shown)


examples/sorted-set.js
```diff
@@ -25,9 +27,23 @@ async function addToSortedSet() {
   // Get all of the values/scores from the sorted set using
   // the scan approach:
   // https://redis.io/commands/zscan
+  console.log('Using ZCSCAN:');
   for await (const memberWithScore of client.zScanIterator('mysortedset')) {
     console.log(memberWithScore);
   }
+
+  // Get all of the values/scores from the sorted set using
```
leibale, line 35: I think that we should add a note that this is not a good idea..

simonprickett: Probably yes, can we do this with `ZSCAN` with `WITHSCORES` and `REV` modifiers?

leibale: `ZSCAN` gives the score (so there is no need for the `WITHSCORES` modifier), but does not support the `REV` modifier

PR #2132, opened 2022-05-11
nit: spelling fix memoery to memory

sparecycles:
### Description

> Tiny spelling fix....  Simultaneously unimportant, and it might be a (trivially) breaking change for some clients, but doesn't seem to have been even raised as an issue yet so here we are.  It's on your radar now. 🤷 

---

### Checklist

- [ ] Does `npm test` pass with this change (including linting)? - no, but `master` doesn't either on my machine. (a typescript complaint about implicit any type) ⚠️ 
- [ ] Is the new or changed code fully tested? - no
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? - no, but this may be a breaking change for some clients.


leibale (approved):
Nice catch! :)

PR #2139, opened 2022-05-19
Fix issue with buffers in objects using hSet

Tyris:
### Description

When using hSet with an object, any buffer values inside the object are converted to strings instead of left as buffers.

This fix specifically handles the special case of buffers, whilst casting everything else strings (to continue "gracefully" handling the case where the value not a valid type).

This is technically a breaking change, but it's unlikely that anyone is relying on the existing broken behaviour - which is that buffers come back out (using hGetAll) mangled. Eg:

```
> b
<Buffer 01 02 81 82 f1>
> r.hSet('temp', {b})
> r.hGetAll(redis.commandOptions({returnBuffers: true}), 'temp').then(console.log)
> [Object: null prototype] {
  b: <Buffer 01 02 ef bf bd ef bf bd ef bf bd>
}
```

---

### Checklist

### DIFF
--- a/packages/client/lib/commands/HSET.spec.ts
+++ b/packages/client/lib/commands/HSET.spec.ts
@@ -41,11 +41,20 @@ describe('HSET', () => {
             );
         });
 
-        it('Object', () => {
-            assert.deepEqual(
-                transformArguments('key', { field: 'value' }),
-                ['HSET', 'key', 'field', 'value']
-            );
+        describe('Object', () => {
+            it('string', () => {
+                assert.deepEqual(
+                    transformArguments('key', { field: 'value' }),
+                    ['HSET', 'key', 'field', 'value']
+                );
+            });
+            
+            it('Buffer', () => {
+                assert.deepEqual(
+                    transformArguments('key', { field: Buffer.from('value') }),
+                    ['HSET', 'key', 'field', Buffer.from('value')]
+                );
+            });
         });
     });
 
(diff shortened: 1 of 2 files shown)


leibale:
@Tyris good catch! can you please add a test for it?

Tyris:
@leibale Will give it a try on the weekend (not familiar with the codebase/build-process, but should be straight-forward).

For the sake of completeness, the easy workaround (which I've implemented on the impacted project) is to simply unpack the object into an array (which handles buffer insertion correctly) using `r.hSet('temp', Object.entries(someObj));`

leibale:
@Tyris review my changes? :pray:

Tyris:
@leibale test looks good.

A conversion function is definitely cleaner, but this introduces an additional edge cases that was caught by the previous implementation, which is if the object value is not a buffer, string, or number.

Javascript only allows strings for object keys, so the key is fine (in fact, the original .toString() was superfluous), but the value can be anything (an object, array, boolean, etc). The previous logic would force any other values to be a string, where is now the value will be passed through as is. I'm not sure how the underlying system will handle this.

This is actually the same as if the following code was used: `r.hSet("key", "field", someObject);`, where `someObject` would be left as is (and similar for arrays/maps).

Really not sure what the intended behaviour should be here. Should it always convert to string for anything that isn't a buffer? Should it throw? Or will the underlying system handle this in some way?

Simple solve would be to explicitly convert anything that isn't already a buffer or string into a string in the new conversion function (assuming these are our only supported data types).

leibale:
The edge cases will be caught in the [RESP encoder](https://github.com/redis/node-redis/blob/master/packages/client/lib/client/RESP2/encoder.ts#L21) (it'll raise an error, which is better than `'[object Object]'` in my opinion)

Tyris:
Even better. I'll mark as ready for review.

OldBuddy:
hey, any reason same is not done for `hSetNX`?

I have different behavior running these 2:
* `client.hSet('key', 'prop1', 1)`
* `client.hSetNX('key', 'prop1', 1)`

PR #2235, opened 2022-08-16
disconnect in basic example

reviewher:
### Description

The basic example just hangs.  It was missing the disconnect.

---

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ await client.connect();
 
 await client.set('key', 'value');
 const value = await client.get('key');
+await client.disconnect();
 ```
 
 The above code connects to localhost on port 6379. To connect to a different host or port, use a connection string in the format `redis[s]://[[username][:password]@][host][:port][/db-number]`:


leibale:
I'm pretty sure that most of the node apps are REST APIs, and in that scenario you don't want to `.disconnect()`/`.quit()`

@simonprickett @guyroyse WDUT?

reviewher:
It's a "basic example" and listed as the first example in the README.  There are no instructions for how to run it.

"Obviously" you start a local server and copy-paste the example to a MJS file and run it with node.  Try it.  The process will just hang.

The documentation can include a note about when disconnect/quit are appropriate.

leibale:
Maybe we should leave the example as is and add a note with a link to [this section](https://github.com/redis/node-redis#disconnecting)

reviewher:
If it is the case that "most of the node apps are REST APIs", then the "basic example" should be a simple REST server using raw Node HTTP server or ExpressJS.

The idea of a "basic example" IMHO is something you can run.  A person who is new to `node-redis` will see the snippet with an actual import statement etc. and just test it out, wondering why the process is stuck.

simonprickett:
I like the idea of having it in there?

taylorsweetman (approved):

PR #2253, opened 2022-09-07
Add WITHSCORES option to ZRANGE command

stories2:
### Description

Screen Shot 2022-09-07 at 13 56 04

> The `ZRANGE` command has `WITHSCORES` option. 
> 
> But, current node-redis package doesn't support `WITHSCORES` option
>
> So, i added.

---

### Checklist

### DIFF
--- a/packages/client/lib/commands/ZRANGE.spec.ts
+++ b/packages/client/lib/commands/ZRANGE.spec.ts
@@ -63,6 +63,30 @@ describe('ZRANGE', () => {
                 ['ZRANGE', 'src', '0', '1', 'BYSCORE', 'REV', 'LIMIT', '0', '1']
             );
         });
+
+        it('with WITHSCORES', () => {
+            assert.deepEqual(
+                transformArguments('src', 0, 1, {
+                    WITHSCORES: true
+                }),
+                ['ZRANGE', 'src', '0', '1', 'WITHSCORES']
+            );
+        });
+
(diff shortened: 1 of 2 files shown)


leibale:
@stories2 use ZRANGE_WITHSCORES instead. The reason it's implemented as a different commands is that the reply type is different.

stories2:
> @stories2 use ZRANGE_WITHSCORES instead. The reason it's implemented as a different commands is that the reply type is different.

My bad. Thanks for your help.

leibale:
@stories2 thanks for trying to fix it yourself :)

SnowdenWintermute:
Thank you! There are a lot of wrong examples out there showing that you can pass "WITHSCORES" to client.zRange as a plain string option and it was causing me a lot of confusion.

leibale:
@SnowdenWintermute can you please share some links?

SnowdenWintermute:
I guess I was exaggerating when I said "a lot", but here are the two I found. They use zrevrange which is deprecated now, and since it was replaced by zrange I assumed zrange could do what zrevrange could do and get the scores via passing an option. I eventually found this thread by searching something like "node redis zrange withscores option not included" or "doesn't exist".

https://stackoverflow.com/questions/24157632/node-redis-get-zrange-withscores

https://gist.github.com/danielrvt/91765654445006bb28e2

leibale:
@SnowdenWintermute thanks for the links, I'll check them later and see if I can comment there to help others.. :)

Issue #2134, opened 2022-05-16
Memory leak on repeated connection failures


evanbattaglia:
I believe there may be a memory leak when we repeatedly cannot connect to redis. I first noticed this because our app was misconfigured and kept trying and failing to connect to redis; memory usage would grow and grow until after a few hours Kubernetes would OOM kill it.

I am not experienced in tracking node memory leaks but I _think_ I have uncovered evidence by running this script locally:

```
const {createClient} = require("redis");
const client = createClient({
  url: "redis://localhost:11111", // anywhere redis isn't running, or a bad hostname
  // this line optional but creates objects faster:
  socket: {reconnectStrategy: retries => 0}
});
client.on('error', () => {});
client.connect();
```

I ran that with `node --inspect`; in DevTools I went to the Memory tab and took two consecutive Heap Snapshots with a few seconds in between. Selecting the second snapshot I changed "All Objects" at the top do "Objects allocated between Snapshot 2 and Snapshot 1". It shows a large number of Promises and other objects created (in some cases I've seen a small number of `TimersList`s also rank high). Additional snapshots just get bigger and bigger in terms of MB.

When I run with `--inspect` I can also see the node process's RSS memory in `ps aux` get bigger and bigger. For some reason, without the flag, it grows but seems to be growing slower (I have to keep it on for a few minutes to notice growth beyond what is GC'd).

![Screenshot from 2022-05-16 13-00-28](https://user-images.githubusercontent.com/669236/168569744-e7a26289-2f23-4003-a80a-6ee8f79e8123.png)

This doesn't happen when redis connects successfully. I also don't see it version 3 of the package; here is what I used there:

```
const redis = require("redis");
const {createClient} = redis;
const client = createClient({
  url: "redis://localhost:11111",
  retry_strategy: options => {
    //console.log("error!"); // this logs here
    return 0;
  }
});
client.on('error', console.log); // doesn't seem to log anything but whatever
client.set("key","value",redis.print);
```

I tried debugging a bit but didn't get anywhere. The Promises from Devtools looked to be related to `_RedisClient_isolationPool`, but I commented out most code related to that and it seemed to be still happening. I've also tried running with --expose-gc and manuallying GC every second; that didn't change much.

Since this only affects things when redis repeatedly can't connect, I realize this is probably not a priority bug, so feel free to ignore it. I did want to let everyone know what I saw and found though in case anyone comes across it.

**Environment:**
 - **Node.js Version**: v14.16.1
[...]

PR #2295, opened 2022-10-16
fix(client): Avoids infinite promise-chaining when socket's creation fails

JonasFaure:
### Description

This change prevents a memory leak when a Redis client tries to connect to an unavailable host. This is achieve by switching from a recursive to an iterative approach.

When a Redis instance fails, our clients slowly build up memory until they get killed. 

https://github.com/redis/node-redis/issues/2134

First, I had to slightly change the tests' implementation to work around timeout issues when running the tests, ie : specifying a timeout using the _connectTimeout_ argument & giving up mocking time by specifying a very small retry delay. This commit should preferably be reverted if someone who manages to run tests locally can validate my implementation with the previous implementation for those tests :shrug: 

Second, I swapped the recursive call to `connect` for a do{}while loop and leverage the `#this.isReady` property as the exit condition. 

---

### Checklist

### DIFF
--- a/packages/client/lib/client/socket.spec.ts
+++ b/packages/client/lib/client/socket.spec.ts
@@ -19,7 +19,7 @@ describe('Socket', () => {
 
     describe('reconnectStrategy', () => {
         it('custom strategy', async () => {
-            const numberOfRetries = 10
+            const numberOfRetries = 10;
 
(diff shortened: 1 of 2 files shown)


leibale:
@JonasFaure good job! sorry for the long delay, I was on vacation... I'll release a version with this fix next week :)

PR #2352, opened 2022-12-28
WIP KNN example

### LINKED ISSUE #2351: Struggling to invoke KNN vector similarity search
It would be great to have a small working example for [vector similarity search](https://redis.io/docs/stack/search/reference/vectors/#storing-vectors-in-hashes) similar to the [search-hashes](https://github.com/redis/node-redis/blob/master/examples/search-hashes.js) example.

I'm struggling to figure out how to encode floating point arrays in the correct way to store in Redis, and also how to represent the vector to search by in the query.

Here's some wreckage showing various things I attempted:

```js
import fs from 'fs';
import redis from 'redis';
// This is an array of 512 floating point numbers
import { embedding } from './sampleEmbedding.js';
const { createClient, SchemaFieldTypes, commandOptions, print } = redis;

// See
// https://redis.js.org/
client.on('error', (err) => console.log('Redis Client Error', err));
await client.connect();

const idx = 'idx:vector';

// See
// https://redis.io/docs/stack/search/reference/vectors/
// From https://github.com/redis/node-redis/blob/master/examples/search-hashes.js#L10
try {
  await client.sendCommand(['FT.DROPINDEX', idx]);
  await client.sendCommand([
    'FT.CREATE',
    idx,
    'SCHEMA',
    'vector',
    'VECTOR',
    'HNSW',
    '6',
    'TYPE',
    'FLOAT32',
    'DIM',
    '512',
    'DISTANCE_METRIC',
    'COSINE',
  ]);
} catch (e) {
  if (e.message === 'Index already exists') {
    console.log('Index exists already, skipped creation.');
  } else {
    // Something went wrong, perhaps RediSearch isn't installed...
    console.error(e);
    process.exit(1);
  }
}
const id = '84790q8758493d';
await client.hSet(
  id,
  'vector',
  Buffer.from(new Float32Array(embedding).buffer) // No idea if this is correct or not
);
const result = await client.hGetAll(id);

// It outputs a crazy looking string with strange characters. Expected I think?
console.log(result.vector);

// From
// https://stackoverflow.com/questions/40031688/javascript-arraybuffer-to-hex
function bufferToHex(buffer) {
  return [...new Uint8Array(buffer)]
    .map((b) => '\\x' + b.toString(16).padStart(2, '0'))
    .join('');
}

// Trying to construct the BLOB string
// that Redis expects in its commands (no idea if this is correct)
const blob = (array) => bufferToHex(new Float32Array(array).buffer);

console.log(
  await client.sendCommand([
    'FT.SEARCH',
    idx,
    '*=>[KNN 10 @vec $BLOB]',
    'PARAMS',
    '2',
    'BLOB',
    blob(embedding), // I feel like this is wrong - there must be a better way?
    'DIALECT',
    '2',
  ])
);

// Prints [ 0 ], disappointingly. Should contain a single document.
console.log(results);

await client.disconnect();
```

Any guidance would be greatly appreciated. Thanks in advance!

curran:
Closes #2351 (if / when we can get it working!)

### DIFF
--- a/examples/search-knn.js
+++ b/examples/search-knn.js
@@ -1,0 +1,85 @@
+// This example demonstrates how to use RediSearch to index and query data 
+// stored in Redis hashes using vector similarity search.
+
+import { createClient, SchemaFieldTypes } from 'redis';
+
+const client = createClient();
+
+await client.connect();
+
+// Create an index...
+try {
+  // Documentation: https://redis.io/docs/stack/search/reference/vectors/
+  await client.sendCommand([
+    'FT.CREATE',
+    'idx:vectors',
+    'SCHEMA',
+    'vector',
+    'VECTOR',
+    'HNSW',
+    '6',
+    'TYPE',
+    'FLOAT32',
+    'DIM',
+    '512',
+    'DISTANCE_METRIC',
+    'COSINE',
+  ]);
+} catch (e) {
+  if (e.message === 'Index already exists') {
+    console.log('Index exists already, skipped creation.');
+  } else {
+    // Something went wrong, perhaps RediSearch isn't installed...
+    console.error(e);
+    process.exit(1);
+  }
+}
+
+// Add some sample data...
+// https://redis.io/commands/hset/
+await Promise.all([
+  client.hSet('noderedis:vectors:1', { vector: [0.4, -0.325, 4.3] },
+  client.hSet('noderedis:vectors:2', { vector: [0.3, -0.3, 4.4] },
+  client.hSet('noderedis:vectors:3', { vector: [-0.4, 0.325, -4.3] },
+  client.hSet('noderedis:vectors:4', { vector: [38, 32.5, -8.1] },
+]);
+
+// Perform a K-Nearest Neighbors vector similarity search
+// Documentation: https://redis.io/docs/stack/search/reference/vectors/#pure-knn-queries
+const results = await client.ft.search(
+  'idx:vectors', 
+
+  // TODO how do we even get this syntax into client.ft.search
+  FT.SEARCH idx "*=>[KNN 10 @vec $BLOB]" PARAMS 2 BLOB "\x12\xa9\xf5\x6c" SORTBY __vec_score DIALECT 2
+);
+
+// results:
+// {
+//   total: 2,
+//   documents: [
+//     { 
+//       id: 'noderedis:animals:3',
+//       value: {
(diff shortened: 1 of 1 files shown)


examples/search-knn.js
```diff
@@ -0,0 +1,85 @@
...
+    // Something went wrong, perhaps RediSearch isn't installed...
+    console.error(e);
+    process.exit(1);
+  }
+}
+
+// Add some sample data...
+// https://redis.io/commands/hset/
+await Promise.all([
+  client.hSet('noderedis:vectors:1', { vector: [0.4, -0.325, 4.3] },
```
curran, line 41: Do we need to encode these vectors with `Buffer`? How should that be done exactly?

curran:
Got it to work! Open to suggestions for cleaning it up.

leibale:
@curran thanks for contributing! I'll update it to use `client.ft.create` and `client.ft.search` (instead of `client.sendCommand`) later today and then merge it :)

leibale:
@curran review my changes? :pray:

examples/search-knn.js
```diff
@@ -13,71 +13,79 @@ await client.connect();
...
-    "TYPE",
-    "FLOAT32",
-    "DIM",
-    "2",
-    "DISTANCE_METRIC",
-    "COSINE",
-  ]);
+  await client.ft.create('idx:knn-example', {
+    v: {
+      type: SchemaFieldTypes.VECTOR,
```
curran, line 18: Oh nice!!! This is exactly what I wanted to do but couldn't figure out the API for. Thank you.

examples/search-knn.js
```diff
@@ -13,71 +13,79 @@ await client.connect();
...
-  "2",
-  "BLOB",
-  tobytes([0.1, 0.1]),
-  "SORTBY",
-  "dist",
-  "DIALECT",
-  "2",
-]);
-
+const results = await client.ft.search('idx:knn-example', '*=>[KNN 4 @v $BLOB AS dist]', {
```
curran, line 52: AMAZING! I could not for the life of me figure out how to formulate this with the `ft.search` API. This is perfect.

curran (commented):
Amazing! Thank you for the clean up, this looks fantastic.

DvirDukhan (approved):
Looks good, thank you

curran:
Hooray!

Issue #2038, opened 2022-03-16
SendCommand hGetall return format changed from v3 to v4


linnerissa:
in v3, calling send_command with hGetall command would result in a return of format 
```
{
    "key1": "value1",
    "key2": "value2",
}
```
in v4, sendCommand with "hGetAll" returns `["key1", "value1", "key2", "value2"]`.  This seems like a breaking change between the versions that wasn't listed

Seems like running `client.HGETALL` however returns in object format. Is there a method to coerce the sendCommand return to be formatted correctly?

**Environment:**
 - **Node.js Version**:  16.13.2
 - **Redis Server Version**:  Redis 6.2.6
 - **Node Redis Version**:  4.0.4
 - **Platform**:  MacOS 12.1

jhavok86:
Ran into this issue as well. I switched to using the wrapped function but the results still return as an array. This seems to only happen when legacyMode = true. If I do not use legacyMode then results return as expected.

PR #2367, opened 2023-01-05
#2038 Resolve legacy mode hGetAll returning in the wrong format compared to v3 results

ChronicLynx:
### Description

Resolves Issue #2038

Noteworthy:

- In V3 empty replies came back as `null` where in V4 empty replies are delivered as an empty object ({}) or array([])
- I did not fix empty reply behavior pertaining to `legacyMode`, perhaps it warrants more investigation but is outside the scope of 2038

Left Side V4 vs Right Side V3

Screenshot 2023-01-05 at 11 50 02 AM

---

During a migration to version 4 our team encountered an issue regarding legacy mode return formatting.

`multi.hGetAll` was returning as a tuple instead of an object. _**Prior**_ to upgrading and enabling `legacyMode` it was returning the value as an object, and the v4 version also returns the reply as an object. Thanks to our joi data validation we were able to catch the mismatch and investigate.

I believe this is because the `legacyMode` `addCommand` for non-v4 currently calls
`this.#multi.addCommand(transformLegacyCommandArguments(args));` and omits the `transformReply` which is exported by the various commands and seemingly needed for `hGetAll` to match the v3 formatting.

I've added an optional boolean to `RedisCommands` needing this transform (`hGetAll` for now). As COMMANDS is lacking typing I cast it as any to avoid errors when accessing the object using the command name string.

With this fix in place the legacy mode `multi.hGetAll` is now returning an object as expected.

### Additional Work

#2038 points out that this issue may also spread to the `sendCommand` method which takes a very similar approach to legacy splitting as the `addCommand` method did.

I've also applied the transform in this case, which formats the reply as an object instead of a tuple, mimicking v3 behavior.

Screenshot 2023-01-05 at 11 39 59 AM

### Important:

By applying `transformReply` to _**everything**_ (as a test case) I've noticed odd behavior. `hGetAll` with `transformReply` applied matches the old v3 behavior. Something like `client.scan` returned a tuple originally but a `transformReply` causes it to incorrectly return as an object.

Since `hGetAll` is the only one known to have an issue I've limited the scope to the `hGetAll / HGETALL` command. I believe applying the transform to everything is dangerous, as it appears this reply type mismatch either only happens to `hGetAll` or a small subset of commands that have yet to be discovered.

### Testing Issues

I was able to run this through my own project test suite and the issue appears to be resolved. 

I've partially figured out how to run the official test suite locally but still having some trouble with some Cluster tests... perhaps I'm not setting it up correctly.

`npm run test -w ./packages/client -- --redis-version 7.0`

**Local Test Run (exact same results as master test run):**
1119 passing
2 pending
9 failing

---

### Checklist

### DIFF
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -8,6 +8,7 @@ import { defineScript } from '../lua-script';
 import { spy } from 'sinon';
 import { once } from 'events';
 import { ClientKillFilters } from '../commands/CLIENT_KILL';
+import { promisify } from 'util';
 
 export const SQUARE_SCRIPT = defineScript({
     SCRIPT: 'return ARGV[1] * ARGV[1];',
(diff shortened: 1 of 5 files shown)


ChronicLynx (commented):
@leibale This looks excellent, great addition to the preceding code! Your deep knowledge of the codebase really shines through with how you were able to expand upon and improve the initial concept.

ChronicLynx:
@leibale I updated this to correct the git author information for my commits. You'll just have to kick off the tests again. Sorry about that! Had it set wrong locally on my machine.

Issue #2345, opened 2022-12-12
"Cannot send commands in PubSub mode" error on ping


TarSzator:
As I understand the sentence 

> The commands that are allowed in the context of a subscribed client are SUBSCRIBE, SSUBSCRIBE, SUNSUBSCRIBE, PSUBSCRIBE, UNSUBSCRIBE, PUNSUBSCRIBE, PING, RESET, and QUIT.

on [Redis Pub/Sub](https://redis.io/docs/manual/pubsub/) 

I should not get the error

> Cannot send commands in PubSub mode

on a ping call with the redis client 

**Environment:**
 - **Node.js Version**: 18.12.1
 - **Redis Server Version**: 7.0.5
 - **Node Redis Version**: 4.5.1
 - **Platform**: Alpine in Docker (Host MacOS 13.0.1 on Apple M1 Max)

TarSzator:
Thank you @leibale 😄 So fingers corsed that you get around to it this week 🤞🏼

TarSzator:
@leibale Where you able to release?

TarSzator:
Thank you @leibale for the response and the workaround hints.

I like your plan for the release 😃

leibale:
@TarSzator apparently PubSub & Sharded PubSub + Cluster is a little bit more complicated than I anticipated, but it's almost ready... sorry for all the delays...

TarSzator:
@leibale Thanks for keeping this issue up2date.

Still planing to finish `PubSub & Sharded PubSub + Cluster` or push this to a later release?

leibale:
I'll release a version with or without it (depends if I'll finish it in time) tomorrow

leibale:
@TarSzator I can't release an RC version of `redis` with an RC version or `@redis/client` without releasing a version for all the sub-packages (because the current "peer dependency" in all the sub-packages does not include RC versions). If you are using `@redis/client` directly I can release an RC version of that.

Sorry for all the delays.

TarSzator:
I should be able to easily switch to `@redis/client` for now. So it would already solve my issue when you for now only release the fix on `@redis/client`

leibale:
@TarSzator `@redis/client@1.5.0-next.0` is on npm :tada:

leibale:
@TarSzator have you tried the "next" version? does it work?

TarSzator:
@leibale Works fine for me. Thanks a lot. When will you release it as fully?

TarSzator:
[...]

PR #2344, opened 2022-12-11
Add support for sharded PubSub

leibale:
Refactor PubSub
fix #2354 - Add support for sharded PubSub
fix #1968 - `PUBSUB SHARDCHANNELS` command
fix #2345 ref #2321 - `PING` in PubSub mode
fix #2302 - error listener note in the main README
Add `minimizeConnections` configuration for clusters

PR #2371, opened 2023-01-09
Allow automatic pings to be sent on pubsub channel

mdouglass:
### Description

If you turn on automatic pings (via pingInterval) on a pubsub channel, you get the error: “Cannot send commands in PubSub mode”. PING is one of the legal commands on a pubsub channel, so we override the ping call to set ignorePubSubMode true and allow the ping to be sent on that channel.

We are unable to use the new pingInterval setting on clients being used for pubsub because of an incorrect error that PING is not allowed on pubsub channels.

---

### Checklist

### DIFF
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -361,7 +361,7 @@ export default class RedisClient<
         this.#pingTimer = setTimeout(() => {
             if (!this.#socket.isReady) return;
 
-            (this as unknown as RedisClientType<M, F, S>).ping()
+            this.#sendCommand(['PING'], { ignorePubSubMode: true })
                 .then(reply => this.emit('ping-interval', reply))
                 .catch(err => this.emit('error', err))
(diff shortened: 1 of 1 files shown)


mdouglass:
fix for https://github.com/redis/node-redis/issues/2372

leibale:
@mdouglass thanks for contributing! this issue was already fixed in #2344 (which is not yet ready for release). In the meantime you can use `@redis/client@1.5.0-next.0` (a "pre-release" from #2344).

mdouglass:
Thanks, I had looked around for something pre-existing and somehow missed that. :)
We use redis instead of @redis/client -- do you know the timeline for a new release?

leibale:
`redis` is using `@redis/client` under the hood, you can override the `@redis/client` version to this one and it should work.
See [this](https://github.com/redis/node-redis/issues/2345#issuecomment-1377737526) for updates.

mdouglass:
Great, thanks

PR #2423, opened 2023-02-23
Execute empty MULTI

Mik13:
### Description

Currently, calling `.exec()` on an "empty" multi command (without calling any other command on it), is not executing `MULTI EXEC` on the server.
This can cause problems when using `WATCH`, because no `MULTI EXEC` means the key is still watched.

### Fix
Execute `MULTI EXEC` even if the multi is empty.

### DIFF
--- a/packages/client/lib/multi-command.spec.ts
+++ b/packages/client/lib/multi-command.spec.ts
@@ -46,14 +46,16 @@ describe('Multi Command', () => {
     });
 
     describe('exec', () => {
-        it('undefined', () => {
+        it('without commands', () => {
             assert.equal(
                 new RedisMultiCommand().exec(),
-                undefined
+                [
+                  { args: ['UNWATCH'] }
+                ]
             );
         });
 
-        it('Array', () => {
+        it('with commands', () => {
             const multi = new RedisMultiCommand();
             multi.addCommand(['PING']);
 
--- a/packages/client/lib/multi-command.spec.ts
+++ b/packages/client/lib/multi-command.spec.ts
@@ -47,11 +47,9 @@ describe('Multi Command', () => {
 
     describe('exec', () => {
         it('without commands', () => {
-            assert.equal(
-                new RedisMultiCommand().exec(),
-                [
-                  { args: ['UNWATCH'] }
-                ]
+            assert.deepEqual(
+                new RedisMultiCommand().queue,
+                []
             );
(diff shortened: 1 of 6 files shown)


leibale:
@Mik13 good catch!
1. IMO executing `MULTI EXEC` (instead of `UNWATCH`) is cleaner.
2. I don't like executing commands when not needed (if there is no `WATCH`, there is no reason to `UNWATCH`/`MULTI EXEC`).

I need to think about that more...

Mik13:
> @Mik13 good catch!
> 
> 1. IMO executing `MULTI EXEC` (instead of `UNWATCH`) is cleaner.
> 2. I don't like executing commands when not needed (if there is no `WATCH`, there is no reason to `UNWATCH`/`MULTI EXEC`).
> 
> I need to think about that more...

Hi @leibale!

To 1: Agree
To 2: Same, thats why I am using one O(1) command instead of O(1) + "Depends on commands in the transaction"

I do not understand the project good enough how we could possibly catch the watch inside of the multi-command when issued outside of course. That would be perfect, if it does not add too much complexity.

packages/client/lib/multi-command.spec.ts
```diff
@@ -46,14 +46,16 @@ describe('Multi Command', () => {
     });
 
     describe('exec', () => {
-        it('undefined', () => {
+        it('without commands', () => {
             assert.equal(
```
Mik13, line 50: Tests fail, needs `assert.deepEqual` instead of `assert.equal`

leibale:
@guyroyse we need your opinion here.. :)

guyroyse:
My thought is that we should call what the user told us to call. If I call MULTI and then call EXEC without doing anything else, then that's what I asked for and that's what I should get. It's cleaner and, more importantly, provides unsurprising behavior to the user.

leibale:
@Mik13 @guyroyse review my changes please? :pray:

guyroyse (approved):
The lack of a space before `as` in places like `[C in keyof M[P]as ExcludeMappedString<C>]` makes me twitch a bit but this looks good to me otherwise.

Mik13 (commented):
Looks good to me 👍 
Thank you two for the fast responses!

guyroyse (approved):

guyroyse (approved):

PR #2433, opened 2023-03-02
Add a note about multiple Pub/Sub channel listeners

jamesw6811:
Clarify that multiple subscriptions create multiple listeners.

### Description

This change informs the developer that multiple subscriptions will create multiple listeners. This is important because redis-cli SUBSCRIBE commands do not behave this way, so it may be counter-intuitive.
I personally spent some time debugging my code and my misunderstanding of this was the root cause.

This changes the pub/sub docs to clarify that multiple subscriptions create multiple listeners.
---

### Checklist

### DIFF
--- a/docs/pub-sub.md
+++ b/docs/pub-sub.md
@@ -31,8 +31,6 @@ The event listener signature is as follows:
 
 ## Subscribing
 
-The code below subscribes a new listener to a channel. Subscribing to the same channel more than once will create multiple listeners which will each be called when a message is recieved.
-
 ```javascript
 const listener = (message, channel) => console.log(message, channel);
 await client.subscribe('channel', listener);
@@ -40,6 +38,8 @@ await client.pSubscribe('channe*', listener);
 // Use sSubscribe for sharded Pub/Sub:
 await client.sSubscribe('channel', listener);
 ```
+
+> ⚠️ Subscribing to the same channel more than once will create multiple listeners which will each be called when a message is recieved.
 
 ## Publishing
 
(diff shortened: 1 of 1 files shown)


leibale:
@jamesw6811 actually that's not true... when you subscribe to an already subscribe channel it won't `SUBSCRIBE` again:
```javascript
// `SUBSCRIBE a`
await client.subscribe('a', message => {
  console.log(message);
});

// no command will be executed on the server
// but the callback will be added to the callback set
await client.subscribe('a', message => {
  console.log(message);
});

await anotherClient.publish('a', 'b');
// `client` will get the message once over the network, and then call both callbacks with the message.
```

jamesw6811:
redis-cli stops accepting new commands after one SUBSCRIBE. I didn't realize this wasn't also the behavior of the api. It might confuse others, so I still think it's a helpful addition.

> 127.0.0.1:6379> subscribe hello
> Reading messages... (press Ctrl-C to quit)
> 1) "subscribe"
> 2) "hello"
> 3) (integer) 1
> subscribe hello     
> subscribe hello

leibale:
@jamesw6811 You can use `telnet` to do whatever you want...
Anyway, you are correct, even if you issue 2 `SUBSCRIBE` on the same channel you'll get the message once, I guess this made you expect the second listener to override the first one (and not just added on top)?

leibale:
@guyroyse :)

jamesw6811:
> @jamesw6811 You can use `telnet` to do whatever you want... Anyway, you are correct, even if you issue 2 `SUBSCRIBE` on the same channel you'll get the message once, I guess this made you expect the second listener to override the first one (and not just added on top)?

Exactly. Expecting either functionality is reasonable, but that documentation one-liner would have saved me some time. Thought it might save some others' time too.

PR #2439, opened 2023-03-13
fix cluster extractFirstKey skip commandOptions() passed to args

carlhopf:
### Description

in cluster mode the node/slot is calculated based on `extractFirstKey`. skip `commandOptions` if passed as first argument

f.ex.

```
cluster.xRead(commandOptions({ isolated: true }), [{ key: 'foo', id: '0-0' }], { BLOCK: 1000 });
```

---

### Checklist

### DIFF
--- a/packages/client/lib/cluster/index.spec.ts
+++ b/packages/client/lib/cluster/index.spec.ts
@@ -190,7 +190,7 @@ describe('Cluster', () => {
     }, {
         ...GLOBAL.CLUSTERS.OPEN,
         clusterConfiguration: {
-            maxCommandRedirections: 0,
+            maxCommandRedirections: 0
         }
     });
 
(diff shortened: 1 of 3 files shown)


leibale:
@carlhopf thanks for contributing! Can you please add a test?

carlhopf:
for the unit test, i've added a simple test that runs a isolated blocking command in cluster mode and expects to read to correct result.

i'm still looking into the internals of node-redis, and would like to make this unit test more specific. let me know if you have suggestions/feedback

---

the suggested chaining `client.withOptions(...).ping();` looks like a great improvement to me 👍

carlhopf:
@leibale 

added/updated the unit test: should assert cluster commands with `commandOptions` go to the correct hash slot without MOVED/ASK

---

the suggested chaining `client.withOptions(...).ping();` looks like a great improvement to me 👍

PR #2444, opened 2023-03-17
remaining FT.* commands - set FIRST_KEY_INDEX to 1

carlhopf:
### Description

remaining FT.* commands also select the cluster node based on FIRST_KEY_INDEX

---

### Checklist

### DIFF
--- a/packages/search/lib/commands/ALIASADD.ts
+++ b/packages/search/lib/commands/ALIASADD.ts
@@ -1,3 +1,5 @@
+export const FIRST_KEY_INDEX = 2;
+
 export function transformArguments(name: string, index: string): Array<string> {
     return ['FT.ALIASADD', name, index];
 }
--- a/packages/search/lib/commands/ALIASDEL.ts
+++ b/packages/search/lib/commands/ALIASDEL.ts
@@ -1,3 +1,5 @@
+export const FIRST_KEY_INDEX = 2;
+
 export function transformArguments(name: string, index: string): Array<string> {
     return ['FT.ALIASDEL', name, index];
 }
--- a/packages/search/lib/commands/ALIASUPDATE.ts
+++ b/packages/search/lib/commands/ALIASUPDATE.ts
@@ -1,3 +1,5 @@
+export const FIRST_KEY_INDEX = 2;
+
 export function transformArguments(name: string, index: string): Array<string> {
     return ['FT.ALIASUPDATE', name, index];
 }
--- a/packages/search/lib/commands/ALTER.ts
+++ b/packages/search/lib/commands/ALTER.ts
@@ -1,5 +1,7 @@
 import { RediSearchSchema, pushSchema } from '.';
 
+export const FIRST_KEY_INDEX = 1;
+
 export function transformArguments(index: string, schema: RediSearchSchema): Array<string> {
     const args = ['FT.ALTER', index, 'SCHEMA', 'ADD'];
     pushSchema(args, schema);
--- a/packages/search/lib/commands/CONFIG_SET.ts
+++ b/packages/search/lib/commands/CONFIG_SET.ts
@@ -1,3 +1,5 @@
+export const FIRST_KEY_INDEX = 1;
+
 export function transformArguments(option: string, value: string): Array<string> {
     return ['FT.CONFIG', 'SET', option, value];
 }
--- a/packages/search/lib/commands/DROPINDEX.ts
+++ b/packages/search/lib/commands/DROPINDEX.ts
@@ -1,3 +1,5 @@
+export const FIRST_KEY_INDEX = 1;
+
 interface DropIndexOptions {
     DD?: true;
 }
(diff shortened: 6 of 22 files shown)


packages/search/lib/commands/ALIASADD.ts
```diff
@@ -1,3 +1,5 @@
+export const FIRST_KEY_INDEX = 1;
```
leibale, line 1: should be `2`?

packages/search/lib/commands/ALIASDEL.ts
```diff
@@ -1,3 +1,5 @@
+export const FIRST_KEY_INDEX = 1;
```
leibale, line 1: should be `2`?

packages/search/lib/commands/ALIASUPDATE.ts
```diff
@@ -1,3 +1,5 @@
+export const FIRST_KEY_INDEX = 1;
```
leibale, line 1: should be `2`?

packages/search/lib/commands/ALIASDEL.ts
carlhopf: though about this for a while:

there could be a setup with each redis-cluster-node having an index with the same name. 

to create an alias on each node, the user can then control/select select the redis-cluster-node via hash tags (on the alias name) and manage aliases.


---


besides (im not sure), isn't redis cluster always using the first key after the command to calulate the hash slot internally? and might reply with MOVED

packages/search/lib/commands/ALIASADD.ts
carlhopf: both alias and index need to be same slot, else

```
127.0.0.1:6379> FT.ALIASADD test:{node-2} idx:{node-1}
(error) CROSSSLOT Keys in request don't hash to the same slot
```

changed FIRST_KEY_INDEX to 2 as seems more reasonable 👍

packages/search/lib/commands/ALIASDEL.ts
carlhopf: both alias and index need to be same slot, else

```
127.0.0.1:6379> FT.ALIASADD test:{node-2} idx:{node-1}
(error) CROSSSLOT Keys in request don't hash to the same slot
```

changed FIRST_KEY_INDEX to 2 as seems more reasonable 👍

packages/search/lib/commands/CONFIG_GET.ts
```diff
@@ -1,3 +1,5 @@
+export const FIRST_KEY_INDEX = 1;
```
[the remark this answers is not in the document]
carlhopf: 👍

packages/search/lib/commands/DICTDEL.ts
```diff
@@ -1,6 +1,8 @@
 import { RedisCommandArguments } from '@redis/client/dist/lib/commands';
 import { pushVerdictArguments } from '@redis/client/dist/lib/commands/generic-transformers';
 
+export const FIRST_KEY_INDEX = 1;
```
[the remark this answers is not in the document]
carlhopf: tested: FT.DICT* commands will always execute on the node they are being sent to

unsure how to correctly handle, and removed FIRST_KEY_INDEX for now 👍

packages/search/lib/commands/DICTDUMP.ts
```diff
@@ -1,3 +1,5 @@
+export const FIRST_KEY_INDEX = 1;
```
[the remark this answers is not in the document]
carlhopf: tested: FT.DICT* commands will always execute on the node they are being sent to

unsure how to correctly handle, and removed FIRST_KEY_INDEX for now 👍

packages/search/lib/commands/DICTADD.ts
```diff
@@ -1,6 +1,8 @@
 import { RedisCommandArguments } from '@redis/client/dist/lib/commands';
 import { pushVerdictArguments } from '@redis/client/dist/lib/commands/generic-transformers';
 
+export const FIRST_KEY_INDEX = 1;
```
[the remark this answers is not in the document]
leibale: It means that just like an index, you should create it in all your shards

packages/search/lib/commands/ALIASADD.ts
```diff
@@ -1,3 +1,5 @@
+export const FIRST_KEY_INDEX = 2;
```
leibale, line 1: Actually `2` is incorrect as well, this should be executed on all shards (same for all alias commands)

leibale:
@carlhopf I guess you close it by mistake?

If you don't have the time to make the changes I can do them.. :)

carlhopf:
the disadvantage i see with creating all indexes (high number) on all cluster nodes, would be that it will not scale horizontally when increasing cluster size. 

the base index memory (even if contains no data) would always be occupied on all nodes?

---

my original reason to shard `FT.*` across the cluster with `FIRST_KEY_INDEX`, was to distribute indexes and hashes to individual nodes (using /redisearch.so only and sharding manually across the cluster with `{hash}:slots`)

but if i understand correctly, this is done automatically under the hood by using RediSearch+coordinator https://redis.io/docs/stack/search/design/overview/#scalable-distributed-search 

hence it might not be useful to set `FIRST_KEY_INDEX` and simply let coordinator take care of it?

Issue #2446, opened 2023-03-19
How to flushAll while using a createCluster instance


egmzy:
### Description

I'm using createCluster as a client instance to run commands like GET and SET. 

`const client: RedisClusterType = createCluster(options)`

I want to implement a deleteAllKeys function but `RedisClusterType` doesn't have FLUSHALL.

```
async deleteAllKeys() {
    return await client.FLUSHALL();
}
```

The only Info I could find is here in the docs:
https://github.com/redis/node-redis/blob/HEAD/docs/clustering.md

It mentions `getSlotMaster()` but no explanation on how to use it. 
Also the code says it's deprecated and we should use `slots()` instead, which I couldn't find docs for.

Any idea on how I can approach this?

leibale:
Until #2431 will be implemented, you can use this to run FLUSHALL on all masters:
```javascript
await Promise.all(
  cluster.masters.map(async master => {
    await (await cluster.nodeClient(master)).flushAll();
  });
);
```

`client.slots` - if you want to get the client of a specific slot
`client.shards` - if you want to work on a "per shard" (master + it's replicas) basis.
`client.masters` - just the masters
`client.replicas` - just the replicas

PR #2447, opened 2023-03-19
docs: added info on how to use flushAll with createCluster

egmzy:
I'm working with redis clusters and was trying to implement my own abstraction layer on top of node-redis. 
In that layer I needed a `deleteAllKeys` function, which is based on the `flushAll` method that doesn't exist on `RedisClusterType`.

There were no related docs or mentions on how to work around it, thought it should be added to the docs: https://github.com/redis/node-redis/issues/2446#issuecomment-1475252348


leibale:
@egmzy thanks!! :)

@guyroyse @simonprickett

PR #2476, opened 2023-04-19
Fix PubSub isActive with Sharded PubSub

sorilove:
Missing PubSubType.SHARDED and duplicate comparison of PubSubType.CHANNELS when comparing listeners size in updateIsActive function

### Description

> When calling the sunsubscribe function, the isActive information is incorrect and the redis client is forcibly disconnected.
In the updateIsActive function, we found that PubSubType.SHARDED was missing where the size of listeners is compared.
The listeners size of PubSubType.CHANNELS is being compared twice.
> We fixed the issue and verified that sunsubscribe is working properly.

---

### Checklist


leibale (approved):
Sorry for the delay, I was on vacation...

LGTM :)
Thanks for contributing!

leibale (approved):

PR #2503, opened 2023-05-10
Add support for client reply command

### LINKED ISSUE #1913: Add support for redis command: CLIENT REPLY

aarond-sp:
### Description

Add support for [client reply](https://redis.io/commands/client-reply/) command. Fixes #1913.

---

### Checklist

### DIFF
--- a/packages/client/lib/client/commands.ts
+++ b/packages/client/lib/client/commands.ts
@@ -24,6 +24,7 @@ import * as CLIENT_KILL from '../commands/CLIENT_KILL';
 import * as CLIENT_LIST from '../commands/CLIENT_LIST';
 import * as CLIENT_NO_EVICT from '../commands/CLIENT_NO-EVICT';
 import * as CLIENT_PAUSE from '../commands/CLIENT_PAUSE';
+import * as CLIENT_REPLY from '../commands/CLIENT_REPLY';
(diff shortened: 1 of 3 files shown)


leibale:
Thanks for contributing! But... this command is a little bit tricky, because after `CLIENT REPLY OFF` the behavior of the client changes:
1. The command promise should be resolved as soon as the command is written to the socket (as there is no other way to know if it was executed or not).
2. The types should change (all the commands should change to `Promise<void>`).

The "Good First Issue" label on the issue was misleading, so I removed it now... Sorry about that, and thanks again for contributing! :)

adoeppner:
> Thanks for contributing! But... this command is a little bit tricky, because after `CLIENT REPLY OFF` the behavior of the client changes:
> 
> 1. The command promise should be resolved as soon as the command is written to the socket (as there is no other way to know if it was executed or not).
> 2. The types should change (all the commands should change to `Promise<void>`).
> 
> The "Good First Issue" label on the issue was misleading, so I removed it now... Sorry about that, and thanks again for contributing! :)

You're right. I'll look later into it and try to implement a solution

leibale:
@adoeppner I've converted this PR to a draft PR, please mark it "ready for review" when you finish working on it :)

PR #2534, opened 2023-06-11
added example usage of set datastructure

avinashkrishna613:
### Description

I faced some difficulties while working on sets and later found the value for set should be string. So i just add the example for how to use set datastructure in a node application.


leibale:
Maybe just extend and rename [this](https://github.com/redis/node-redis/blob/master/examples/set-scan.js) example instead of creating a new one?

avinashkrishna613:
Made changes as suggested @leibale

leibale:
@avinashkrishna613 I made some more changes, I think it's ready.. wanna review my changes please? :pray:

examples/set.js
```diff
@@ -1,39 +1,28 @@
...
-// another alternate way to retrieve values from the set
-// https://redis.io/commands/sscan/
-let iCursor = 0;
-do {
-  const { cursor, members } = await client.SSCAN(setName, iCursor);
-  members.forEach((member) => {
-    console.log(member);
-  });
-  iCursor = cursor;
-} while (iCursor !== 0)
```
avinashkrishna613: I think we can keep this. Let's say for example, if user needs some kind of pagination or something then he can make use of this method.
Rest everything looks fine.

leibale: @avinashkrishna613 this is exactly how "sScanIterator" works, there is no reason to implement it again..

avinashkrishna613: @leibale Then it's fine.

simonprickett (changes requested):
This also needs to update the `README.md` file to remove the old set example and add the new one: https://github.com/redis/node-redis/blob/master/examples/README.md

avinashkrishna613:
Updated readme @simonprickett

examples/README.md
```diff
@@ -88,7 +88,7 @@ const client = createClient();
 
 await client.connect();
 
-// Add your example code here...
+// Add your example code here...README
```
simonprickett, line 91:
```suggestion
// Add your example code here...
```

Removed `README` as it doesn't need to be here.

simonprickett (approved):
Approved with suggestion.

avinashkrishna613:
Any update on this @guyroyse ?

avinashkrishna613:
Anything needs to be done on this? @leibale @simonprickett

PR #2535, opened 2023-06-11
fix #1970 - add support for `RESTORE`

### LINKED ISSUE #1970: Add support for redis command: RESTORE

Moshe-RS:
### Description

Currently, the DUMP command exists in the client, but there is no way to update the server with that dumped values.
The current pull request adds the RESTORE functionality and allows this.
It follows issue #1970 request.

- A new file that builds the RESTORE command line arguments.
- Update to the general client's commands list.
- A test file for the restore functionality (unit + integration).

---

### Checklist

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -3,7 +3,7 @@
 This folder contains example scripts showing how to use Node Redis in different scenarios.
 
 | File Name                                | Description                                                                                                                                          |
-| ---------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- |
+|------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `blocking-list-pop.js`                   | Block until an element is pushed to a list.                                                                                                          |
 | `bloom-filter.js`                        | Space efficient set membership checks with a [Bloom Filter](https://en.wikipedia.org/wiki/Bloom_filter) using [RedisBloom](https://redisbloom.io).   |
 | `check-connection-status.js`             | Check the client's connection status.                                                                                                                |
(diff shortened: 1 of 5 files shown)


leibale:
These commands are for "advanced" users, most users will never use them. I don't think we need a section in the readme, but an example in the examples folder will be awesome :)

dtikhonov:
It would be great if these functions supported streams so that they could be used with large keys.

leibale:
@dtikhonov Currently redis (the server) does not RESP3 streaming, which means to be able to stream you'll have to know the size of the string/buffer ahead of time, which makes it not so useful. Once Redis will implement "[streamed strings](https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md#streamed-strings)", we'll be able to support that from the client side as well.

Regarding the PR itself - overall looks good, I'll merge it soon. :)

Thanks for contributing!

dtikhonov:
@leibale I just meant something like this: #2552

leibale:
I thought you want to stream values *into* redis, not the other way around... I really like this idea, and it could fit nicely with [the new `typeMapping` feature in v5](https://github.com/redis/node-redis/blob/v5/docs/command-options.md#type-mapping). The only problem I see is what to do when a string is inside a nested data type (array/map)? Maybe [just like with map keys and set members](https://github.com/redis/node-redis/blob/v5/docs/RESP.md#map-keys-and-set-members), when decoding BlobString to `Stream`, if the BlobString is in a nested data type, it'll be decoded to `string` (or maybe `Buffer` make more sense?) instead.

WDUT?

dtikhonov:
I was thinking that both DUMP and RESTORE could benefit from ability to stream that.  The PR was just to show one part of it.  (It was simple to write because I was just looking at that part of the code recently.)

> The only problem I see is what to do when a string is inside a nested data type

We only have to deal with it if we want to use streams in type maps.  Is it useful to type all strings as streams in a instance of a Redis client?

The API I envision would something like this:

```javascript
redis.dumpAsStream(key).pipe(outputSteam);

redis.restore(key, 0, inputStream);

/* Anything can be a stream, the client code is smart enough to do this as well: */
redis.hset("foo", "bar", inputStream);
```

leibale:
@dtikhonov

Streaming *to* Redis is problematic at the moment because Redis expects an array of "blob strings" as input, each blob string is encoded as `"${length}\r\n{value}\r\n"`, so we cannot start streaming unless we know the size of the stream in advance.. (see [encoder.ts](https://github.com/redis/node-redis/blob/master/packages/client/lib/client/RESP2/encoder.ts) for reference) We'll be able to do it only when "[streamed strings](https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md#streamed-strings)" will be implemented on the server side.

Streaming *from* Redis is something we can do, but I think that string is the only "streamable" data type, how would you stream a map?

Regarding "Is it useful to type all strings as streams in an instance of a Redis client?" you don't have to do it for the whole client, you'll be able to do:
```javascript

const streamClient = client.withTypeMapping({
  [RESP_TYPES.BLOB_STRING]: ReadableStream
});

await client.get('key'); // Promise<string>
await streamClient.get('key'); // Promise<ReadableStream>
```

Hope this is clearer...

leibale:
@Moshe-RS `redis@4.6.9`/`@redis/client@1.5.10` is on npm :tada:

PR #2541, opened 2023-06-19
Add support for `PUBSUB SHARDNUMSUB`

david02324:
### Description

very similar to the existing implementation of the PUBSUB NUMSUB command. :)

---

### Checklist

### DIFF
--- a/packages/client/lib/commands/PUBSUB_SHARDNUMSUB.spec.ts
+++ b/packages/client/lib/commands/PUBSUB_SHARDNUMSUB.spec.ts
@@ -1,6 +1,6 @@
 import { strict as assert } from 'assert';
 import testUtils, { GLOBAL } from '../test-utils';
-import { transformArguments } from './PUBSUB_NUMSUB';
+import { transformArguments } from './PUBSUB_SHARDNUMSUB';
 
 describe('PUBSUB SHARDNUMSUB', () => {
     describe('transformArguments', () => {
--- a/packages/client/lib/commands/PUBSUB_SHARDNUMSUB.spec.ts
+++ b/packages/client/lib/commands/PUBSUB_SHARDNUMSUB.spec.ts
@@ -31,5 +31,15 @@ describe('PUBSUB SHARDNUMSUB', () => {
             await client.pubSubShardNumSub(),
             Object.create(null)
         );
+
+        assert.deepEqual(
+            await client.pubSubShardNumSub('channel'),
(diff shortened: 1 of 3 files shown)


packages/client/lib/commands/PUBSUB_SHARDNUMSUB.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    if (channels) return pushVerdictArguments(args, channels);
+
+    return args;
+}
+
+export function transformReply(rawReply: Array<string | number>): Record<string, number> {
+    const transformedReply = Object.create(null);
+
+    for (let i = 0; i < rawReply.length; i +=2) {
```
younes-io, line 19: You can use `entries()` to iterate over the array of records.. something like:
```js
 for (const [channel, numSubscribers] of rawReply.entries()) {
    transformedReply[channel] = numSubscribers;
  }
  ```

Comment (david02324) [reply packages/client/lib/commands/PUBSUB_SHARDNUMSUB.ts:19]: Every first factor of the iterator returned by the entries() is index, so I think it is not an appropriate use 🤔

Comment (younes-io) [reply packages/client/lib/commands/PUBSUB_SHARDNUMSUB.ts:19]: How does "rawReply" look like ?

Comment (david02324) [reply packages/client/lib/commands/PUBSUB_SHARDNUMSUB.ts:19]: https://redis.io/commands/pubsub-shardnumsub/

Same as reply in pubsub numsub. Repeated channel(string) and count(number)

File: packages/client/lib/commands/PUBSUB_SHARDNUMSUB.spec.ts [hunk 2]
```diff
@@ -0,0 +1,35 @@
...
+        it('array', () => {
+            assert.deepEqual(
+                transformArguments(['1', '2']),
+                ['PUBSUB', 'SHARDNUMSUB', '1', '2']
+            );
+        });
+    });
+
+    testUtils.testWithClient('client.pubSubShardNumSub', async client => {
+        assert.deepEqual(
```
leibale, line 30: can you please make the test "non empty" so it'll actually test `transforReply` as well?

leibale (commented):
overall looks good, the tests need some changes.. :)

leibale:
@david02324 @younes-io `redis@4.6.9`/`@redis/client@1.5.10` is on npm :tada:

Issue #2031, opened 2022-03-10
ZRANGE DOESN'T CONTAIN THE WITHSCORES OPTION IN YOUR ABSTRACTION, MUST USE SENDCOMMAND INSTEAD


jyoung4242:
**Environment:**
 - **Node.js Version**: 
 - **Redis Server Version**: 
 - **Node Redis Version**: 
 - **Platform**:

jyoung4242:
```JS
//DOESNT WORK
//seen = await client.zRange(`seen:${user}`, 0, -1, { WITHSCORES: true });
//DOES WORK
seen = await client.sendCommand(['ZRANGE', `seen:${user}`, 0, -1, 'WITHSCORES']);
```

leibale:
Because the `WITHSCORES` argument changes the reply type (from `Array<string>` to `Array<{ score: number; value: string; }>` , it's implemented as `zRangeWithScores/ZRANGE_WITHSCORES`:

```javascript
await client.zRangeWithScores('key', 0, -1);
await client.ZRANGE_WITHSCORES('key', 0, -1);

jyoung4242:
Thanks for quick reply

PR #2545, opened 2023-06-21
adding support for WITHSCORES for zrange command (updated from 6.2.0)

cleardarkz:
### Description

Since ZRangeByScore was deprecated from 6.2.0, I've added a small fix for ZRANGE command to allow the WITHSCORES argument


leibale:
See https://github.com/redis/node-redis/issues/2031#issuecomment-1064532303

Issue #2417, opened 2023-02-20
TypeError when connecting to RedisDB in Docker Compose container


JohnnyRacer:
### Description

I am getting a very strange type error everytime I try to use `docker compose up` to launch any script that connects to redis with the following error message:
```
redis-test /node_modules/@redis/client/dist/lib/client/commands-queue.js:62
redis-test                 const { resolve, reject } = __classPrivateFieldGet(this, _RedisCommandsQueue_waitingForReply, "f").shift();
redis-test                         ^
redis-test 
redis-test TypeError: Cannot destructure property 'resolve' of '__classPrivateFieldGet(...).shift(...)' as it is undefined.
redis-test     at Object.onReply (/node_modules/@redis/client/dist/lib/client/commands-queue.js:62:25)
redis-test     at RESP2Decoder.write (/node_modules/@redis/client/dist/lib/client/RESP2/decoder.js:119:26)
redis-test     at RedisCommandsQueue.onReplyChunk (/node_modules/@redis/client/dist/lib/client/commands-queue.js:154:72)
redis-test     at RedisSocket.<anonymous> (/node_modules/@redis/client/dist/lib/client/index.js:388:84)
redis-test     at RedisSocket.emit (node:events:512:28)
redis-test     at Socket.<anonymous> (/node_modules/@redis/client/dist/lib/client/socket.js:202:42)
redis-test     at Socket.emit (node:events:512:28)
redis-test     at addChunk (node:internal/streams/readable:324:12)
redis-test     at readableAddChunk (node:internal/streams/readable:297:9)
redis-test     at Readable.push (node:internal/streams/readable:234:10)
redis-test 
redis-test Node.js v19.6.1
redis-test exited with code 1
```

This error does not happen when I connect to redis using the node.js REPL without docker.

This is the docker-compose.yml file I am using to create the redis container:
```
  redis:
    image: redis/redis-stack:latest
    ports:
      - 6379:6379
    command: ["redis-server", "--appendonly", "yes"]
    volumes:
      - redis-data:/data
```
And the code I am using in node.js to connect to redis 
```javascript

import { createClient } from "redis";
(async function () {
    const rdb = createClient({
        url: 'redis://redis:6379'
        });
   
    await rdb.connect();
})();
```

### Node.js Version

19.6.1

### Redis Server Version

6.2.8

### Node Redis Version

4.6.4

### Platform

Linux

leibale:
[...]

PR #2549, opened 2023-06-26
Handle early server errors (issue #2417)

dtikhonov:
### Description

This fixes a bug where the library does not handle early server errors such as `-DENIED` or `-ERR max number of clients reached`.

There may be a better way to do this, this is just a first stab that seems to work.

---

### Checklist

### DIFF
--- a/packages/client/lib/client/commands-queue.ts
+++ b/packages/client/lib/client/commands-queue.ts
@@ -43,6 +43,7 @@ export default class RedisCommandsQueue {
     readonly #waitingToBeSent = new LinkedList<CommandWaitingToBeSent>();
     readonly #waitingForReply = new LinkedList<CommandWaitingForReply>();
     readonly #onShardedChannelMoved: OnShardedChannelMoved;
+    readonly #earlyServerError: Function;
 
     readonly #pubSub = new PubSub();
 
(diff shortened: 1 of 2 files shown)


leibale:
These errors should be handled, but I don't think this is the way to go.. I think that an early error should be the same as an error from the "`#initator`" function, reset the connection and run the "reconnect strategy". I'll do it in v5, then maybe backport to v4. Will keep you posted.

leibale:
Maybe we should run `PING` if there is no `AUTH` or `SELECT` just to make sure the connection is actually open? need to think about that..

dtikhonov:
That is an extra command that every user of the library will have to execute.  I think the logic _if the server responds without any previous command, install a handler on demand_ is best.

leibale:
@dtikhonov you are right, that makes more sense.. but, instead of pushing a "fake command" into the queue, the logic can be moved to the `onReply` function that is passed to the `#decoder`. Also, when this happens it should destroy the current socket and the "reconnectStrategy" should kick in.

dtikhonov:
> Also, when this happens it should destroy the current socket and the "reconnectStrategy" should kick in.

In case of these errors, the [server closes the connection](https://github.com/redis/redis/blob/e7a3d3d152c251cc25aed3d89a47a525812e72de/src/networking.c#L1321), so we don't need anything extra to kick off the reconnect strategy.  Or did you want to pass the callback this error?  At first glance, it is not obvious to me how to wire this through.

leibale:
@dtikhonov in the case of "max clients" the server closes the connection, but what about other errors (like LOADING)? I think it's reasonable to "restart" when the client reaches an "undefined state"..

dtikhonov:
IIUC, LOADING is a transient error: the connection is not closed in that case.

dtikhonov-iex:
I looked, LOADING does not close the connection:

```c
    /* Loading DB? Return an error if the command has not the
     * CMD_LOADING flag. */
    if (server.loading && !server.async_loading && is_denyloading_command) {
        rejectCommand(c, shared.loadingerr);
        return C_OK;
    }

    /* During async-loading, block certain commands. */
    if (server.async_loading && is_deny_async_loading_command) {
        rejectCommand(c,shared.loadingerr);
        return C_OK;
    }
```

I think we do not need to do much, just handle those errors by bubbling them up.

PR #2565, opened 2023-07-10
Fix: XAUTOCLAIM after a TRIM with pending messages returns nil

Calyhre:
### Description

Running a `XAUTOCLAIM` after a `TRIM` while messages are pending results in Redis returning an array with `nil` as the unique element.

```shell
0.0.0.0:6379> XAUTOCLAIM stream group user 0 0-0 COUNT 1
1) "1689008209763-0"
2) 1) (nil)
```

The generic transformer will then crash [here](https://github.com/redis/node-redis/blob/a7d5bc7ca4da6cde66140ae4fea96a8ea6e7ce64/packages/client/lib/commands/generic-transformers.ts#L100) trying to iterate on `null` with `TypeError: .for is not iterable`

You can reproduce the issue with this:
```ts
const id = await client.XADD("stream", "*", { foo: "bar" });
await client.XREADGROUP("group", "consumer-1", { key: "stream", id: ">" });
await client.XTRIM("stream", "MAXLEN", 0);
await client.XAUTOCLAIM("stream", "group", "consumer-2", 0, id); // <- Will crash
```

---

### Checklist

### DIFF
--- a/packages/client/lib/commands/XAUTOCLAIM.spec.ts
+++ b/packages/client/lib/commands/XAUTOCLAIM.spec.ts
@@ -24,71 +24,75 @@ describe('XAUTOCLAIM', () => {
     });
 
     testUtils.testWithClient('client.xAutoClaim without messages', async client => {
-        await Promise.all([
+        const [,, reply] = await Promise.all([
             client.xGroupCreate('key', 'group', '$', { MKSTREAM: true }),
             client.xGroupCreateConsumer('key', 'group', 'consumer'),
+            client.xAutoClaim('key', 'group', 'consumer', 1, '0-0')
         ]);
 
-        assert.deepEqual(
-            await client.xAutoClaim('key', 'group', 'consumer', 1, '0-0'),
-            {
-                nextId: '0-0',
-                messages: []
-            }
-        );
+        assert.deepEqual(reply, {
+            nextId: '0-0',
+            messages: []
+        });
     }, GLOBAL.SERVERS.OPEN);
 
     testUtils.testWithClient('client.xAutoClaim with messages', async client => {
-        await client.xGroupCreate('key', 'group', '$', { MKSTREAM: true });
-        await client.xGroupCreateConsumer('key', 'group', 'consumer');
-        const id = await client.xAdd('key', '*', { foo: 'bar' });
-        await client.xReadGroup('group', 'consumer', { key: 'key', id: '>' });
+        const [,, id,, reply] = await Promise.all([
+            client.xGroupCreate('key', 'group', '$', { MKSTREAM: true }),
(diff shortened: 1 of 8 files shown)


leibale:
@Calyhre Nice catch! but only `XAUTOCLAIM` can return `null`, not `XCLAIM` and `XRANGE`, so the change shouldn't be made in the `transformStreamMessagesReply` function (which all of them are using).
I've already fixed it in the v5 (see https://github.com/redis/node-redis/pull/2485/commits/c9dae346a295030adc10c46d1b7bf26bc4b46bc5), but I guess we need to backport it to v4 as well...

Calyhre:
@leibale Nice! I also noticed the issue disappeared in Redis V7. 

> but only `XAUTOCLAIM` can return `null`, not `XCLAIM` and `XRANGE`, so the change shouldn't be made in the `transformStreamMessagesReply` function (which all of them are using).

I moved the fix to it's own method in `xAutoClaim` to not pollute the other methods. Let me know what you think.

leibale:
@Calyhre actually both `XCLAIM` and `XAUTOCLAIM` might return `null` before redis 7, see [here](https://github.com/redis/redis/pull/10227) for more details and why it has changed.
I've fixed it in v5 (see 8369448d05b2782d766bc30326f3423c319de504), wanna backport it one last time? :pray:

Calyhre:
@leibale Alright, I think I ended up porting a very similar fix to the one you did on V5. Let me know what you think.

I do have one doubt though: It seems right to me to reflect the differences between Redis versions (returning null or not based on Redis answer), but the end type is not version dependent. Using Redis V7, `x[Auto]Claim` messages will be typed with as `StreamMessageReply | null` even though `null` is not possible anymore in this case.

This might be complicated, but should the typing be dependent of the Redis server version?

leibale:
@Calyhre Thanks for working on that! :)

Regarding the deiffrences bettwen Redis versions -

leibale:
@Calyhre Thanks for working on that! I'll review it next week :)

Regarding the differences between Redis versions - at the moment all the "command interfaces" are implemented with Redis 5-7.2 in mind. We might split the commands into their own sub-package at some point, then we'll be able to publish `@redis/${redis-version}-commands` or something like that... (BTW, we have the exact same problem with command arguments, for example, `SET` with `NX` and `GET` is only supported in Redis 7+, but we show it for all users regardless of their Redis version...)

leibale:
@Calyhre `redis@4.6.9`/`@redis/client@1.5.10` is on npm :tada:

Calyhre:
Awesome thank you! And sorry for committing the hard coded redis test urls 🤦

leibale:
@Calyhre I'll take PRs where all I need to do is remove "hard coded redis tests URLs" any day.. ;) Thank you!

Issue #2601, opened 2023-08-17
Have client.connect() return a Promise<client> instead of Promise<void>


franciscop:
### Motivation

It's not possible to create a client in a single step that defers the creation (even if async), so now if we need a promise of the client we need to do this:

```js
const storeProm = (async() => {
  const client = createClient();
  await client.connect();
  return client;
})();
// Can use the client promise
```

If I want to use Redis as a KV store, it needs to be initialized and awaited:

```js
const client = createClient();

// Need to await, which we might not yet be in the right place to do so
await client.connect();

const api = fch.create({ cache: { store: client } });
```

However it'd be nice if we could join both steps into one, so that it could be done in a single step without the extra syntax:

```js
const store = await createClient().connect();
const api = fch.create({ cache: { store }});
```

More specifically, I have a library called [Swear](https://www.npmjs.com/swear) that is able to very nicely wrap the creation promise and avoid any await at all until the "leaf" function is called and awaited:

```js
// Currently
const store = swear((async() => {
  const client = createClient();
  await client.connect();
  return client;
})());
await store.set('a', 'b');
await store.get('a');

// If .connect() returned a Promise<client>
const store = swear(createClient().connect());
await store.set('a', 'b');
await store.get('a');
```

### Basic Code Example

```JavaScript
// Would love to do this:
const store = await createClient().connect();
const api = fch.create({ cache: { store }});

// Which would also unlock also this with my library `swear`:
const store = swear(createClient().connect());
await store.set('a', 'b');
await store.get('a');
```

franciscop:
I'd be happy to try giving it a stab at it, I believe it would be here:

```ts
// Before
connect(): Promise<void> {
    // see comment in constructor
    this.#isolationPool ??= this.#initiateIsolationPool();
    return this.#socket.connect();
}

// After
async connect(): Promise<void> {
    // see comment in constructor
    this.#isolationPool ??= this.#initiateIsolationPool();
    await this.#socket.connect();
    return this;
}
```

leibale:
I like this idea! :) Wanna open a PR?

BTW, don't forget to add a listener to `error`  events as well, otherwise it will crash your application (see [here](https://github.com/redis/node-redis#events))

bigDP:
Hi, Can i take this task?

leibale:
@bigDP This is already done in #2602 (and should've been closed when I merged it, which did not happen for some reason). Sorry for the confusion, and thanks for wanting to contribute! :)

[...]

PR #2602, opened 2023-08-19
 Have client.connect() return a Promise<RedisClient>

franciscop:
### Description

Make the `.connect()` method to return a promise with an instance of itself, so we can do:

```js
const client = await createClient().connect();
```

Issue: [ Have client.connect() return a `Promise<client>` instead of `Promise<void>`](https://github.com/redis/node-redis/issues/2601)

As explained above, added a case that simplifies the creation of the client for those who want to initialize AND connect at the same time, while retaining the ability to do so separately as before otherwise.

Added a test to make sure the return of `connet()` is the same client instance. I noticed there was no test explicitly for `.connect()`, so created it under the `describe('.connect()')` name.

---

### Checklist

- [x] Does `npm test` pass with this change (including linting)? ([tests](https://github.com/franciscop/node-redis/actions/runs/5911314679))

### DIFF
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -108,10 +108,12 @@ describe('Client', () => {
     });
 
     describe('connect', () => {
-        testUtils.testWithClient('connect returns an instance', async (client) => {
-            const client2 = await client.connect();
-            assert.equal(client, client2);
-            await client.disconnect();
+        testUtils.testWithClient('connect should return the clietn instance', async client => {
+            try {
+                assert.equal(await client.connect(), client);
+            } finally {
+                if (client.isOpen) await client.disconnect();
+            }
         }, {
             ...GLOBAL.SERVERS.PASSWORD,
             disableClientSetup: true
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -112,7 +112,10 @@ describe('Client', () => {
             const client2 = await client.connect();
             assert.equal(client, client2);
             await client.disconnect();
-        }, GLOBAL.SERVERS.PASSWORD);
+        }, {
+            ...GLOBAL.SERVERS.PASSWORD,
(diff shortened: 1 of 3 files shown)


leibale:
@franciscop thanks for contributing! I made some changes, wanna review them? 🙏

packages/client/lib/client/index.ts
```diff
@@ -426,11 +426,11 @@ export default class RedisClient<
...
     }
 
-    async connect(): Promise<RedisClientType<M, F, S>> {
+    async connect() {
         // see comment in constructor
         this.#isolationPool ??= this.#initiateIsolationPool();
         await this.#socket.connect();
-        return this;
+        return this as unknown as RedisClientType<M, F, S>;
     }
```
franciscop, line 434: Looks good! I am not familiar enough with TS to know if there's any meaningful difference here, but I've seen that style in the past so sounds good.

packages/client/lib/client/index.spec.ts
```diff
@@ -108,10 +108,12 @@ describe('Client', () => {
     });
 
     describe('connect', () => {
-        testUtils.testWithClient('connect returns an instance', async (client) => {
-            const client2 = await client.connect();
-            assert.equal(client, client2);
-            await client.disconnect();
+        testUtils.testWithClient('connect should return the clietn instance', async client => {
```
franciscop, line 111:
```suggestion
        testUtils.testWithClient('connect should return the client instance', async client => {
```

leibale:
I'll merge it and release it "soon" (along with a few more PRs) :tada:
Thanks a bunch! :)

leibale:
@franciscop `redis@4.6.9`/`@redis/client@1.5.10` is on npm :tada:

franciscop:
Thanks!! 🥳🥳

leppaott:
How about `cluster.connect()`?

leibale:
@leppaott in v5.. :)

PR #2653, opened 2023-11-15
fix #2632 - handle socket close in "socket initiator" phase

### LINKED ISSUE #2632: Connection hangs after temporarily lost connection from redis
### Description

If I open a connection like this:

```typescript
const conn = await createClient(options)
      .on("connect", () => {
        console.log("redis connect")
      })
      .on("ready", () => {
        console.log("redis ready")
      })
      .on("end", () => {
        console.log("redis end")
      })
      .on("error", async (error) => {
        console.log("redis error", error)
      })
      .on("reconnecting", () => {
        console.log("redis reconnecting")
      })
      .connect()

await conn.set('key', 'value');

```

I see this logs:

```
>> redis connect 
>> redis ready 
```

thank I can use the connection without any issue.

If I restart Redis to simulate a connection lost, I see an automatic reconnection with the following logs:

```
>> redis error SocketClosedUnexpectedlyError: Socket closed unexpectedly
>> redis reconnecting
>> redis connect
```

First of all, I never see ready and then if I try to use the connection `await conn.set('key', 'value');` the call remain blocked indefinitely.

### Node.js Version

v16.13.0 + v18.17.1

### Redis Server Version

6.2.6

### Node Redis Version

4.6.10

### Platform

linux / docker

### Logs

_No response_

leibale:
### Description

fix #2632.

Until this fix, when the server closes the socket (for example, when the server is restarting) in the "socket initiator" phase (while running commands like `AUTH`, which has to run before all commands), the client hangs.

### DIFF
--- a/packages/client/lib/client/socket.ts
+++ b/packages/client/lib/client/socket.ts
@@ -199,7 +199,7 @@ export default class RedisSocket extends EventEmitter {
                         .off('error', reject)
                         .once('error', (err: Error) => this.#onSocketError(err))
                         .once('close', hadError => {
-                            if (!hadError && this.#isReady && this.#socket === socket) {
+                            if (!hadError && this.#isOpen && this.#socket === socket) {
                                 this.#onSocketError(new SocketClosedUnexpectedlyError());
                             }
                         })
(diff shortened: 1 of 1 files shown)


gguimond (commented):
It fixes the issue, the only drawback being that the error event is fired.twice in this case.

PR #2664, opened 2023-12-05
Sentinel Support

sjpotter:
### Description

An object that provides a managed client for accessing Redis clusters built around the Redis Sentinel protocol.

### Checklist


packages/client/lib/client/socket.ts
```diff
@@ -28,15 +28,15 @@ export interface RedisSocketCommonOptions {
   reconnectStrategy?: false | number | ((retries: number, cause: Error) => false | Error | number);
 }
 
-type RedisNetSocketOptions = Partial<net.SocketConnectOpts> & {
+export interface RedisNetConnectOpts extends Omit<Partial<net.TcpNetConnectOpts>, 'keepAlive'>, Partial<net.IpcNetConnectOpts>, RedisSocketCommonOptions {
```
leibale, line 31: I'm pretty sure this solves #2679, need to check (and if it does, cherry-pick it)

packages/client/lib/sentinel/index.ts
```diff
@@ -0,0 +1,1495 @@
...
+    let iter = 0;
+
+    const watchEpoch = clientInfo?.watchEpoch;
+
+    while (true) {
+      if (this.#connectPromise !== undefined) {
+        await this.#connectPromise;
+      }
+
+      if (execType === "WATCH") {
```
leibale, line 719: This should be handled in `RedisClient`

packages/client/lib/sentinel/index.ts
```diff
@@ -0,0 +1,1495 @@
...
+  connect() {
+    return this.self.#internal.connect();
+  }
+
+  async _execute<T>(
+    execType: execType | undefined,
+    isReadonly: boolean | undefined,
+    fn: (client: RedisClient<RedisModules, RedisFunctions, RedisScripts, RespVersions, TypeMapping>) => Promise<T>
+  ): Promise<T> {
+    let clientInfo: clientInfo | undefined;
```
leibale, line 376: Share the master lease between coucurrent runs

sjpotter: implemented this and pushed, but not sure it makes sense as is.

we only get a lease on a write command so a pipeline of read/write/read will only get the leased client on the second command.  I think it might be better the way it was before.

PR #2682, opened 2024-01-03
invalidate watches when client reconnects

sjpotter:
also provide API for other wrapper clients to also create invalid watch states programatically.

Reasoning: if a user does a WATCH and then the client reconnects, the watch is no longer active, but if a user does a MULTI/EXEC after that, they wont know, and since the WATCH is no longer active, the request has no protection.

The API is needed for when a wrapper client (say sentinel, cluster) might close the underlying client and reopen a new one transparently to the user.  Just like in the reconnection case, this should result in an error, but its up to the wrapping client to provide the appropriate error

### Description

---

### Checklist


sjpotter:
this PR is how I imagine the API as needed, need to remove WATCH/UNWATCH command files probably now.

packages/client/lib/client/index.ts
```diff
@@ -279,6 +279,9 @@ export default class RedisClient<
   #monitorCallback?: MonitorCallback<TYPE_MAPPING>;
   private _self = this;
   private _commandOptions?: CommandOptions<TYPE_MAPPING>;
+  #dirtyWatch?: string;
+  #epoch: number;
```
sjpotter, line 283: does it default to Zero? don't see it set besides the ++ on 'ready'

sjpotter: wondering if the epoch should be moved into the RedisSocket and not in the client itself.

leibale: 100% agree with you

leibale: fixed

leibale (approved):

sjpotter:
LGTM

PR #2689, opened 2024-01-15
cache clientId on socket connection, to not require it to be awaited

sjpotter:
### Description

clientId will be removed from a redis command sent over the wire each time (and therefore need to be awaited) and instead cached.  This moves it from the dynamic RedisClientType to the static RedisClient making it more usable for internal functionality.

I don't think this should break backwards compatibility as await on number should in effect be the same thing as await on Promise<number> (but I might be wrong?)

### Checklist


leibale:
1. it breaks backward compatibility for `redis-server < 5.0.0` (the main README does say its the minimum supported version because its the minimum version we test with, but im not sure if we wanna fully break it from older versions).
2. it adds some overhead for something most users don't need.

sjpotter:
Ok, I buy the < 5.0 as some sort of reasoning, but in terms of overhead, I think its actually something users "do" need.  i.e. for watch tracking across reconnect, while it can be accomplished in other ways, this kills many birds with one stone without tracking other things.

PR #2690, opened 2024-01-16
Fix pool task removal

sjpotter:
### Description

When the pool added pending tasks, it did so with a timeout timer, that could remove it from the pending queue.  However, it never cleared this timer if the task is executed.  Resulting in items being removed multiple times from list and thereby breaking it.

---

### Checklist


leibale:
Nice catch! :)
I'm not sure about adding the `removed` property thought, the linked list is not exposed and is intended for internal use only, I think that we should go with performance rather than "defensive code" in this case.. WDUT?

leibale:
Already fixed in the other branches

Issue #2705, opened 2024-02-20
Support XREAD reading from last message of stream


chayim:
As documented [here](https://github.com/redis/redis/issues/7388), to be tested against redis unstable.

PR #2711, opened 2024-02-27
Support the NOVALUES option of HSCAN

gerzse:
### Description

Issue #2705

The NOVALUES option instructs HSCAN to only return keys, without their values.
The new option will become available from Redis 7.6.

### Checklist

### DIFF
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -808,7 +808,10 @@ describe('Client', () => {
         }
 
         assert.deepEqual(keys.sort(sort), expectedKeys);
-    }, GLOBAL.SERVERS.OPEN);
+    }, {
+        ...GLOBAL.SERVERS.OPEN,
+        minimumDockerVersion: [7, 4]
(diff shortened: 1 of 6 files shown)


gerzse:
> @gerzse arguments that change the reply type should be implemented as their own command, see [HRANDFIELD_COUNT](https://github.com/redis/node-redis/blob/master/packages/client/lib/commands/HRANDFIELD_COUNT.ts) and [HRANDFIELD_COUNT_WITHVALUES](https://github.com/redis/node-redis/blob/master/packages/client/lib/commands/HRANDFIELD_COUNT_WITHVALUES.ts) for example.
> 
> For this command, we should have:
> 
> 1. `HSCAN/hScan`
> 2. `HSCAN_NOVALUES/hScanNoValues`
> 3. `hScanIterator`
> 4. `hScanNoValuesIterator`

@leibale Got it, thanks! Makes sense, the same is done in other clients too. I'll adapt the code.

sjpotter (approved):

PR #2713, opened 2024-03-05
asking / command can be separated on cluster redirect.

sjpotter:
this handles normal commands, but it doesn't handle the multi (exec or pipeline) cases.

unsure if pipeline makes sense in the context of cluster because of this (i.e. can get exception in middle, and will have no idea which commands were executed).  Handling pipeline might require a total rewrite.

to handle multi, I think we will have to pass an optional chainId into the executeMulti(), and only set chainId if it's undefined.  With multi, it either all executes or none of it executes, so an exception will be meaningful.


leibale:
already part of #2716

PR #2716, opened 2024-03-13
Command Parser 

sjpotter:
### Description

Start of implementation of changing how command arguments are transformed to enable easier and more consistent usage across the different clients we have and to enable client side caching.

---

### Checklist

### DIFF
--- a/packages/client/lib/client/parser.ts
+++ b/packages/client/lib/client/parser.ts
@@ -19,6 +19,7 @@ export interface CommandParser {
   setPreserve: (val: unknown) => unknown;
 }
 
+// TODO: make multiple parsers to improve performance?
 export class BasicCommandParser implements CommandParser {
   #redisArgs: Array<RedisArgument> = [];
   #keys: Array<RedisArgument> = []; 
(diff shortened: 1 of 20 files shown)


packages/client/lib/client/parser.ts
```diff
@@ -0,0 +1,112 @@
...
+      this.#redisArgs.push(keys);
+    }
+  }
+
+  setPreserve(val: unknown) {
+    this.#preserve = val;
+  }
+
+  setCachable() {
+    this.#cachable = true;
```
leibale, line 111: is cachable "dynamic" (can be either `true` or `false` depending on the command arguments)? if not, we should move it from the parser to a constant property on the command.

packages/client/lib/cluster/index.ts
```diff
@@ -462,8 +526,14 @@ export default class RedisCluster<
             throw new Error(`Cannot find node ${address}`);
           }
 
-          await redirectTo.asking();
           client = redirectTo;
+
+          const chainId = Symbol('Asking Chain');
```
leibale, line 531: is this fix part of another PR too?

packages/client/lib/client/parser.ts
leibale: @sjpotter

PR #2717, opened 2024-03-14
Handle the invalidate OOB push message.

sjpotter:
### Description

Provides a register callback that if set will be called when the client receives a key invalidate message from the server.

---

### Checklist


sjpotter:
this should probably be closed and https://github.com/redis/node-redis/pull/2735 used instead.

PR #2724, opened 2024-03-25
move some usage of RedisArgument arrays to ReadonlyArray

sjpotter:
motivated by wanting to return a read only array from the parser object


leibale:
already part of #2716

Issue #2704, opened 2024-02-16
Cannot read properties of undefined (reading 'replicas')  [calling "get"]


danielbush:
### Description

Environment:

- client: node 18 (public.ecr.aws/docker/library/node:18-alpine)
- redis server: AWS elasticache
    - engine version: 6.2.6
    - 2 shards; 2 nodes per shard

We're getting this error - I think it's intermittent ie it works maybe even a lot of the time:

Stacktrace (redactions using `xxx`):

```
TypeError: Cannot read properties of undefined (reading 'replicas')
    at RedisClusterSlots.getSlotRandomNode (/server/node_modules/@redis/client/dist/lib/cluster/cluster-slots.js:118:19)
    at RedisClusterSlots.getClient (/server/node_modules/@redis/client/dist/lib/cluster/cluster-slots.js:110:37)
    at Commander._RedisCluster_execute (/server/node_modules/@redis/client/dist/lib/cluster/index.js:217:79)
    at Commander.sendCommand (/server/node_modules/@redis/client/dist/lib/cluster/index.js:129:98)
    at Commander.commandsExecutor (/server/node_modules/@redis/client/dist/lib/cluster/index.js:126:75)
    at BaseClass.<computed> [as get] (/server/node_modules/@redis/client/dist/lib/commander.js:8:29)
    at cacheItem (/server/node_modules/xxx:46:34)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async HttpClient.xxx (/server/node_modules/xxx:193:29)
    at async Promise.all (index 1)
    at async prefetchTask (/server/api/controllers/tasks.js:349:5)
    at async Promise.all (index 2)
```

You can see HttpClient (an internal library being asked to make a request.  It then calls `cacheItem` (an internal module) which then calls `redis.get` as per code below to check if there's something in the cache using `get` before making the http request.

Code:

We get the client like this (within an async function).  It is long-running, we keep using the same instance once created.  Note we pass only one url to `rootNodes` - the cluster configuration endpoint provided by AWS Elasticache.  (I wonder if that is an issue as the docs say to pass several nodes, I struggled to find docs on this).

```ts
    ...
    redisClient = redis.createCluster({
        // Assumes we are using AWS Elasticache cluster configuration endpoint
        // or similar.
        rootNodes: [{ url: config.url }],
        defaults: {
            // IMPORTANT: Simply not specifying `socket` here seems to
            // cause the client to hang regardless of what is in
[...]

PR #2731, opened 2024-04-02
Fix race condition when slots are re-calculated

jjsimps:
### Description

I was getting an error similar to <https://github.com/redis/node-redis/issues/2704>. 

What was happening for me is that I was creating a redis cluster client and then calling `client.eval()`. Since `eval` doesn't seem to be cluster-aware, it returned a `MOVE` error since it was routed to an incorrect node. This seems to cause a call to `discover()` to re-discover the hash slot mapping.

The bug I am trying to address is a race condition between when the internal data is cleared to when it's re-populated again. In the current implementation, there's an asynchronous command being run in-between, which means if any other command was run in between, would cause `this.slots[val]` of `cluster-slots.ts` (such as in `getSlotRandomNode()`) to return undefined and throw an exception.

```typescript
 async #discover(rootNode?: RedisClusterClientOptions) {
        this.#resetSlots(); // !!! Reset is being called here
        const addressesInUse = new Set<string>();

        try {
            const shards = await this.#getShards(rootNode), // !!! But there's an async command here
                promises: Array<Promise<unknown>> = [],
                eagerConnect = this.#options.minimizeConnections !== true;
            for (const { from, to, master, replicas } of shards) {
```

This changes moves when it is cleared so that it's cleared then populated, without any async stuff inbetween.

---

### Checklist

### DIFF
--- a/packages/client/lib/cluster/cluster-slots.ts
+++ b/packages/client/lib/cluster/cluster-slots.ts
@@ -158,13 +158,13 @@ export default class RedisClusterSlots<
     }
 
     async #discover(rootNode?: RedisClusterClientOptions) {
-        this.#resetSlots();
         const addressesInUse = new Set<string>();
 
         try {
             const shards = await this.#getShards(rootNode),
                 promises: Array<Promise<unknown>> = [],
                 eagerConnect = this.#options.minimizeConnections !== true;
+            this.#resetSlots();
             for (const { from, to, master, replicas } of shards) {
                 const shard: Shard<M, F, S> = {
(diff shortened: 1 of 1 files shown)


jjsimps:
Hi, can this be taken a look at? I can't seem to add people as reviewers. Thank you.

leibale:
@jjsimps sorry for the delay. I'll take a look at the start of next week. If i forget - ping me.. ;)

leibale (approved):
Good catch! Sorry about the (way too long) dealy

PR #2742, opened 2024-04-17
add support for all hash field expiration commands

sjpotter:
### Description

adds the has field expiration commands - for testing purposes, need a docker container that has it in it.

---

### Checklist

### DIFF
--- a/packages/client/lib/commands/HEXPIRE.ts
+++ b/packages/client/lib/commands/HEXPIRE.ts
@@ -1,0 +1,42 @@
+import { Command, NullReply, RedisArgument } from "../RESP/types";
+import { RedisVariadicArgument, pushVariadicArgument } from "./generic-transformers";
+
+/**
+ * @readonly
+ * @enum {number}
+ */
+export const HASH_EXPIRATION = {
+  /** @property {number} */
+  /** The field does not exist */
+  FieldNotExists: -2,
+  /** @property {number} */
+  /** Specified NX | XX | GT | LT condition not met */
+  ConditionNotMet: 0,
+  /** @property {number} */
+  /** Expiration time was set or updated */
+  Updated: 1,
+  /** @property {number} */
+  /** Field deleted because the specified expiration time is in the past */
+  Deleted: 2
+} as const;
+  
+export type HashExpiration = typeof HASH_EXPIRATION[keyof typeof HASH_EXPIRATION];
+
+export default {
+  FIRST_KEY_INDEX: 1,
+  transformArguments(
+    key: RedisArgument, 
+    fields: RedisVariadicArgument,
+    seconds: number,
+    mode?: 'NX' | 'XX' | 'GT' | 'LT',
+  ) {
+    const args = ['HEXPIRE', key, seconds.toString()];
+
+    if (mode) {
+      args.push(mode);
+    }
+
+    return pushVariadicArgument(args, fields);
+  },
+  transformReply: undefined as unknown as () => NullReply | Array<HashExpiration>
+} as const satisfies Command;
--- a/packages/client/lib/commands/HEXPIREAT.ts
+++ b/packages/client/lib/commands/HEXPIREAT.ts
@@ -1,0 +1,23 @@
+import { RedisArgument, Command, NullReply } from '../RESP/types';
+import { HashExpiration } from './HEXPIRE';
+import { RedisVariadicArgument, pushVariadicArgument, transformEXAT } from './generic-transformers';
+
+export default {
+  FIRST_KEY_INDEX: 1,
+  IS_READ_ONLY: true,
+  transformArguments(
+    key: RedisArgument,
+    fields: RedisVariadicArgument,
+    timestamp: number | Date,
+    mode?: 'NX' | 'XX' | 'GT' | 'LT'
+  ) {
+    const args = ['HEXPIREAT', key, transformEXAT(timestamp)];
+
+    if (mode) {
+      args.push(mode);
+    }
+
+    return pushVariadicArgument(args, fields);
+  },
+  transformReply: undefined as unknown as () => NullReply | Array<HashExpiration>
+} as const satisfies Command;
--- a/packages/client/lib/commands/HEXPIRETIME.ts
+++ b/packages/client/lib/commands/HEXPIRETIME.ts
@@ -1,0 +1,20 @@
+import { RedisArgument, NumberReply, Command, NullReply, ArrayReply } from '../RESP/types';
+import { RedisVariadicArgument, pushVariadicArgument } from './generic-transformers';
+
+export const HASH_EXPIRATION_TIME = {
+  /** @property {number} */
+  /** The field does not exist */
+  FieldNotExists: -2,
+  /** @property {number} */
+  /** The field exists but has no associated expire */
+  NoExpiration: -1,
+} as const;
+    
+export default {
+  FIRST_KEY_INDEX: 1,
+  IS_READ_ONLY: true,
+  transformArguments(key: RedisArgument, fields: RedisVariadicArgument) {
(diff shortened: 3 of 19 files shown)


packages/client/lib/commands/HPERSIST.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { strict as assert } from 'node:assert';
+import testUtils, { GLOBAL } from '../test-utils';
+import HPERSIST from './HPERSIST';
+
+describe('PERSIST', () => {
```
leibale, line 5: Missing the H (in both the tests & the implementation..)

packages/client/lib/commands/HEXPIRETIME.ts
```diff
@@ -0,0 +1,20 @@
...
+  NoExpiration: -1,
+} as const;
+    
+export default {
+  FIRST_KEY_INDEX: 1,
+  IS_READ_ONLY: true,
+  transformArguments(key: RedisArgument, fields: RedisVariadicArgument) {
+    return pushVariadicArgument(['HEXPIRETIME', key], fields);
+  },
+  transformReply: undefined as unknown as () => NullReply | ArrayReply<NumberReply>
```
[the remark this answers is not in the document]
leibale: No it shouldn't..

packages/client/lib/commands/HPERSIST.spec.ts
leibale: Fixed

leibale (commented):
Overall looking good, but:
1. I've made some changes, please review them..
2. Seems like `HGETF` and `HSETF` are missing

sjpotter:
change look fine

packages/client/lib/commands/HEXPIREAT.ts
```diff
@@ -0,0 +1,23 @@
...
+import { RedisVariadicArgument, pushVariadicArgument, transformEXAT } from './generic-transformers';
+
+export default {
+  FIRST_KEY_INDEX: 1,
+  IS_READ_ONLY: true,
+  transformArguments(
+    key: RedisArgument,
+    fields: RedisVariadicArgument,
+    timestamp: number | Date,
+    mode?: 'NX' | 'XX' | 'GT' | 'LT'
```
leibale, line 12: You should be able to do NX and GT at the same time..? see SET as a template

packages/client/lib/commands/HPERSIST.ts
```diff
@@ -0,0 +1,10 @@
+import { RedisArgument, ArrayReply, NumberReply, Command } from '../RESP/types';
+import { RedisVariadicArgument, pushVariadicArgument } from './generic-transformers';
+
+export default {
+  FIRST_KEY_INDEX: 1,
+  transformArguments(key: RedisArgument, fields: RedisVariadicArgument) {
+    return pushVariadicArgument(['HPERSIST', key], fields);
+  },
+  transformReply: undefined as unknown as () => ArrayReply<NumberReply>
```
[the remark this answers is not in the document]
sjpotter: created an enum

sjpotter:
> Overall looking good, but:
> 
>     1. I've made some changes, please review them..
> 
>     2. Seems like `HGETF` and `HSETF` are missing

wondering why you didn't do an enum like response for HEXPIRETIME?

sjpotter:
> > Overall looking good, but:
> > ```
> > 1. I've made some changes, please review them..
> > 
> > 2. Seems like `HGETF` and `HSETF` are missing
> > ```
> 
> wondering why you didn't do an enum like response for HEXPIRETIME?

I see why, as it returns a negative number on "error" (enum) and any positive number on "success", and can't capture that conception in typescript.

sjpotter:
otherwise changes looks fine, and I messed up initial tests, thanks for fixing.

leibale:
We can make an enum for `HEXPIRETIME` and use a type like
```typescript
type x = 1 | 2 | number;
```
I think that some users will find the enum useful..

sjpotter:
> We can make an enum for `HEXPIRETIME` and use a type like
> 
> ```ts
> type x = 1 | 2 | number;
> ```
> 
> I think that some users will find the enum useful..

so I tried something like that, but it seemed like it ended up being shown as just number (as number is a superset of 1/2)

leibale:
It's possible, but a bit more complicated than that
https://www.totaltypescript.com/tips/create-autocomplete-helper-which-allows-for-arbitrary-values

and even if we don't use that in the return type, I think that some users will still find the enum useful (something like `reply === HASH_EXPIRATION.FIELD_DOES_NOT_EXISTS` is more readable than `reply === -2`)

sjpotter:
I agree, but I can't seem to get omit to work with NumberReply as NumerReply expects a number subtype, but Omit doesn't seem to return an number that can be used by it, and when I created a type manually without NumberReply, it didn't seem to make a difference in practice, was still just number.

but yes, I created the enum to use, just unsure how to expose it to the user.

export const HASH_FIELD_EXPIRETIME = {
  /** The field does not exist */
  FIELD_NOT_EXISTS: -2,
  /** The field has no expiration time */
  NO_EXPIRATION_TIME: -1,
} as const;

packages/client/lib/commands/HEXPIRE.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export type HashExpirationReply = NumberReply<HashExpiration>;
+
+export default {
+  FIRST_KEY_INDEX: 1,
+  transformArguments(
+    key: RedisArgument, 
+    fields: RedisVariadicArgument,
+    seconds: number,
+    mode?: 'NX' | 'XX' | 'GT' | 'LT',
```
leibale, line 25: Should we do
```typescript
interface HExpireOptions {
  mode?: 'NX' | 'XX' | 'GT' | 'LT';
}
```
for "future proofing" (in case there will be more optional arguments)

packages/client/lib/commands/HTTL.ts
```diff
@@ -0,0 +1,11 @@
+import { RedisArgument, ArrayReply, NumberReply, Command } from '../RESP/types';
+import { RedisVariadicArgument, pushVariadicArgument } from './generic-transformers';
+
+export default {
+  FIRST_KEY_INDEX: 1,
+  IS_READ_ONLY: true,
+  transformArguments(key: RedisArgument, fields: RedisVariadicArgument) {
+    return pushVariadicArgument(['HTTL', key], fields);
+  },
+  transformReply: undefined as unknown as () => ArrayReply<NumberReply>
```
leibale, line 10: the reply is array of

> Integer reply: -2 if no such field exists in the provided hash key, or the provided key does not exist.
Integer reply: -1 if the field exists but has no associated expiration set.
Integer reply: the TTL in seconds.

should we create an "enum" (not the TS one) for the -2 and -1 values? IDK if I like having them anymore..

sjpotter: an enum can have value to refrer to negative values by name.  but isn't that what I did before ala

```javascript
export const HASH_FIELD_EXPIRETIME = {
  /** The field does not exist /
  FIELD_NOT_EXISTS: -2,
  /* The field has no expiration time */
  NO_EXPIRATION_TIME: -1,
} as const;
```

packages/client/lib/commands/HEXPIRE.ts
sjpotter: it doesn't really future proof, it just makes it easier if these options are shared between multiple commands.  Unsure it adds that much.

elena-kolevska:
Branch redis:v5 has been merged to master, which already contained https://github.com/redis/node-redis/pull/2787.
@nkaradzhov  this can be closed, right?

nkaradzhov:
@elena-kolevska correct, closing this one

PR #2753, opened 2024-05-07
CAE-193: add "IGNORE" options to time series commands (for v5 branch)

sjpotter:
### Description

---

### Checklist

### DIFF
--- a/packages/time-series/lib/commands/ADD.spec.ts
+++ b/packages/time-series/lib/commands/ADD.spec.ts
@@ -57,16 +57,44 @@ describe('TS.ADD', () => {
       );
     });
 
-    it('with RETENTION, ENCODING, CHUNK_SIZE, ON_DUPLICATE, LABELS', () => {
+    it('with IGNORE no values', () => {
+      assert.deepEqual(
+        ADD.transformArguments('key', '*', 1, {
+          IGNORE: { }
+        }),
+        ['TS.ADD', 'key', '*', '1', 'IGNORE', '0', '0']
+      )
+    });
+
+    it('with IGNORE with MAX_TIME_DIFF', () => {
+      assert.deepEqual(
+        ADD.transformArguments('key', '*', 1, {
+          IGNORE: { MAX_TIME_DIFF: 1}
+        }),
+        ['TS.ADD', 'key', '*', '1', 'IGNORE', '1', '0']
+      )
+    });
+
+    it('with IGNORE with MAX_VAL_DIFF', () => {
+      assert.deepEqual(
+        ADD.transformArguments('key', '*', 1, {
+          IGNORE: { MAX_VAL_DIFF: 1}
+        }),
+        ['TS.ADD', 'key', '*', '1', 'IGNORE', '0', '1']
+      )
+    });
+
+    it('with RETENTION, ENCODING, CHUNK_SIZE, ON_DUPLICATE, LABELS, IGNORE', () => {
       assert.deepEqual(
         ADD.transformArguments('key', '*', 1, {
           RETENTION: 1,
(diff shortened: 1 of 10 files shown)


packages/time-series/lib/commands/ADD.spec.ts
```diff
@@ -57,16 +57,44 @@ describe('TS.ADD', () => {
       );
     });
 
-    it('with RETENTION, ENCODING, CHUNK_SIZE, ON_DUPLICATE, LABELS', () => {
+    it('with IGNORE no values', () => {
+      assert.deepEqual(
+        ADD.transformArguments('key', '*', 1, {
+          IGNORE: { }
```
leibale, line 63: Is "IGNORE no values" supported? (https://github.com/RedisTimeSeries/RedisTimeSeries/pull/1543)
If it is - add support for `IGNORE: true` for this case

packages/time-series/lib/commands/ADD.ts
```diff
@@ -11,12 +11,18 @@ import {
   Timestamp
 } from '.';
 
+export interface TsIgnoreOptions {
+  MAX_TIME_DIFF: number;
+  MAX_VAL_DIFF: number;
```
leibale, line 16: Is "IGNORE no values" supported? (https://github.com/RedisTimeSeries/RedisTimeSeries/pull/1543)
If it is - both `MAX_TIME_DIFF` and `MAX_VAL_DIFF` should be optional..

leibale (changes requested):

packages/time-series/lib/commands/ADD.ts
sjpotter: doesn't seem like it, from code

```
int parseIgnoreArgs(RedisModuleCtx *ctx,
                    RedisModuleString **argv,
                    int argc,
                    long long int *ignoreMaxTimeDiff,
                    double *ignoreMaxValDiff) {
    int idx = RMUtil_ArgIndex("IGNORE", argv, argc);
    if (idx > 0) {
        long long maxTimeDiff;
        double maxValDiff;

        if (idx + 2 >= argc ||
            RMUtil_ParseArgs(argv, argc, idx + 1, "ld", &maxTimeDiff, &maxValDiff) !=
                REDISMODULE_OK) {
            RTS_ReplyGeneralError(ctx, "TSDB: Couldn't parse IGNORE");
            return TSDB_ERROR;
        }

        if (maxTimeDiff < 0 || maxValDiff < 0) {
            RTS_ReplyGeneralError(ctx, "TSDB: IGNORE arguments cannot be negative");
            return TSDB_ERROR;
        }

        *ignoreMaxTimeDiff = maxTimeDiff;
        *ignoreMaxValDiff = maxValDiff;
    }

    return TSDB_OK;
}
```

leibale: So if not, why `MAX_TIME_DIFF` and `MAX_VAL_DIFF` are optional? also, rename them to `maxTimeDiff` and `maxValDiff` to match the command interface.

sjpotter: if we want to require the user to specify them as 0, I'm good with that as well

leibale: If the server does not have defaults we shouldn't as well..

packages/time-series/lib/commands/ADD.spec.ts
```diff
@@ -57,14 +57,43 @@ describe('TS.ADD', () => {
       );
     });
 
+    it ('with IGNORE', () => {
+      it('no values', () => {
```
leibale, line 61: invalid tests?

sjpotter: yes, now, didn't remove afte removing from code above.

PR #2766, opened 2024-05-26
module commands to be marked as unstable and require clients to opt in

sjpotter:
infrastrucutre code to enable search module commands to be marked unstable in RESP3 client usage. 

needs, finish conversion of commands to v5, where I'm stuck at the moment


packages/client/lib/client/index.ts
```diff
@@ -867,10 +874,11 @@ export default class RedisClient<
   }
 
   MULTI() {
-    type Multi = new (...args: ConstructorParameters<typeof RedisClientMultiCommand>) => RedisClientMultiCommandType<[], M, F, S, RESP, TYPE_MAPPING>;;
+    type Multi = new (...args: ConstructorParameters<typeof RedisClientMultiCommand>) => RedisClientMultiCommandType<[], M, F, S, RESP, TYPE_MAPPING>;
     return new ((this as any).Multi as Multi)(
       this._executeMulti.bind(this),
-      this._executePipeline.bind(this)
+      this._executePipeline.bind(this),
+      this._self.#options
```
leibale, line 881: ?

packages/client/lib/client/index.ts
```diff
@@ -163,6 +167,9 @@ export default class RedisClient<
   static #createModuleCommand(command: Command, resp: RespVersions) {
     const transformReply = getTransformReply(command, resp);
     return async function (this: NamespaceProxyClient, ...args: Array<unknown>) {
+      if (command.unstableResp3Module && resp == 3 && !this._self._self.#options?.unstableResp3Modules) {
+        throw new Error("unstable resp3 module, client not configured with proper flag");
```
leibale, line 171: need to update the message + include a link

packages/client/lib/client/index.ts
```diff
@@ -163,6 +167,9 @@ export default class RedisClient<
   static #createModuleCommand(command: Command, resp: RespVersions) {
     const transformReply = getTransformReply(command, resp);
     return async function (this: NamespaceProxyClient, ...args: Array<unknown>) {
+      if (command.unstableResp3Module && resp == 3 && !this._self._self.#options?.unstableResp3Modules) {
```
leibale, line 170: move the if + throw to a function and reuse it everywhere?

packages/client/lib/client/index.ts
sjpotter: agreed, but unsure what it should be at the moment.

packages/client/lib/client/index.ts
sjpotter: ok, with what we discussed yesterday, putting the handler on object via attachCommand, this is no longer neccessary, so reverted.

packages/client/lib/client/index.ts
sjpotter: as discussed on zoom, moved this to how the command is actually attached to object in attachCommand

sjpotter: added a FIXME to it, to know we have to work on that string

packages/client/lib/client/index.ts
sjpotter: as discussed on zoom, moved this to how the command is actually attached to object in attachCommand

packages/bloom/lib/commands/top-k/INFO.ts
```diff
@@ -14,12 +21,43 @@ export default {
...
-      depth: reply[5],
-      decay: Number(reply[7])
-    }),
-    3: undefined as unknown as () => TopKInfoReply
+    2: (reply: UnwrapReply<Resp2Reply<TopKInfoReplyMap>>): TkInfoReply => {
+      return {
+        k: reply[1],
+        width: reply[3],
+        depth: reply[5],
+        decay: Number(reply[7])
```
leibale, line 29: Why just `decay`? what about `width` and `depth`?

sjpotter: width/depth are Number's in redis, decay is a double in redis, which in resp2 means a string I believe.  (btw, this isn't a change I made, you did this, this was me trying to intrepret what you did).

see current v5: https://github.com/redis/node-redis/blob/v5/packages/bloom/lib/commands/top-k/INFO.ts / https://github.com/redis/node-redis/blob/fdd1978d92b192c2a85cf50e5dbe5c504270750a/packages/bloom/lib/commands/top-k/INFO.ts

sjpotter: why?  a user can force its type mapping to a string....

leibale: if a user "forces its type to a string" he probably wants a string.. that's a feature, not a bug..

sjpotter: ok.  just means differnt types between resp2/3 and was trying to avoid that.

packages/client/lib/RESP/types.ts
```diff
@@ -128,16 +128,18 @@ export interface MapReply<K, V> extends RespType<
   Map<any, any> | Array<any>
 > {}
 
-type MapKeyValue = [key: BlobStringReply, value: unknown];
+type MapKeyValue = [key: SimpleStringReply, value: unknown];
 
 type MapTuples = Array<MapKeyValue>;
 
+type ExtractMapKey<T> = T extends SimpleStringReply<infer S> ? S : never;
```
leibale, line 135:
```suggestion
type MapKeyValue = [key: BlobStringReply | SimpleStringReply, value: unknown];

type MapTuples = Array<MapKeyValue>;

type ExtractMapKey<T> = (
    T extends BlobStringReply<infer S> ? S :
    T extends SimpleStringReply<infer S> ? S :
    never
```

packages/json/lib/commands/TYPE.ts
```diff
@@ -18,8 +18,8 @@ export default {
   },
   transformReply: {
     2: undefined as unknown as () => NullReply | BlobStringReply | ArrayReply<BlobStringReply>,
-    // TODO: ?!??!
-    3: undefined as unknown as () => any
-  }
+    3: undefined as unknown as () => ArrayReply<NullReply | BlobStringReply | ArrayReply<BlobStringReply>>
```
leibale, line 21: I'm not too sure if this is right..

packages/search/lib/commands/CURSOR_DEL.ts
```diff
@@ -3,8 +3,8 @@ import { SimpleStringReply, Command, RedisArgument } from '@redis/client/dist/li
 export default {
   FIRST_KEY_INDEX: undefined,
   IS_READ_ONLY: true,
-  transformArguments(index: RedisArgument, cursorId: RedisArgument) {
-    return ['FT.CURSOR', 'DEL', index, cursorId];
+  transformArguments(index: RedisArgument, cursorId: number) {
```
leibale, line 6: cursorId can go outside the float64 range.. string has to be supported, number is nice to have

packages/search/lib/commands/CURSOR_READ.ts
```diff
@@ -8,14 +8,15 @@ export interface FtCursorReadOptions {
 export default {
   FIRST_KEY_INDEX: undefined,
   IS_READ_ONLY: true,
-  transformArguments(index: RedisArgument, cursor: RedisArgument, options?: FtCursorReadOptions) {
-    const args = ['FT.CURSOR', 'READ', index, cursor];
+  transformArguments(index: RedisArgument, cursor: number, options?: FtCursorReadOptions) {
```
leibale, line 11: same cursor comment

packages/search/lib/commands/EXPLAIN.ts
```diff
@@ -1,28 +1,28 @@
-// import { RedisArgument, SimpleStringReply, Command } from '@redis/client/dist/lib/RESP/types';
-// import { Params, pushParamsArgs } from ".";
+import { RedisArgument, SimpleStringReply, Command } from '@redis/client/dist/lib/RESP/types';
```
leibale, line 1: Missing tests

packages/time-series/lib/commands/MGET.ts
```diff
@@ -18,14 +19,47 @@ export function pushFilterArgument(args: CommandArguments, filter: RedisVariadic
   return pushVariadicArguments(args, filter);
 }
 
+export type MGetRawReply2 = Array<[
```
leibale, line 22: use `ArrayReply` instead of `Array`

packages/time-series/lib/commands/MGET.ts
```diff
@@ -18,14 +19,47 @@ export function pushFilterArgument(args: CommandArguments, filter: RedisVariadic
...
+  key: string,
```
leibale, line 23: `BlobStringReply`?

sjpotter: yes, it should be BlobStringReply, this was just a port over from v2, but it's a RedisModule_ReplyWithStringBuffer() so BlobString.  Will fix.

packages/json/lib/commands/TYPE.ts
sjpotter: in resp3, the same reply that is possible in resp2 is returned in resp3, just wrapped in an array.

```rust
pub fn json_type<M: Manager>(manager: M, ctx: &Context, args: Vec<RedisString>) -> RedisResult {
    let mut args = args.into_iter().skip(1);
    let key = args.next_arg()?;
    let path = Path::new(args.next_str().unwrap_or(default_path()));

    let key = manager.open_key_read(ctx, &key)?;

    let value = if path.is_legacy() {
        json_type_legacy::<M>(&key, path.get_path())?
    } else {
        json_type_impl::<M>(&key, path.get_path())?
    };

    // Check context flags to see if RESP3 is enabled and return the appropriate result
    if is_resp3(ctx) {
        Ok(vec![value].into())
    } else {
        Ok(value)
    }
}
```

so if one is ok with the resp2, response, the resp3 response looks right to me.

the problem is really in me messing up the conversion in v4, we have

```
export declare function transformReply(): string | null | Array<string | null>;
```

so what we need is the ArrayReply<BlobStringReply | NullReply>

packages/search/lib/commands/CURSOR_DEL.ts
sjpotter: ok, was a copy form v4 which used number.  trying to remember why I changed it from what you did originally, I suspect number might be used elsewhere and needs to be fixed, will look into that.

sjpotter: so I'm also unconvinced that "number" is nice to have, its a "number", but in practice, cursors should be opaque, and we shouldn't encourage users to think of it as a number, so thinking that RedisArgument as is, should be enough (as you had originally)

sjpotter: so the issue here is really due to AGGREGATE_WITHCURSOR (where cursor id is returned to user).  Changed it to be a NumberReply there.  I think before I ran into some problems with tests, will have to redo and see what the problem was.

sjpotter: i.e. the problem is that AGGREGATE_WITHCURSOR when cursor is a NumberReply will default to returning as a number.

instead returning NumberReply (from WITHCURSOR) for cursor and taking an UnwrapReply<NumbeReply> for cursor in cursor read/del and its up to the user to therefore have a proper number conversion setup.

Honestly, not happy with it. I'd want to force the NumberReply to be treated as a string and not as a number in WITHCURSOR (and then can force a string in cursor functions)

packages/search/lib/commands/EXPLAIN.ts
sjpotter: I'm confused.  I see a spec.ts

packages/time-series/lib/commands/index.ts
```diff
@@ -137,28 +139,51 @@ export function pushLabelsArgument(args: Array<RedisArgument>, labels?: Labels)
   return args;
 }
 
-export type SampleRawReply = {
-  2: TuplesReply<[timestamp: NumberReply, value: BlobStringReply]>;
-  3: TuplesReply<[timestamp: NumberReply, value: DoubleReply]>;
+export interface SampleRawReply2 {
+  timestamp: NumberReply;
+  value: BlobStringReply;
 };
```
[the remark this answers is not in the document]
sjpotter: ok, but was a copy form v4

https://github.com/redis/node-redis/blob/d5355d43275fedf1b2afc4db8a926f72b05f79c5/packages/time-series/lib/commands/index.ts#L246-L249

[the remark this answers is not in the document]
sjpotter: copied from v4 badly I think.  will do what you said.

packages/time-series/lib/commands/MGET.ts
sjpotter: done and BlobStringReply instead of string and UnwrapReply in transformReply

packages/time-series/lib/commands/MGET.ts
sjpotter: addressed above

packages/time-series/lib/commands/MGET.ts
sjpotter: though had some issues with typing in the transformReply, needs to be validated.

sjpotter: and same in RANGE/MRANGE/_WITHLABELS/

packages/search/lib/commands/CURSOR_DEL.ts
leibale: "I'd want to force the NumberReply to be treated as a string and not as a number in WITHCURSOR (and then can force a string in cursor functions)" - I'd want the server to reply with a string and not a number..
The server not gonna change, we should probably support both.

packages/search/lib/commands/EXPLAIN.ts
leibale: unit tests only..

packages/time-series/lib/commands/RANGE.ts
```diff
@@ -108,10 +108,10 @@ export default {
   IS_READ_ONLY: true,
   transformArguments: transformRangeArguments.bind(undefined, 'TS.RANGE'),
   transformReply: {
-    2(reply: UnwrapReply<ArrayReply<SampleRawReply[2]>>) {
-      return reply.map(sample => transformSampleReply['2'](sample));
+    2(reply: UnwrapReply<ArrayReply<SampleRawReply2>>) {
```
leibale, line 111: `transformSamplesReply`?
also, do the unwrap inside of the generic transformer instead of in its args, so it'll be easier to use it..

packages/json/lib/commands/TYPE.ts
sjpotter: made the change I mentioned now

packages/search/lib/commands/CURSOR_DEL.ts
sjpotter: so we have the NumberReply that can be both.  With that said, I'd think we should document this (i.e. we strongly reccomend one overrides the type mapping for NumberReplies when using the WITHCURSOR function).

A better solution (which I dont know how to implement) would be to be able to tag commands with a default type mapping that can only override default (blank) client type mapping, but not type mapping that the user specifies.

i.e. by default we might say a NumberReply maps to a a js number, but for some commands we can say it maps to string, and the client would then just do the right thing by default always without user intervention required.

packages/time-series/lib/commands/index.ts
```diff
@@ -137,28 +139,58 @@ export function pushLabelsArgument(args: Array<RedisArgument>, labels?: Labels)
...
     return {
-      timestamp,
-      value: Number(value)
+      timestamp: reply[0],
+      value: Number(reply[1])
     };
   },
-  3(reply: SampleRawReply[3]) {
-    const [timestamp, value] = reply as unknown as UnwrapReply<typeof reply>;
+  3(reply: SampleRawReply3): SampleReply3 {
```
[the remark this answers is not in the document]
sjpotter: so I'm confused about how tuplereply and array reply work.  do they have to be used together?

i.e. in looking at it now i see for range, we basically just have a single element but the function is a an arrayreply wrapping the tuple reply.  I'm confused overall.

packages/time-series/lib/commands/RANGE.ts
sjpotter: confusion mentioned above probably applies here as well.

packages/time-series/lib/commands/index.ts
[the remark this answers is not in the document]
sjpotter: in looking at it more, I dont see why you would make SampleRawReply3 an ArrayReply, I'd think both resp2/resp3 should just be TupleReply and then RANGE et al would have it wrapped in an ArrayReply as get an array of the tuples.  In practice then, resp2/resp3 would be the exact same then (at least for range) and therefore the code could be simplified.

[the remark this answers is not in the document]
leibale: The type in my comments is for samples, but named sample, my bad..

This should work
```typescript
type SampleRawReply = TuplesReply<[timestamp: NumberReply, value: DoubleReply]>;

const transformSampleReply = {
  2(reply: Resp2Reply<SampleRawReply>) {
    const [ timestamp, value ] = reply as unknown as UnwrapReply<typeof reply>;
    return {
      timestamp,
      value: Number(value)
    };
  },
  3(reply: SampleRawReply) {
    const [ timestamp, value ] = reply as unknown as UnwrapReply<typeof reply>;
    return {
      timestamp,
      value
    };
  }
};

type SamplesRawReply = ArrayReply<SampleRawReply>;

const transformSamplesReply = {
  2(reply: Resp2Reply<SamplesRawReply>) {
    return (reply as unknown as UnwrapReply<typeof reply>)
      .map(sample => transformSampleReply[2](sample));
  },
  3(reply: SamplesRawReply) {
    return (reply as unknown as UnwrapReply<typeof reply>)
      .map(sample => transformSampleReply[3](sample));  }
};
```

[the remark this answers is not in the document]
sjpotter: tried implementing this, but cant do it right.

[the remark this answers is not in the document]
sjpotter: maybe got it now

packages/search/lib/commands/EXPLAIN.ts
sjpotter: @leibale thoughts?

packages/search/lib/commands/INFO.ts
```diff
@@ -1,167 +1,177 @@
...
+    return ['FT.INFO', index];
+  },
+  transformReply: {
+    2: transformV2Reply,
+    3: undefined as unknown as () => ReplyUnion
+  },
+  unstableResp3Module: true
+} as const satisfies Command;
 
-// type InfoRawReply = [
```
sjpotter, line 8: I'm wondering if this should be using the TuplesToMapReply helper type?

packages/search/lib/commands/index.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { strict as assert } from 'node:assert';
-import { pushArgumentsWithLength, pushSortByArguments } from '.';
+
+/* import { pushArgumentsWithLength, pushSortByArguments } from '.';
```
sjpotter, line 3: have a feeling this can just be deleted, as functions being tested no longer exist.

packages/time-series/lib/commands/INFO.ts
```diff
@@ -1,82 +1,86 @@
-// import { TimeSeriesAggregationType, TimeSeriesDuplicatePolicies } from '.';
+import { BlobStringReply, Command, NumberReply, SimpleStringReply } from "@redis/client/dist/lib/RESP/types";
+import { TimeSeriesDuplicatePolicies } from ".";
+import { TimeSeriesAggregationType } from "./CREATERULE";
 
-// export const FIRST_KEY_INDEX = 1;
+export type InfoRawReply = [
+  'totalSamples',
```
sjpotter, line 6: same comment re should this be using TuplesToMapReply?

packages/time-series/lib/commands/INFO_DEBUG.ts
```diff
@@ -1,57 +1,57 @@
...
-//     InfoReply,
-//     transformReply as transformInfoReply
-// } from './INFO';
+import { BlobStringReply, Command, NumberReply, SimpleStringReply } from "@redis/client/dist/lib/RESP/types";
+import INFO, { InfoRawReply, InfoReply } from "./INFO";
 
-// export { IS_READ_ONLY, FIRST_KEY_INDEX } from './INFO';
+type InfoDebugRawReply = [
+  ...InfoRawReply,
+  'keySelfName',
```
sjpotter, line 6: same comment re TuplesToMapReply (but perhaps more complicated due to the ... expansion)

packages/bloom/lib/commands/top-k/INFO.ts
sjpotter: ok done.

packages/client/lib/commander.ts
```diff
@@ -15,6 +15,11 @@ interface AttachConfigOptions<
   config?: CommanderConfig<M, F, S, RESP>;
 }
 
+/* FIXME: better error message / link */
```
leibale, line 18: isn't this ready?

sjpotter: the error message is better now, unsure its as good as could be?  If its fine, can remove the comment

packages/client/lib/RESP/types.ts
```diff
@@ -77,7 +77,9 @@ export interface BlobStringReply<
   T,
   Buffer,
   string | Buffer
-> {}
+> {
+  toString(): string
```
leibale, line 81: ?!

sjpotter:
An experiment that makes using the type so much easier.  No need to unwrap
a basic functionality.

packages/bloom/lib/commands/bloom/INFO.ts
```diff
@@ -1,11 +1,63 @@
...
 
 export default {
   FIRST_KEY_INDEX: 1,
   IS_READ_ONLY: true,
   transformArguments(key: RedisArgument) {
     return ['BF.INFO', key];
   },
-  // TODO
-  transformReply: undefined as unknown as () => any
+  transformReply: {
```
leibale, line 25: https://github.com/redis/node-redis/blob/v5/docs/v4-to-v5.md?plain=1#L238
https://github.com/redis/node-redis/discussions/2506

I'm not sure if we should keep the keys as is for the sake of consistency or transform them for the sake of "friendly/javascripty"..

packages/bloom/lib/commands/bloom/INFO.ts
```diff
@@ -1,11 +1,63 @@
...
+        size: reply[3],
+        numberOfFilters: reply[5],
+        numberOfInsertedItems: reply[7],
+        expansionRate: reply[9]
+      };
+    },
+    3(reply: UnwrapReply<BfInfoReplyMap>): BfInfoReply {
+      if (reply instanceof Map) {
+        return {
+          capacity: reply.get('Capacity') as NumberReply,
```
leibale, line 38: The user asked for a `Map`, why are we converting it to `object`?

packages/client/lib/commands/HEXPIRE.ts
```diff
@@ -0,0 +1,43 @@
+import { Command, RedisArgument } from '../RESP/types';
+import { pushVariadicArgument } from './generic-transformers';
+
+/**
+ * @readonly
+ * @enum {number}
+ */
+export const HASH_EXPIRATION = {
+  /** @property {number} */
```
leibale, line 9: ?!

packages/client/lib/client/index.spec.ts
```diff
@@ -117,14 +117,18 @@ describe('Client', () => {
   });
 
   testUtils.testWithClient('connect, ready and end events', async client => {
+    const p1 = once(client, 'connect');
```
leibale, line 120: I think it's better to commentout this test and add a TODO to check whats going on instead of changing it to "work" without knowning why..

packages/bloom/lib/commands/bloom/INFO.ts
sjpotter: it says there "to avoid unneccessary transformations" (and confusion but I honestly don't think that is the case).

In this case a transformation is neccessary anyways, so might as well make it friendly.  I also distinguish between "info" objects that are defined data structures, and generic maps that can contain any set of keys/values (as I'll discuss on the next one)

packages/bloom/lib/commands/bloom/INFO.ts
sjpotter: We aren't defining it as a MapReply return object, we are defining it as an object return object, so the user never sees that it can be converted.

There are a few thoughts

1) what is the advantage of using a Map object over a plain object (Record<>)?  Does it acutally gain us anything?  the keys still can't be binary.

2) info objects are not generic, they are predefined data structures, so the concept of needing to "iterate" over them doesn't seem valid (but goes back to #1, what advantage would a map give in this situation).  type transformation makes sense when they improve the user experience, not when they make it more complicated for zero gain.

packages/client/lib/commands/HEXPIRE.ts
sjpotter: I was trying to write jdoc I guess?  probably don't know what I'm doing with it.

packages/client/lib/client/index.spec.ts
sjpotter: reasonable.  even more reasonable would be to have CI mark it as a failure to ignore (if possible), but have it error out for real when we do npm run test manually, so it says in our mind.

packages/time-series/lib/commands/index.ts
```diff
@@ -168,3 +219,96 @@ export function pushWithLabelsArgument(args: CommandArguments, selectedLabels?:
...
+        //ret[key.toString()] = parseFunc(sample);
+      }
+    
+      return ret as unknown as MapReply<string, T>;
+    }
+  }
+  
+}
+
+export function resp3MapToValue<
```
leibale, line 274: all the transformations can be done in-place, right?

packages/time-series/lib/commands/index.ts
```diff
@@ -137,28 +141,75 @@ export function pushLabelsArgument(args: Array<RedisArgument>, labels?: Labels)
...
+  const labels: Labels = {};
+
+  const reply = r as unknown as UnwrapReply<typeof r>;
+
+  for (const t of reply) {
+    const [k, v] = t as unknown as UnwrapReply<typeof t>;
+    const key = k as unknown as UnwrapReply<BlobStringReply>;
+    const value = v as unknown as UnwrapReply<BlobStringReply>;
+
+    labels[key.toString()] = value.toString()
```
leibale, line 185: `value.toString()`? why?

packages/time-series/lib/commands/index.ts
```diff
@@ -137,28 +141,75 @@ export function pushLabelsArgument(args: Array<RedisArgument>, labels?: Labels)
...
+  }
+
+  return labels
+}
+
+export function transformLablesReply3(r: RawLabels3): Labels {
+  const reply = r as unknown as UnwrapReply<RawLabels3>;
```
leibale, line 192: why labels is always an object and not "MapReply"?

packages/time-series/lib/commands/MRANGE.ts
```diff
@@ -58,10 +60,79 @@ export function transformMRangeArguments(
...
+
+export interface MRangeReplyItem2 {
+  samples: Array<ReturnType<typeof transformSampleReply[2]>>;
+}
+
+export interface MRangeReplyItem3 {
+  samples: Array<ReturnType<typeof transformSampleReply[3]>>;
+}
+
+export function getSamples(
```
leibale, line 98: im not sure i understand what is going on here.. in resp2 the reply with and without groupby is compatible, but in resp3 its not? (right?)
and the code is just ignoring the new field in resp3? (right?)
i'm confused..

packages/time-series/lib/commands/index.ts
sjpotter: conceptually yes.  for simplicity (but not for garbage collector) i recreated the object from scratch, can just have it overwrite the "value".  slightly worried how that would impact iteration, but if its ok to do that,

packages/time-series/lib/commands/index.ts
sjpotter: In V4, it was string->string

https://github.com/redis/node-redis/blob/2fc79bdfb375602e2aaba15962f57c88d8afe46b/packages/time-series/lib/commands/index.ts#L185C1-L199C2

so the Q is, do we want to require labels still be simple strings (and therefore, not dependent on the BlobString type mapping, i.e. force to string).  Or in V5 do we want to say, no, we can support BlobString values for labels.

packages/time-series/lib/commands/index.ts
sjpotter: So, the Q is, what does returning it as all possible map reply type mappings give us (over a simple object).  i.e. related to above Q.  If we define that we only support labels as string->string, I'd argue it gives us minimal if anything over a simple object.  If we want to support binary labels and binary values, then we do gain something.

packages/time-series/lib/commands/MRANGE.ts
sjpotter: yes, the resp2/resp3 responses aren't compatible.  resp2 pushes groupby metadata into the labels (i.e. not really part of labels, but no where else to return them, and hence gets pushed out as labels), while in resp3 they are separate fields in the returned value, that are just ignored (as not labels, and all we do is return labels).

PR #2791, opened 2024-07-12
fix: XRangeOptions COUNT

ljy1017010:
``` ts
interface XRangeOptions {
    COUNT?: number;
}

export function transformArguments(
    key: RedisCommandArgument,
    start: RedisCommandArgument,
    end: RedisCommandArgument,
    options?: XRangeOptions
): RedisCommandArguments {
    const args = ['XRANGE', key, start, end];

    if (options?.COUNT) {
        args.push('COUNT', options.COUNT.toString());
    }

    return args;
}
```

And in examples by Node.js is a number, so i fix it.

---

### Checklist

### DIFF
--- a/doctests/dt-streams.js
+++ b/doctests/dt-streams.js
@@ -49,7 +49,7 @@ assert.equal(await client.xLen('race:france'), 3);
 // REMOVE_END
 
 // STEP_START xRange
-const res4 = await client.xRange('race:france', '1691765278160-0', '+', 2);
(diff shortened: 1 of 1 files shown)


sjpotter:
@dwdougherty can you take a look at this if it makes sense as you wrote (or rewrote) the original doctest.

thanks.

dwdougherty (approved):
LGTM. Thank you @ljy1017010 and @sjpotter .

PR #2814, opened 2024-08-08
understand that null responses can be returned for the key/value array

### LINKED ISSUE #2772: FT.SEARCH: Cannot read properties of undefined (reading 'length')
### Description

Usage: `client.ft.search(index, query, option)`
Exception: `Cannot read properties of undefined (reading 'length')`
Error location:  [https://github.com/redis/node-redis/blob/master/packages/search/lib/commands/SEARCH.ts](https://github.com/redis/node-redis/blob/master/packages/search/lib/commands/SEARCH.ts) 
`while (i < tuples.length) {`

The debug log shows that the reply value is `[100, [] ]`

```js
export type SearchRawReply = Array<any>;

export function transformReply(reply: SearchRawReply, withoutDocuments: boolean): SearchReply {
    const documents = [];
    let i = 1;
    while (i < reply.length) {
        documents.push({
            id: reply[i++],
            value: withoutDocuments ? Object.create(null) : documentValue(reply[i++])
        });
    }

    return {
        total: reply[0],
        documents
    };
}

function documentValue(tuples: any) {
    const message = Object.create(null);

    let i = 0;
    while (i < tuples.length) {
        const key = tuples[i++],
            value = tuples[i++];
        if (key === '$') { // might be a JSON reply
            try {
                Object.assign(message, JSON.parse(value));
                continue;
            } catch {
                // set as a regular property if not a valid JSON
            }
        }

        message[key] = value;
    }

    return message;
}
```

### Node.js Version

v18.19.1

### Redis Server Version

6.2.6

### Node Redis Version

4.6.14

### Platform

macOS

### Logs

```bash
"TypeError: Cannot read properties of undefined (reading 'length')
    at documentValue (/Users/my/test/node_modules/@redis/search/dist/commands/SEARCH.js:30:23)
    at Object.transformReply (/Users/my/test/node_modules/@redis/search/dist/commands/SEARCH.js:18:61)
    at transformCommandReply (/Users/my/test/node_modules/@redis/client/dist/lib/commander.js:90:20)
    at Commander.commandsExecutor (/Users/my/test/node_modules/@redis/client/dist/lib/client/index.js:190:54)
```

sjpotter:
in #2772 users noted that we are throwing a javascript exception as we are trying to apply .length to a null and hence that fails.

we have 2 options.  First, we can just elide these results from the response we give the user, but this wouldn't correspond to the ft.search documentation.  Therefore, instead, as redis doesn't return a document, but a null, we should do the same, and return an empty null object for these documents.  It's up to the user to handle them

fixes #2772 

### Description

---

### Checklist

### DIFF
--- a/packages/search/lib/commands/SEARCH.ts
+++ b/packages/search/lib/commands/SEARCH.ts
@@ -73,11 +73,14 @@ export type SearchRawReply = Array<any>;
 export function transformReply(reply: SearchRawReply, withoutDocuments: boolean): SearchReply {
     const documents = [];
     let i = 1;
+
     while (i < reply.length) {
         documents.push({
-            id: reply[i++],
-            value: withoutDocuments ? Object.create(null) : documentValue(reply[i++])
+            id: reply[i],
+            value: withoutDocuments || reply[i+1] == null ? Object.create(null) : documentValue(reply[i+1])
         });
+
+        i += 2;
     }
 
     return {


CaptainTux:
is there any update on this? 👀

samyakkkk:
any update on this?

nkaradzhov:
Closing this one as #2772 was fixed in #2988

Issue #2522, opened 2023-06-01
Maximum call stack size exceeded when calling mset with an array bigger than 103,000


omer727:
### Description

When passing an array to  mset  bigger  than 103000  I get 

```
const arr = new Array(103000).fill('1');
cacheRepo.multiSetObject(arr); 
```
will throw a `Maximum call stack size exceeded` exception on 
```
if (Array.isArray(toSet)) {
        args.push(...toSet.flat());
    }
```

### Node.js Version

v18.14.2

### Redis Server Version

aws redis

### Node Redis Version

1.5.7

### Platform

macOS

higoramp:
@omer727 Same problem here!
I've just pushed an PR that fix this issue

PR #2817, opened 2024-08-14
fix: #2522 max call stack size exceeded on MSET function

higoramp:
Passing an array with several items to mset will throw a Maximum call stack size exceeded exception on args.push


higoramp:
@sjpotter Could you please review the PR?

sjpotter:
this isn't a bug in node-redis, it could be a feature to add, but its not a bug.  The issue is that redis-cluster isn't designed to operate behind a NAT.  io-redis seems to have some support for mapping "cluster IP/ports" to a different ip/port.

sjpotter:
for those that got notified, I accidentally replied to a different issue to this one.

nkaradzhov:
Closing this as the issue seems to be fixed in version 5

PR #2835, opened 2024-09-16
createClient url+tls invariant violation check

dearlordylord:
### Description

on url `rediss:..` + tls: false passed or on url: `redis:` + rls: true passed, the lib blows up during connection with a non-descriptive error

I explicitly check for this invariant violation during initialisation

- parseOptions was added to make it testable and to avoid the pesky `this.#selectedDB = options.database;` in `#initiateOptions` I dare not to touch
- parseOptions is supposed to check cross-field invariants that parseUrl alone won't be able to do
- parseUrl explicitly states on type-level that it affects `tls` (it did this anyways, implicitly, by blowing up the connection on wrong tls passed)
- parseOptions mutates its arguments as well as returns the mutated value in accord with how `#initiateOptions` was implemented (let me know if you'd like me to not-mutate the argument; I'd actually prefer it this way)
- the technical change is that initialization will now set `tls` explicitly to boolean if it wasn't set but an `url` was passed: reasons are both technical (I needed to test without changing your interfaces much) and logical: `tls` undefined with URL: `redis[s]:` seems to be an incorrect invariant anyways

Options:

- not throw any errors, silently overriding user-given value - I prefer to fail-fast, because the instance will fail anyways on connection

---

### Checklist

### DIFF
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -157,20 +157,44 @@ export default class RedisClient<
         return new (RedisClient.extend(options))(options);
     }
 
-    static parseURL(url: string): RedisClientOptions {
+    static parseOptions<O extends RedisClientOptions>(options: O): O {
+        if (options?.url) {
+            const parsed = RedisClient.parseURL(options.url);
+            if (options.socket) {
+                if (options.socket.tls !== undefined && options.socket.tls !== parsed.socket.tls) {
+                    throw new TypeError(`tls socket option is set to ${options.socket.tls} which is mismatch with protocol or the URL ${options.url} passed`)
+                }
+                parsed.socket = Object.assign(options.socket, parsed.socket);
+            }
+
+            Object.assign(options, parsed);
+        }
+        return options;
+    }
+
+    static parseURL(url: string): RedisClientOptions & {
+        socket: Exclude<RedisClientOptions['socket'], undefined> & {
+            tls: boolean
+        }
+    } {
         // https://www.iana.org/assignments/uri-schemes/prov/redis
         const { hostname, port, protocol, username, password, pathname } = new URL(url),
-            parsed: RedisClientOptions = {
+            parsed: RedisClientOptions & {
+              socket: Exclude<RedisClientOptions['socket'], undefined> & {
+                tls: boolean
+              }
+            } = {
                 socket: {
-                    host: hostname
+                    host: hostname,
+                    tls: false
                 }
             };
 
-        if (protocol === 'rediss:') {
-            (parsed.socket as RedisTlsSocketOptions).tls = true;
-        } else if (protocol !== 'redis:') {
+        if (protocol !== 'redis:' && protocol !== 'rediss:') {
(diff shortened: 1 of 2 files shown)


packages/client/lib/client/index.ts
```diff
@@ -157,21 +157,45 @@ export default class RedisClient<
...
+                parsed.socket = Object.assign(options.socket, parsed.socket);
+            }
+
+            Object.assign(options, parsed);
+        }
+        return options;
+    }
+
+    static parseURL(url: string): RedisClientOptions & {
+        socket: Exclude<RedisClientOptions['socket'], undefined> & {
```
dearlordylord, line 176: union is a bit wordy, please guide me where you'd like to define an interface if you think it's not ok to inline that

packages/client/lib/client/index.ts
```diff
@@ -157,21 +157,45 @@ export default class RedisClient<
...
             };
 
-        if (protocol === 'rediss:') {
-            (parsed.socket as RedisTlsSocketOptions).tls = true;
-        } else if (protocol !== 'redis:') {
+        if (protocol !== 'redis:' && protocol !== 'rediss:') {
             throw new TypeError('Invalid protocol');
         }
 
+        parsed.socket.tls = protocol === 'rediss:';
```
dearlordylord, line 197: "DRY enough": at this point `protocol` is a literal union and won't accept anything except those two literals

packages/client/lib/client/index.ts
```diff
@@ -239,19 +263,14 @@ export default class RedisClient<
...
-
-            Object.assign(options, parsed);
-        }
-
         if (options?.database) {
             this.#selectedDB = options.database;
         }
 
+        if (options) {
+            return RedisClient.parseOptions(options);
```
dearlordylord, line 271: maybe better to put before `this.#selectedDB = options.database` but it's a private method use in a constructor anyways?

nkaradzhov:
@dearlordylord sorry for very late reply!  There have been some changes in the ownership of this repository. If you are willing to continue, we can pick this up.

dearlordylord:
@nkaradzhov no worries, thanks for checking this. I updated the branch.

PavelPashov (approved):
LGTM

packages/client/lib/client/index.ts
```diff
@@ -464,15 +488,6 @@ export default class RedisClient<
       };
     }
 
-    if (options?.url) {
-      const parsed = RedisClient.parseURL(options.url);
-      if (options.socket) {
-        parsed.socket = Object.assign(options.socket, parsed.socket);
-      }
-
-      Object.assign(options, parsed);
```
ildaruz, line 473: By removing this lines nothing will fullfull `options?.database` from url anymore, as result db selector will work if only it was set as option, but not as url, it is braking change for url setup connection
`redis[s]://[[username][:password]@][host][:port][/db-number]:`

ildaruz: https://github.com/redis/node-redis/issues/3051

PR #2849, opened 2024-10-07
feat: add credentialProvider option when creating clients

### LINKED ISSUE #821: Handle enabling/changing redis password without application restart

dhensby:
### Description

In some instances, credentials for the redis client will be short-lived and need to be fetched on-demand when connecting to redis. This is the case when connecting in AWS using IAM authentication or Entra ID in Azure.

This feature allows for a credentialProvider to be provided which is a callable function returning a Promise that resolves to a username/password object.

Closes #821

A potential enhancement to this feature could be to also return a TTL with the credentials and then automatically attempt to re-auth the client before the credentials expire; however I've assumed that logic will be the responsibility of the consumer rather than the library. Worst case scenario is that the socket disconnects due to authentication error and then re-authenticates on a future reconnection retry attempt.

Something else that is missing is an explicit decision on how errors from the credentialProvider should be surfaced.

---

### Checklist


zachary-blackbird (approved):

packages/client/lib/client/index.spec.ts
```diff
@@ -103,6 +104,21 @@ describe('Client', () => {
...
+
+    testUtils.testWithClient('should accept a credentialSupplier', async client => {
+      assert.equal(
+          await client.ping(),
+          'PONG'
+      );
+    }, {
+      ...GLOBAL.SERVERS.PASSWORD,
+      clientOptions: {
+        // simulate a slight pause to fetch the credentials
```
leibale, line 116: Does "simulating a pause" make the test better in any way? can't we just `Promise.resolve` directly?

leibale (commented):
Overall looks good, we might have a different name for it

cc @bobymicroby

packages/client/lib/client/index.spec.ts
dhensby: I found that the order of the resolution of promises could cause a problem if it was just an instantly resolving promise, that edge case wouldn't be covered.

bobymicroby:
Hi @dhensby, thank you very much for the PR! We (Redis) are currently working on enabling not only async credential providers, but credential providers that also support rolling credentials/credential invalidation. We need them to enable EntraID (former Azure Active Directory) and more 3rd party identity providers in the future. We have a [PR](https://github.com/redis/node-redis/pull/2877) in the works that already covers async credential providers (check the first [commit](https://github.com/redis/node-redis/pull/2877/commits/a0c324b9a0e9e11f82256748b93a0f97816f482b)). It is similar to what you already have. We will be very happy to get your feedback.

dhensby:
@bobymicroby great, that's exactly what I'm after (EntraID support), so glad to see that's in the works.

I'll close this as superseded by #2877

PR #2858, opened 2024-10-21
Set FIRST_KEY_INDEX if a script accepts keys

poiuj:
### Description

Setting FIRST_KEY_INDEX ensures we send the script to the right node, instead of sending to random node and repeating on getting MOVED response. Relying on a random node send increases traffic and exceeds maxCommandRedirections configuration with non-zero probability.

### Checklist


poiuj:
@leibale do you have specific concerns about this change?

leibale:
1. sorry about the long delay
2. IIRC, NUMBER_OF_KEYS is used by the cluster client to extract the key, I don't think FIRST_KEY_INEDX is needed
3. this API does not exist anymore, we are using https://github.com/redis/node-redis/blob/master/packages/client/lib/client/parser.ts instead

poiuj:
> IIRC, NUMBER_OF_KEYS is used by the cluster client to extract the key, I don't think FIRST_KEY_INEDX is needed
That's not what I observed, that's why I propose the change. But maybe it has changed since then.

> sorry about the long delay
IIRC, NUMBER_OF_KEYS is used by the cluster client to extract the key, I don't think FIRST_KEY_INEDX is needed
this API does not exist anymore, we are using https://github.com/redis/node-redis/blob/master/packages/client/lib/client/parser.ts instead
Not sure I understand. Do you mean we no longer use `defineScript`? This PR is not relevant anymore?

leibale:
Transform arguments uses the `BasicCommandParser` class instead of an array, and instead of using "FIRST_KEY_INDEX", we use `pushKey`.

There are a few reasons for this change:
1. in some commands they first key position is dynamic
2. for client-side caching we need to know all the keys, not just the first one

poiuj:
It makes sense. I updated the branch and verified that the issue is gone with the new approach.

PR #2864, opened 2024-11-08
DOC-4345 added testable JSON index/query example

andy-stark-redis:
### Description

[DOC-4345](https://redislabs.atlassian.net/browse/DOC-4345)

The other client library pages have an JSON index/query example so we should have one for Node.js. Also, these examples are currently just ordinary text in the client pages but I am adding them to the client repos to make them testable.

Adds the example code in the emb-examples branch. I'll add the corresponding text to the doc page when this PR is merged.

---

### Checklist

[DOC-4345]: https://redislabs.atlassian.net/browse/DOC-4345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doctests/cmds-hash.js
+++ b/doctests/cmds-hash.js
@@ -55,7 +55,7 @@ const res8 = await client.hGet('myhash', 'field1')
 console.log(res8) // foo
 
 const res9 = await client.hGet('myhash', 'field2')
-console.log(res9) // foo
+console.log(res9) // null
 
 // REMOVE_START
 assert.equal(res7, 1);


doctests/home-json.js
```diff
@@ -0,0 +1,149 @@
...
+// STEP_END
+
+// STEP_START connect
+const client = await createClient();
+await client.connect();
+// STEP_END
+// REMOVE_START
+try {
+    await client.ft.dropIndex('idx:users', { DD: true });
+} catch{}
```
leibale, line 21: ?!

doctests/home-json.js
```diff
@@ -0,0 +1,149 @@
...
+
+await client.del('user:1', 'user:2', 'user:3');
```
leibale, line 23: `FT.DROPINDEX idx:users DD` already delete the keys, why are we deleting them again?

doctests/home-json.js
```diff
@@ -0,0 +1,149 @@
...
+// REMOVE_END
+
+// STEP_START create_data
+var user1 = {
+    name: "Paul John",
```
leibale, line 28: use `'` instead of `"`
use `const`/`let` instead of `var`
add missing `;`

doctests/home-json.js
```diff
@@ -0,0 +1,149 @@
...
+        AS: 'age'
+    }
+}, {
+    ON: 'JSON',
+    PREFIX: 'user:'
+});
+// STEP_END
+
+// STEP_START add_data
+var user1Added = await client.json.set('user:1', '$', user1);
```
leibale, line 70: there is no reason to make 3 round trips for 3 SET, you can do:
```javascript
const [user1Reply, user2Reply, user3Reply] = await Promise.all([
  client.json.set('user:1', '$', user1),
  client.json.set('user:2', '$', user2),
  client.json.set('user:3', '$', user3)
]);
```

doctests/home-json.js
```diff
@@ -0,0 +1,149 @@
...
+    console.log(cityDoc.value);
+});
+// >>> {city: 'London'}
+// >>> {city: 'Tel Aviv'}
+// >>> {city: 'Tel Aviv'}
+// STEP_END
+// REMOVE_START
+assert.strictEqual(3, citiesResult.total);
+
+citiesResult.documents.sort((a, b)=> a.id < b.id);
```
leibale, line 115: why are we sorting on the client side and not using RediSearch it self?!

doctests/home-json.js
andy-stark-redis: The reason for this is that dropIndex generally throws an error if you try to drop an index that doesn't exist (at least in the clients for other languages). It seemed to work OK without the try...catch locally, but I got an error from the CI when I made the PR. I assumed the try...catch was required and it fixed the error. If I'm doing this the wrong way or misinterpreting the build error then please let me know.

doctests/home-json.js
andy-stark-redis: Often, there isn't an existing index (hence the try...catch as I mentioned in the previous reply), so presumably the keys don't get deleted in this case. However, keys like user:1, etc, are common in docs and test code. If user:1 exists but isn't a JSON key then you'll get the "wrong kind of value" error. If you think I'm being overly cautious here then I'll remove the client.del here, but I have seen this kind of thing happen occasionally.

doctests/home-json.js
andy-stark-redis: This is just for the test code. For other client examples I've written, the order of the documents returned is not deterministic, so you need to sort before using the asserts, otherwise you get an occasionally flaky test. The reason I don't add the sort option to the command is simply that the other client examples don't use the sort option here - just keeping them all consistent.

doctests/home-json.js
andy-stark-redis: Found/replaced these throughout the file now.

doctests/home-json.js
andy-stark-redis: Good tip, thanks.

PR #2870, opened 2024-11-25
Fix creation of cluster client again 

soccermax:
### Description

In the following PR https://github.com/redis/node-redis/pull/2783 we fixed the issue https://github.com/redis/node-redis/issues/2721. That fix got removed from master and therefore also in the current releases. Could you please include this fix again and release a new version? Thank you very much 👍  

---

### Checklist


soccermax:
FYI @leibale

leibale:
Sorry about it.. I'm not sure what happened, but I'll need to make sure that we did not lose more work..

soccermax:
Thanks for taking care 👍

rlindner81:
@leibale I worked on this with @soccermax. We had an issue recently and we realized that the our reconnecting handler is not called even when the reconnectStrategy is used and reconnect attempts are happening.

From local debugging, it looks to me like the socket is emitting both "error" and "reconnecting"
packages/client/lib/client/socket.ts:164
```js
                const retryIn = this.#shouldReconnect(retries++, err as Error);
                if (typeof retryIn !== 'number') {
                    throw retryIn;
                }

                this.emit('error', err);
                await promiseTimeout(retryIn);
                this.emit('reconnecting');
```

These events are forwarded to the client
packages/client/lib/client/index.ts:346
```js
            .on('error', err => {
                this.emit('error', err);
                if (this.#socket.isOpen && !this.#options?.disableOfflineQueue) {
                    this.#queue.flushWaitingForReply(err);
                } else {
                    this.#queue.flushAll(err);
                }
            })
//...
            .on('reconnecting', () => this.emit('reconnecting'))
```

but only the error is forwarded to the cluster interface
packages/client/lib/cluster/cluster-slots.ts:331
```js
    async #createClient(
        node: ShardNode<M, F, S>,
        readonly = node.readonly
    ) {
        const client = new this.#Client(
            this.#clientOptionsDefaults({
                socket: this.#getNodeAddress(node.address) ?? {
                    host: node.host,
                    port: node.port
                },
                readonly
            })
        );
        client.on('error', err => this.#emit('error', err));

        await client.connect();

        return client;
    }
```

Maybe I'm misunderstanding the client vs cluster usage intentions. I would have assumed that it makes sense to forward either event.

Our code is something like 
```js
    if (...) { //env has cluster credentials
       client = redis.createCluster(...)
    } else if () { // env has client credentials
       client = redis.createClient(...)
    } else { ... }
    client.on("error", (err) => { ... });
    client.on("reconnecting", () => { ... });
```

Hope you can help us understand this better. Sorry for the verbosity, but this is most easily explained directly with code :-)

PR #2907, opened 2025-03-11
[CAE-686] Added hash field expiration commands

htemelski:
### Description

Added support for hSetEx, hGetEx and hGetDel commands

---

### Checklist

- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?>

### DIFF
--- a/packages/client/lib/commands/HSETEX.ts
+++ b/packages/client/lib/commands/HSETEX.ts
@@ -89,9 +89,7 @@ function pushTuples(parser: CommandParser, tuples: HSETEXTuples): void {
     }
 
     parser.push((tmpParser.redisArgs.length/2).toString())
-    tmpParser.redisArgs.forEach(arg => {
-        parser.push(arg)
-    });
+    parser.push(...tmpParser.redisArgs)
 }
 
 function _pushTuples(parser: CommandParser, tuples: HSETEXTuples): void {
--- a/packages/client/lib/commands/HSETEX.ts
+++ b/packages/client/lib/commands/HSETEX.ts
@@ -19,16 +19,12 @@ type HSETEXMap = Map<HashTypes, HashTypes>;
 
 type HSETEXTuples = Array<[HashTypes, HashTypes]> | Array<HashTypes>;
 
-type SingleFieldArguments = [field: HashTypes, value: HashTypes];
-
-type MultipleFieldsArguments = [value: HSETEXObject | HSETEXMap | HSETEXTuples];
-
 export default {
   parseCommand(
     parser: CommandParser,
     key: RedisArgument,
-    options?: HSetExOptions,
-    ...[value, fieldValue]: SingleFieldArguments | MultipleFieldsArguments
+    fields: HSETEXObject | HSETEXMap | HSETEXTuples,
+    options?: HSetExOptions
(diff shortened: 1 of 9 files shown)


packages/client/lib/commands/HSETEX.ts
```diff
@@ -0,0 +1,124 @@
...
+        throw Error('passed undefined or null object')
+      } else if (value instanceof Map) {
+        pushMap(parser, value);
+      } else if (Array.isArray(value)) {
+        pushTuples(parser, value);
+      } else {
+        pushObject(parser, value);
+      }
+  },
+  transformReply: undefined as unknown as () => NumberReply
```
bobymicroby, line 69: We can improve the typing to 
`NumberReply<0 | 1>` ,  check the reply schema for [more](https://github.com/redis/redis/blob/b704179f158feef916cca9ba1907a1d0e817886c/src/commands/hsetex.json#L37)

packages/client/lib/commands/HSETEX.ts
```diff
@@ -0,0 +1,124 @@
...
+        } else {
+          parser.push(
+            options.expiration.type,
+            options.expiration.value.toString()
+          );
+        }
+    }
+
+    parser.push('FIELDS')
+    if (typeof value === 'string' || typeof value === 'number' || value instanceof Buffer) {
```
bobymicroby, line 53: I saw that this code block is pretty much the same as the one [here](https://github.com/redis/node-redis/blob/08f664a76e61da0c1532f47ad5d328b3f434c6ec/packages/client/lib/commands/HSET.ts#L25) 
Maybe we can extract it into a common place and reuse it?

bobymicroby (commented):
looks good

packages/client/lib/commands/HSETEX.ts
htemelski: It looks similar to the implementation in `HSET`, but there are a couple of key differences.

1. I've added an `undefined/null` check
2. Inside the `push...()` functions there's additional logic to push the number of fields
HSETEX requires `FIELDS numfields field value [field value]....` [spec](https://github.com/redis/redis/blob/b704179f158feef916cca9ba1907a1d0e817886c/src/commands/hsetex.json#L105-L112)
HSET doesn't accept the `FIELDS numfields` part of the command [spec](https://github.com/redis/redis/blob/b704179f158feef916cca9ba1907a1d0e817886c/src/commands/hset.json#L47)

packages/client/lib/commands/HSETEX.ts
```diff
@@ -0,0 +1,122 @@
...
+
+type SingleFieldArguments = [field: HashTypes, value: HashTypes];
+
+type MultipleFieldsArguments = [value: HSETEXObject | HSETEXMap | HSETEXTuples];
+
+export default {
+  parseCommand(
+    parser: CommandParser,
+    key: RedisArgument,
+    options?: HSetExOptions,
```
bobymicroby, line 30: @htemelski The options argument should be last to be consistent with all other commands that take options.

bobymicroby (changes requested):

bobymicroby (approved):
lgtm

PR #2911, opened 2025-03-21
[CAE-827] Changed ft create vector types to union, added support for int8/uint8

htemelski:
### Description

Changed the vector type field from string to union with all supported versions (including int8 and uint8)

---

### Checklist


bobymicroby (approved):
lgtm

PR #2912, opened 2025-03-21
[CAE-342] Initial sentinel issues fix

### LINKED ISSUE #1: Include design considerations in README
Why would I want to use this instead of fictorial's lib?  What differences are there (philosophical, primarily; I'd assume that the API differs)?

htemelski:
### Description

Fixed a couple of issues with the sentinel implementation

1. Indefinite reconnects without taking into consideration `maxCommandRediscovers`
2. Client losing it's `watch` status if there are multiple topology changes during a transaction

Fixed an issue with the testing setup itself, using the new docker image and setting the `requirepass` and `masterauth` configuration settings in a persistent way.

Enabled the majority of sentinel tests, added new test for fix #1.
Still skipping some tests that are flaky and prone to failing 

---

### Checklist

### DIFF
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -325,6 +325,14 @@ export default class RedisClient<
     return this._self.#watchEpoch !== undefined;
   }
 
+  get watchEpoch() {
+    return this._self.#watchEpoch
+  }
+
+  setWatchEpoch(watchEpoch: number | undefined) {
+    this._self.#watchEpoch = watchEpoch
+  }
+
   setDirtyWatch(msg: string) {
     this._self.#dirtyWatch = msg;
   }
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -325,12 +325,8 @@ export default class RedisClient<
     return this._self.#watchEpoch !== undefined;
   }
 
-  get watchEpoch() {
-    return this._self.#watchEpoch
-  }
-
-  setWatchEpoch(watchEpoch: number | undefined) {
-    this._self.#watchEpoch = watchEpoch
+  get isDirtyWatch() {
+    return this._self.#dirtyWatch !== undefined
   }
 
   setDirtyWatch(msg: string) {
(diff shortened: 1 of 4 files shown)


nkaradzhov (approved):

packages/test-utils/lib/dockers.ts
```diff
@@ -291,3 +299,91 @@ after(() => {
...
+
+    const sentinelPromises: Array<Promise<RedisServerDocker>> = [];
+    const sentinelCount = 3;
+    
+    for (let i = 0; i < sentinelCount; i++) {
+      sentinelPromises.push((async () => {
+        const sentinelArgs: Array<string> = ["sh", "-c"];
+        
+        let sentinelConfig = `
+port {port}
```
bobymicroby, line 359: Hi @htemelski 👋  We should use our docker [container](https://github.com/redis-developer/cae-client-testing/tree/main/docker/client-libs-test-container#pre-configured-redis-instances-sentinel-mode) for spawning sentinel instances if we want to be able to switch/skip tests easily between server versions.

packages/test-utils/lib/dockers.ts
```diff
@@ -51,21 +50,30 @@ export interface RedisServerDocker {
 async function spawnRedisServerDocker({
   image,
   version
-}: RedisServerDockerConfig, serverArguments: Array<string>): Promise<RedisServerDocker> {
+}: RedisServerDockerConfig, serverArguments: Array<string>, dockerEnv?: Map<string, string>): Promise<RedisServerDocker> {
```
bobymicroby, line 53: Instead of adding dockerEnv, dockerBinds, and so on,  please add them as optional properties to the `RedisServerDockerConfig` like : 

export interface VolumeMount {
  source: string;      // Host path
  target: string;      // Container path
  readonly?: boolean;  // Read-only mount
}

export interface RedisServerDockerConfig {
  image: string;
  version: string;
  env?: Record<string, string>;
  volumes?: VolumeMount[];
}

bobymicroby (changes requested):

packages/test-utils/lib/dockers.ts
htemelski: Do you have a strong preference if it's a volume or a mount?

bobymicroby (approved):

PR #2920, opened 2025-03-28
fix: loosen @azure/identity constraint for @redis/entraid

dhensby:
### Description

This loosens the dependency constraint applied to `@azure/identity` so that versions other than 4.7.0 only can be used.

---

### Checklist


bobymicroby (approved):
lgtm

PR #2926, opened 2025-04-23
feat: added commandTimeout option

bronikabrisbon:
### Description

 Some calls to redis took too long in our app by this feature this calls can be aborted.
The PR adds a commandTimeout to the `RedisClientOptions`. If the commandTimeout is set in the createClient, then a timeout is set in the `_executeCommand` function. 
The redisClient is used in the RateLimiterRedis in our app. In the case of a timeout the RateLimiterRedis insurance is supposed to be used as fallback. Therefore a `Promise.race` does not work.

---

### Checklist

### DIFF
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -34,6 +34,7 @@ describe('Client', () => {
         },
         username: 'user',
         password: 'secret',
+        commandTimeout: 1000,
         database: 0,
         credentialsProvider: {
           type: 'async-credentials-provider',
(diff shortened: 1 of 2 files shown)


nkaradzhov:
Hi @bronikabrisbon, sorry for the late reply. Thanks for your efforts! There is [another PR](https://github.com/redis/node-redis/pull/2981) thats doing the same as this one. We already have some traction there and I think it will turn out a bit more elegant to use the abortSignal, so I think we can close this one.

bronikabrisbon:
closing Draft due to [another PR](https://github.com/redis/node-redis/pull/2981).

PR #2946, opened 2025-05-07
fix `cluster.sUnsubscribe` - make `listener` optional

Clsan:
### Description

- I don't see any good reason for cluster.client.sUnsubscribe function must hasve listener.
- So, I changed cluster.client.sUnsubscribe definition. (Make listener optional)

- When sUnsubscribe is called, client.unsubscribe is called. [#](https://github.com/redis/node-redis/blob/master/packages/client/lib/cluster/cluster-slots.ts#L608)
- It finally goes to pubSub.unsubscribe. [#](https://github.com/redis/node-redis/blob/master/packages/client/lib/client/pub-sub.ts#L209)
- If we register listener on sUnsubscribe, pubSubClient's isActive is true due to pre-registered listener. [#](https://github.com/redis/node-redis/blob/master/packages/client/lib/client/pub-sub.ts#L258)
- So, command is not actually issued. [#](https://github.com/redis/node-redis/blob/master/packages/client/lib/client/pub-sub.ts#L263)
- And, also not disconnected. [#](https://github.com/redis/node-redis/blob/master/packages/client/lib/cluster/cluster-slots.ts#L616)
- See #2641

---

### Checklist


nkaradzhov (approved):
LGTM

PR #2854, opened 2024-10-15
Client Side Caching

sjpotter:
### Description

> Describe your pull request here

---

### Checklist

- [ ] Does `npm test` pass with this change (including linting)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

PR #2947, opened 2025-05-07
Client Side Caching

bobymicroby:
- rebased #2854 on top of master
- added docs, fixed typos
- reworked metrics collection


packages/client/lib/client/cache.ts
```diff
@@ -0,0 +1,472 @@
...
+    // "2"
+    let cacheEntry = this.get(cacheKey);
+    if (cacheEntry) {
+      // If instanceof is "too slow", can add a "type" and then use an "as" cast to call proper getters.
+      if (cacheEntry instanceof ClientSideCacheEntryValue) { // "2b1"
+        this.#cacheHit();
+
+        return structuredClone(cacheEntry.value);
+      } else if (cacheEntry instanceof ClientSideCacheEntryPromise) { // 2b2
+        // unsure if this should be considered a cache hit, a miss, or neither?
```
htemelski, line 172: I think that we should count the number of times we get in this case. Ideally, we should be able to tally the total number of cache requests
Not sure what the best approach would be:
1. Count as both cacheHit and cacheInFlight
2. Count only as cacheInFlight

And following this train of thought, should we count the `unknown cache entry type` too?

bobymicroby: If we really want to improve the stats and make them really meaningful, we should only have 3 metrics:

hitRate(): returns the **ratio** of hits to requests
evictionCount(): the number of cache evictions
averageLoadPenalty(): the average time spent loading new values

like in one of the best oss [cache](https://github.com/ben-manes/caffeine/wiki/Statistics) out there:



-- 

Personally, I think we should count ClientSideCacheEntryPromise as a miss.  

For unknown types, with the current design, throwing an error makes sense to me. But maybe we should make the type of what can be stored in the cache a sealed union, not an interface, and make it private. 

I will update the PR later to improve the metrics collection. 

Thanks for the catch, I thought I fixed that already when I last reviewed the PR...

bobymicroby: @htemelski @nkaradzhov I refactored the cache statistics and we are now correctly marking entries being loaded as misses and their load penalties. You can have a look here at this [commit](https://github.com/redis/node-redis/pull/2947/commits/7fd768694a967152069bde8e6568873c309ee4f0)

htemelski (approved):

nkaradzhov (approved):

elena-kolevska:
Hey @andy-stark-redis, we have client-side caching in node-redis now. Bobi included some useful docs you can use.

PR #2949, opened 2025-05-07
fix(client): add type annotations

### LINKED ISSUE #2933: 5.0.0 HGETALL returns type doesn't make sens in TS
## Since the 5.0.0 I'm having typescript error when using HGETALL

**In the documentation :** 
```
Replies will be transformed into useful data structures:

await client.hGetAll("key"); // { field1: 'value1', field2: 'value2' }
await client.hVals("key"); // ['value1', 'value2']
```

**In VS Code:** 
```
const data = await this.redisClient.HGETALL(group);
return Object.entries(data).map((entry) => {
    const value = parseFloat(entry[1]);
                             ^^^^^^^^^
    return { key: entry[0], value: isNaN(value) ? null : value };
});

```

> Argument of type '{ toString: {}; }' is not assignable to parameter of type 'string'.
> (parameter) entry: [string, {
>     toString: {};
> }]

**In 4.7.0, the TS type of the entry was:** 
```(parameter) entry: [string, string]```

### PR DESCRIPTION
**Pull request by nkaradzhov**
Fix type parameter for transformTuplesReply in CONFIG_GET and HGETALL commands

fixes #2933

### Description

---

### Checklist

### DIFF
--- a/packages/client/lib/commands/CONFIG_GET.ts
+++ b/packages/client/lib/commands/CONFIG_GET.ts
@@ -10,7 +10,7 @@ export default {
     parser.pushVariadic(parameters);
   },
   transformReply: {
-    2: transformTuplesReply,
+    2: transformTuplesReply<BlobStringReply>,
     3: undefined as unknown as () => MapReply<BlobStringReply, BlobStringReply>
   }
 } as const satisfies Command;
--- a/packages/client/lib/commands/HGETALL.ts
+++ b/packages/client/lib/commands/HGETALL.ts
@@ -11,7 +11,7 @@ export default {
   },
   TRANSFORM_LEGACY_REPLY: true,
   transformReply: {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by htemelski — APPROVED** [review]

File: packages/client/lib/commands/CONFIG_GET.ts [hunk 1]
```diff
@@ -10,7 +10,7 @@ export default {
     parser.pushVariadic(parameters);
   },
   transformReply: {
-    2: transformTuplesReply,
+    2: transformTuplesReply<BlobStringReply>,
```
htemelski, line 13: Maybe we should check the other commands for similar issues

nkaradzhov: thats twe only 2 places i saw transformTuplesReply being used in this way. everywhere else, the types look to be picking up ok

PR #2952, opened 2025-05-08
Update Redis version to 8.0.1-pre

bobymicroby:
since >=8.0.1 
> FT.AGGREGATE returns an array reply where each row is an array reply
and represents a single aggregate result.
The integer reply at position 1 does not represent a valid value.

We now calculate the result length bazed on the number of results instead of the integer reply at pos 1


nkaradzhov (approved):
LGTM

PR #2955, opened 2025-05-08
revert the 'total' count logic in AGGREGATE response introduced in #2952

bobymicroby:
> FT.AGGREGATE returns an array reply where each row is an array reply and represents a single aggregate result.
The integer reply at position 1 does not represent a valid value.

https://redis.io/docs/latest/commands/ft.aggregate/#return

### DIFF
--- a/packages/search/lib/commands/PROFILE_AGGREGATE.spec.ts
+++ b/packages/search/lib/commands/PROFILE_AGGREGATE.spec.ts
@@ -45,7 +45,9 @@ describe('PROFILE AGGREGATE', () => {
     const res = await client.ft.profileAggregate('index', '*');
 
     const normalizedRes = normalizeObject(res);
-    assert.equal(normalizedRes.results.total, 2);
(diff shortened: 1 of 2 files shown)


packages/search/lib/commands/PROFILE_AGGREGATE.spec.ts
```diff
@@ -45,7 +45,6 @@ describe('PROFILE AGGREGATE', () => {
     const res = await client.ft.profileAggregate('index', '*');
 
     const normalizedRes = normalizeObject(res);
-    assert.equal(normalizedRes.results.total, 2);
```
elena-kolevska, line 48: It may be better to comment this instead with a `TODO` so we can get back to it when it's fixed on the server side.

elena-kolevska (approved):

packages/search/lib/commands/PROFILE_AGGREGATE.spec.ts
bobymicroby: Yeah, makes sense. I also added links to the issue.

PR #2959, opened 2025-05-09
fix(client): bring disableClientInfo option back

### LINKED ISSUE #2958: disableClientInfo option is gone in v5
### Description

The disableClientInfo option is not present in the `RedisClientOptions` interface since v5. Further, v5 client doesnt send the `SETINFO` command at all, regardless of the option. 

Expected behavior is to match v4 which sends `SETINFO` by default, and provide the option to ` disableClientInfo`

### Node.js Version

_No response_

### Redis Server Version

_No response_

### Node Redis Version

_No response_

### Platform

_No response_

### Logs

```bash

```

nkaradzhov:
It disappeared in v5

fixes #2958

---

### Checklist

### DIFF
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -525,7 +525,7 @@ export default class RedisClient<
 
   async #handshake(chainId: symbol, asap: boolean) {
     const promises = [];
-    const commandsWithErrorHandlers = await this.#getHandshakeCommands(this.#selectedDB ?? 0);
+    const commandsWithErrorHandlers = await this.#getHandshakeCommands();
     
     if (asap) commandsWithErrorHandlers.reverse()
 
@@ -542,9 +542,7 @@ export default class RedisClient<
     return promises;
   }
 
-  async #getHandshakeCommands(
-    selectedDB: number
-  ): Promise<
+  async #getHandshakeCommands(): Promise<
     Array<{ cmd: CommandArguments } & { errorHandler?: (err: Error) => void }>
   > {
     const commands = [];
@@ -617,7 +615,7 @@ export default class RedisClient<
       }
     }
 
-    if (selectedDB !== 0) {
+    if (this.#selectedDB !== 0) {
       commands.push({ cmd: ['SELECT', this.#selectedDB.toString()] });
     }
 
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -4,11 +4,11 @@ import { BasicAuth, CredentialsError, CredentialsProvider, StreamingCredentialsP
 import RedisCommandsQueue, { CommandOptions } from './commands-queue';
 import { EventEmitter } from 'node:events';
 import { attachConfig, functionArgumentsPrefix, getTransformReply, scriptArgumentsPrefix } from '../commander';
-import { ClientClosedError, ClientOfflineError, DisconnectsClientError, WatchError } from '../errors';
+import { ClientClosedError, ClientOfflineError, DisconnectsClientError, SimpleError, WatchError } from '../errors';
 import { URL } from 'node:url';
 import { TcpSocketConnectOpts } from 'node:net';
 import { PUBSUB_TYPE, PubSubType, PubSubListener, PubSubTypeListeners, ChannelListeners } from './pub-sub';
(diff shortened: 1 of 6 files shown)


packages/client/lib/client/index.ts
```diff
@@ -538,6 +547,36 @@ export default class RedisClient<
...
+          asap: true
+        }).catch(err => {
+          // Client libraries are expected to ignore failures since they could be
+          // connected to an older server that doesn't support them.
+          if (err !instanceof SimpleError || !err.isUnknownCommand()) {
+            return;
+          }
+        });
+
+        this.#queue.addCommand(['CLIENT', 'SETINFO', 'LIB-VER', version],{
```
bobymicroby, line 568: I can see why you don't add CLIENT SETINFO inside the handshake - because you need a specific catch error clause just for CLIENT SETINFO. However, I think we should add it to the promises array before calling write so we can return it along with the other promises created during the handshake [here](https://github.com/bobymicroby/node-redis/blob/ebd03036d69528d11693711b910222a2d1b61e9d/packages/client/lib/client/index.ts#L634) as CLIENT SETINFO should be threaded as a "handshake" command.

bobymicroby:
> Im not sure if this needs to be done for Sentinel, Cluster as well?

I think Sentinel and Cluster reuse the handshake logic defined in lib/client/index.ts, so you don't have to do something extra.

packages/client/lib/client/index.ts
nkaradzhov: thanks!, addressed

nkaradzhov:
@bobymicroby it looks like we are using [#handshake in one more place - in the `reset` function](https://github.com/redis/node-redis/blob/6f961bd7154c899907cc6419af9ca362c58ec81c/packages/client/lib/client/index.ts#L1136)
Im not sure if we need to add the client setinfo commands there as well?

bobymicroby:
> @bobymicroby it looks like we are using [#handshake in one more place - in the `reset` function](https://github.com/redis/node-redis/blob/6f961bd7154c899907cc6419af9ca362c58ec81c/packages/client/lib/client/index.ts#L1136) Im not sure if we need to add the client setinfo commands there as well?

Why don't we refactor the handshake to separate concerns better? We could nest the current command creation logic in a private `#createHandshakeCommands` method, then have `#handshake` actually queue these commands and return promises for their completion. Then we can reuse it in reset and the socket initiator. Something like : 

```diff
diff --git a/packages/client/lib/client/index.ts b/packages/client/lib/client/index.ts
index 5dae1271ec..c6ed508202 100644
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -410,7 +410,7 @@ export default class RedisClient<
     });
   }
 
-  async #handshake(selectedDB: number) {
+  async #createHandshakeCommands(selectedDB: number) {
     const commands = [];
     const cp = this.#options?.credentialsProvider;
 
@@ -494,10 +494,38 @@ export default class RedisClient<
         parseArgs(COMMANDS.READONLY)
       );
     }
+    
+    
+    if(this.options?.credentialsProvider) {
+      commands.push()
+    }
 
     return commands;
   }
 
+  async #handshake(selectedDB: number) {
+    const commands = await this.#createHandshakeCommands(selectedDB);
+    const promises = [];
+    const chainId = Symbol('Handshake');
+
+    // Preserve the original reverse order iteration
+    for (let i = commands.length - 1; i >= 0; --i) {
+      promises.push(
+        this.#queue.addCommand(commands[i], {
+          chainId,
+          asap: true
+        })
+      );
+    }
+
+    if (promises.length) {
+      this.#write();
+      return Promise.all(promises);
+    }
+    
+    return Promise.resolve();
+  }
+
   #initiateSocket(): RedisSocket {
     const socketInitiator = async () => {
       const promises = [],
@@ -521,15 +549,8 @@ export default class RedisClient<
         );
       }
 
-      const commands = await this.#handshake(this.#selectedDB);
-      for (let i = commands.length - 1; i >= 0; --i) {
-        promises.push(
-          this.#queue.addCommand(commands[i], {
-            chainId,
-            asap: true
-          })
-        );
-      }
+      const handshakePromise = this.#handshake(this.#selectedDB);
+      promises.push(handshakePromise);
 
       if (promises.length) {
         this.#write();
```


What do you think ?

nkaradzhov:
@bobymicroby the option to move the setinfo commands in the `#getHandshakeCommands` method looks to be the most clean

bobymicroby (approved):
LGTM

htemelski (approved):
🚀

PR #2963, opened 2025-05-14
fix(client): cache subsequent clients

### LINKED ISSUE #2954: createClient in v5 is much slower than in v4
### Description

I im using this redis client in an AWS Lambda to write data to an ElastiCache instance. When updating to v5, I noticed a sharp increase in lambda execution time (previous avg 70ms, new avg 850ms).

Image

The function only connects to the server, does a single RPUSH, and closes the connection. 

After downgrading to v4.7.1, the execution time is back down to normal.

### Node.js Version

NodeJS 20

### Redis Server Version

7.1.0

### Node Redis Version

5.0.0

### Platform

AWS Lambda

### Logs

```bash

```

nkaradzhov:
we dont need to recreate a client if
its config hasnt changed

fixes #2954

### Description

---

### Checklist

### DIFF
--- a/packages/client/lib/single-entry-cache.ts
+++ b/packages/client/lib/single-entry-cache.ts
@@ -1,0 +1,25 @@
+export default class SingleEntryCache {
+  #cached?: any;
+  #key?: string;
+
+  /**
+   * Retrieves an instance from the cache based on the provided key object.
+   * 
+   * @param keyObj - The key object to look up in the cache.
+   * @returns The cached instance if found, undefined otherwise.
+   * 
+   * @remarks
+   * This method uses JSON.stringify for comparison, which may not work correctly
+   * if the properties in the key object are rearranged or reordered.
+   */
+  get(keyObj?: object) {
+    return JSON.stringify(keyObj) === this.#key
+      ? this.#cached
+      : undefined;
+  }
+
+  set(keyObj: object | undefined, obj: any) {
+    this.#cached = obj;
+    this.#key = JSON.stringify(keyObj);
+  }
+}
--- a/packages/client/lib/single-entry-cache.ts
+++ b/packages/client/lib/single-entry-cache.ts
@@ -1,19 +1,6 @@
-function makeCircularReplacer() {
-  const seen = new WeakSet();
-  return function serialize(_: string, value: any) {
-    if (value && typeof value === 'object') {
-      if (seen.has(value)) {
-        return 'circular';
-      }
-      seen.add(value);
-      return value;
-    }
-    return value;
-  }
-}
-export default class SingleEntryCache {
-  #cached?: any;
-  #key?: string;
+export default class SingleEntryCache<K, V> {
+  #cached?: V;
+  #serializedKey?: string;
 
   /**
    * Retrieves an instance from the cache based on the provided key object.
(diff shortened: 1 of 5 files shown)


nkaradzhov:
using the following test script we can see reasonable numbers
```js
import { createClient, createClientPool, createCluster } from './packages/client'


const n = 200
const start1 = performance.now();
for (let i = 0; i < n; i++) {
  createClient();
}
console.log(`createClient(${n})`, performance.now() - start1);

const start2 = performance.now();
for (let i = 0; i < n; i++) {
  createClientPool();
}
console.log(`createClientPool(${n})`, performance.now() - start2);

const start3 = performance.now();
for (let i = 0; i < n; i++) {
 createCluster()
}
console.log(`createCluster(${n})`, performance.now() - start3);
```

packages/client/lib/single-entry-cache.ts
```diff
@@ -0,0 +1,25 @@
+export default class SingleEntryCache {
```
bobymicroby, line 14: Hi Niki,

I've been thinking about the `SingleEntryCache` and have a suggestion for simplifying its keying mechanism. Since it's designed to store only a single entry (rather than acting like a multi-key hashmap), we might not need to calculate a comparable representation (like `JSON.stringify`) for the key.

Instead, we could directly store the key (e.g., the config object) and use strict reference equality (`===`) for lookups. This approach could offer a couple of advantages:

*   **Simplicity**: The internal logic of the cache becomes more straightforward.
*   **Performance**: We'd avoid the overhead associated with serializing the key on `get` and `set` operations.

Here's the implementation I have in mind: 
```typescript
export default class SingleEntryCache<K, V> {
  #cached?: V;
  #key?: K;

  /**
   * Retrieves the cached value if the provided key strictly matches (`===`) the stored key.
   *
   * If `K` is an object type, comparison is by reference. If `K` is a primitive
   * type (e.g., string, number), comparison is by value.
   *
   * @template K - The type of the key.
   * @template V - The type of the cached value.
   * @param [key] - The key to look up. If undefined, it will be compared with the stored key.
   * @returns The cached value if the key matches, otherwise `undefined`.
   */
  get(key?: K): V | undefined {
    return key === this.#key
      ? this.#cached
      : undefined;
  }

  /**
   * Sets or updates the cached value and its associated key.
   * Any previously cached entry is overwritten.
   *
   * @template K - The type of the key.
   * @template V - The type of the cached value.
   * @param key - The key to associate with the value.
   * @param value - The value to cache.
   */
  set(key: K, value: V): void {
    this.#cached = value;
    this.#key = key;
  }

  /**
   * Clears the cache, removing the stored key and value.
   * After calling `clear()`, `get()` will return `undefined` until `set()` is called again.
   */
  clear(): void {
    this.#cached = undefined;
    this.#key = undefined;
  }
}
```

The main consideration with this change is that it relies on the key object (our config) having a stable reference or being treated as immutable. If the config object is mutable and its reference changes, or if new, distinct config objects are created with the same logical content, the cache would treat them as different keys. 

This is something we'd need to be mindful of and ensure our usage patterns are compatible, possibly with some testing.

bobymicroby (approved):
LGTM

PR #2974, opened 2025-05-25
[Doc] Fix Node.js console.log output

JCornat:
### Description

This is not much, but reading the Node.js documentation, I saw the console.log output should be 0 instead of an array


JCornat:
When reading the documentation on the website, the others languages return 0

You can see it here : https://redis.io/docs/latest/develop/data-types/json/

Here some screenshots : 

Screenshot from 2025-05-27 16-33-49
Screenshot from 2025-05-27 16-37-10
Screenshot from 2025-05-27 16-36-59


To me, sending the command `await client.json.get("newbike", "$[1].crashes");` returnes the value 0, not `[ 'Deimos', { crashes: 0 }, null ]`

Issue #2175, opened 2022-07-01
is there a way to specify command timeout while creating the node client?


agarwalarjun123:
I could not find a way to add a timeout for a command being executed. can I set it while creating the client or during command execution?

leibale:
We don't have it right now, but it think it'll be nice to have
@guyroyse @simonprickett @chayim WDUT?

guyroyse:
Seems like a straightforward thing to add. Doesn't break backwards compatibility. Sounds good to me.

bmeverett:
I would also be interested in having an option for this.

florian-schunk:
Hi @nkaradzhov,

Thank you for implementing the timeout feature in https://github.com/redis/node-redis/pull/3008!

I noticed the timeout applies only while commands are still pending. Could you help me understand the rationale behind this decision? It's challenging for me to assess whether this covers our use-case effectively, and having more context might greatly aid our evaluation.

Also, if extending the timeout feature beyond pending commands becomes a consideration, I’d be eager to contribute again. Would you be open to discussing a potential follow-up PR? Any insights into challenges or concerns with such an extension would be greatly appreciated.

Thank you!

nkaradzhov:
Hi @florian-schunk,

If timeouts were not restricted to only apply when the command is pending, then we could get in a situation where a command is executed on the server, but is timed out on client side before server returns the result. If the command was non-idempotent, like INCR, LPUSH, SADD, APPEND, XADD.. etc., then we could end up in a broken server state. 

See this example:

```ts
const client = createClient({
  commandOptions: {
    timeout: 1
  },
});

await client.connect();
await client.set('key', 10)
try {
  await client.incrBy('key', 10)
} catch (error) {
    if(error instanceof TimeoutError) {
      // Retry
      await client.incrBy('key', 10)
    }
}

console.log(await client.get('key')) // 20 or 30?
```

Implementation of a true timeout mechanism becomes tricky because then a global setting will not be suitable for both idempotent and non-idempotent commands. The best way to handle this difference is to have per-command setting, that would give the user full flexibility. Such change is not trivial, and while in principal is possible and good to have, Im not sure how it will be prioritized in the near future.

florian-schunk:
Thanks for the explanation! I understand your concern, however I think even for non-idempotent commands an unrestricted timeout makes sense.

In the example that you provided, it would be the conscious choice of the developer to retry in case of the timeout and it can be a valid choice when the consistency isn't critical.
[...]

PR #2981, opened 2025-05-28
feat: Add command timeout

florian-schunk:
### Description

This PR introduces a timeout for commands sent to redis.
Our use-case is that we are using node-rate-limiter-flexible with node-redis, but sometimes the connection to redis is slow. In these cases we want to fall back the the insurance limiter of node-rate-limiter-flexible.

This feature was also requested in https://github.com/redis/node-redis/issues/2175

---

### Checklist

### DIFF
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -283,11 +283,11 @@ describe('Client', () => {
     });
 
   testUtils.testWithClient('CommandTimeoutError', async client => {
-    const promise = assert.rejects(client.sendCommand(['PING']), CommandTimeoutError),
-      start = process.hrtime.bigint();
+    const promise = assert.rejects(client.sendCommand(['PING']), AbortError);
+    const start = process.hrtime.bigint();
 
     while (process.hrtime.bigint() - start < 50_000_000) {
-      // block the event loop for 1ms, to make sure the connection will timeout
+      // block the event loop for 50ms, to make sure the connection will timeout
     }
 
     await promise;
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -263,6 +263,23 @@ describe('Client', () => {
           AbortError
         );
       }, GLOBAL.SERVERS.OPEN);
+
+      testUtils.testWithClient('AbortError with timeout', client => {
+        const controller = new AbortController();
+        controller.abort();
+
+        return assert.rejects(
+          client.sendCommand(['PING'], {
+            abortSignal: controller.signal
+          }),
+          AbortError
+        );
+      }, {
+        ...GLOBAL.SERVERS.OPEN,
+        clientOptions: {
+          commandTimeout: 50,
+        }
+      });
     });
 
   testUtils.testWithClient('CommandTimeoutError', async client => {
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -1,7 +1,7 @@
 import { strict as assert } from 'node:assert';
 import testUtils, { GLOBAL, waitTillBeenCalled } from '../test-utils';
 import RedisClient, { RedisClientOptions, RedisClientType } from '.';
-import { AbortError, ClientClosedError, ClientOfflineError, ConnectionTimeoutError, DisconnectsClientError, ErrorReply, MultiErrorReply, SocketClosedUnexpectedlyError, WatchError } from '../errors';
+import { AbortError, ClientClosedError, ClientOfflineError, ConnectionTimeoutError, CommandTimeoutError, DisconnectsClientError, ErrorReply, MultiErrorReply, SocketClosedUnexpectedlyError, WatchError } from '../errors';
 import { defineScript } from '../lua-script';
 import { spy } from 'sinon';
 import { once } from 'node:events';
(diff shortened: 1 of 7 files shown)


florian-schunk:
Hi @nkaradzhov sorry for pinging you directly, but I couldn't find information on how to contact maintainers for this project.
What should we do for getting this PR reviewed and merged?

nkaradzhov:
@florian-schunk hi, sorry for the delay, been a bit busy. I will take a look at this.

packages/client/lib/client/index.ts
```diff
@@ -889,9 +893,41 @@ export default class RedisClient<
...
+        abortSignal = AbortSignal.any([
+          abortSignal,
+	  options.abortSignal
+	]);
+      }
+      options = {
+        ...options,
+        abortSignal
+      }
+    }
```
nkaradzhov: We can simplify this by using the AbortSignal.timeout static method:

```
    if (this._self.#options?.commandTimeout) {
      let abortSignal = AbortSignal.timeout(
        this._self.#options?.commandTimeout
      );
      if (options?.abortSignal) {
        abortSignal = AbortSignal.any([abortSignal, options.abortSignal]);
      }
      options = {
        ...options,
        abortSignal
      };
    }
```

packages/client/lib/client/index.ts
```diff
@@ -889,9 +893,41 @@ export default class RedisClient<
...
+        reject(new CommandTimeoutError());
+      }, this._self.#options?.commandTimeout)
+      promise.then(result => {
+        clearInterval(timeoutId);
+        resolve(result)
+      }).catch(error => {
+        clearInterval(timeoutId);
+        reject(error)
+      });
+    })
```
nkaradzhov: Then, this change becomes obsolete

packages/client/lib/errors.ts
```diff
@@ -22,6 +22,12 @@ export class SocketTimeoutError extends Error {
   }
 }
 
+export class CommandTimeoutError extends Error {
+  constructor() {
+    super('Command timeout');
+  }
+}
+
```
nkaradzhov: No need to introduce another error, we can use the `AbortError`

packages/client/lib/client/index.spec.ts
```diff
@@ -263,8 +263,41 @@ describe('Client', () => {
...
+    }
+
+    await promise;
+  }, {
+    ...GLOBAL.SERVERS.OPEN,
+    clientOptions: {
+      commandTimeout: 50,
+    }
+  });
+
```
nkaradzhov: This test would need to change like this:

```
  testUtils.testWithClient('CommandTimeoutError', async client => {
    const promise = client.sendCommand(['PING'])
    const start = process.hrtime.bigint();
    while (process.hrtime.bigint() - start < 50_000_000) {
      // block the event loop for 50ms, to make sure the connection will timeout
    }
    assert.rejects(promise, AbortError);
  }, {
    ...GLOBAL.SERVERS.OPEN,
    clientOptions: {
      commandTimeout: 50,
    }
  });
```

packages/client/lib/errors.ts
florian-schunk: Ok, I changed it to use AbortError, but couldn't there be situation where it makes sense to distinguish between a timeout and having the command manually aborted? Wouldn't it then make more sense to have a different error?

packages/client/lib/client/index.ts
florian-schunk: I tried this, also together with the change in the test that you proposed, but somehow for me the test keeps failing with this approach. I am not quite understanding, why.

nkaradzhov:
I will take a look, maybe will make a pr to your pr

nkaradzhov:
@florian-schunk i made a PR to your fork

packages/client/lib/client/index.spec.ts
```diff
@@ -263,8 +263,47 @@ describe('Client', () => {
...
+        clientOptions: {
+          commandTimeout: 50,
+        }
+      });
     });
 
+
+  testUtils.testWithClient('rejects with AbortError on commandTimeout timer', async client => {
+    const start = process.hrtime.bigint();
+    const promise = client.ping();
```
florian-schunk, line 293: Just for my understanding: What is the difference between  sendCommand(['PING'] and ping()?

nkaradzhov: not much. `sendCommand` is a catch-all command that can send any command. Usually, we would have all the commands available as dedicated methods, like `client.ping()`, `client.hset()` etc.  But sometimes there are commands that are not yet implemented as a dedicated method. In those cases, users can fallback to using `sendCommand`

One significant difference is that `sendCommand` takes additional options. In the options, there is the `abortSignal`. So in practice, anyone who needs to be able to timeout a command can use `sendCommand` with `abortSignal` to achieve this.

nkaradzhov (approved):
CLGTM

nkaradzhov:
@florian-schunk, I just spoke with our architect and he suggested we also test if this works properly with cluster and sentinel. While looking into this, i understood that we are not exposing this parameter in the cluster config and sentinel config. So it will need a bit more work. Let me take this for now and I will make a PR to your fork once im ready. It will take some time as I will not be available next week.

nkaradzhov:
Hi @florian-schunk, it turns out the feature is a little bit more complicated - we also want this to work with sentinel and cluster. There[ is another PR](https://github.com/redis/node-redis/pull/3008) that is gonna take care of it, so this one could be closed. Thanks for being the main driver of this feature, I will see if I can make you a co-author in the other PR.

PR #2990, opened 2025-06-05
fix(client): make unstable cmds throw

### LINKED ISSUE #2989: Unstable core commands dont throw without unstableResp3
### Description

As per [the docs](https://github.com/redis/node-redis/blob/master/docs/v5.md#unstable-module-commands), 

> Some Redis module commands have unstable RESP3 transformations. These commands will throw an error when used with RESP3 unless you explicitly opt in to using them by setting unstableResp3: true in your client configuration

- update docs, remove `module`
- fix commands

### Node.js Version

_No response_

### Redis Server Version

_No response_

### Node Redis Version

_No response_

### Platform

_No response_

### Logs

```bash

```

nkaradzhov:
As per [the docs](https://github.com/redis/node-redis/blob/master/docs/v5.md), unstableResp3 commands should throw when client is created with { RESP: 3, unstableResp3: false|undefined }

fixes #2989

---

### Checklist

### DIFF
--- a/packages/client/lib/commander.ts
+++ b/packages/client/lib/commander.ts
@@ -38,7 +38,11 @@ export function attachConfig<
     Class: any = class extends BaseClass {};
 
   for (const [name, command] of Object.entries(commands)) {
-    Class.prototype[name] = createCommand(command, RESP);
+    if (config?.RESP == 3 && command.unstableResp3 && !config.unstableResp3) {
+      Class.prototype[name] = throwResp3SearchModuleUnstableError;
+    } else {
+      Class.prototype[name] = createCommand(command, RESP);
+    }
   }
 
   if (config?.modules) {
--- a/docs/v5.md
+++ b/docs/v5.md
@@ -42,9 +42,9 @@ RESP3 uses a different mechanism for handling Pub/Sub messages. Instead of modif
 
 ## Known Limitations
 
-### Unstable Module Commands
+### Unstable Commands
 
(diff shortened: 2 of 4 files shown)


packages/client/lib/commander.ts
```diff
@@ -38,7 +38,11 @@ export function attachConfig<
     Class: any = class extends BaseClass {};
 
   for (const [name, command] of Object.entries(commands)) {
-    Class.prototype[name] = createCommand(command, RESP);
+    if (config?.RESP == 3 && command.unstableResp3 && !config.unstableResp3) {
```
bobymicroby, line 41: @nkaradzhov I just remembered that I saw [this](https://github.com/bobymicroby/node-redis/blob/b33a662e501fcea56aa27d7e90ff604b9f965183/packages/client/lib/commander.ts#L49) piece of code which was supposed to attach a command that throws an error. Maybe we can fix that code instead.

nkaradzhov: @bobymicroby [this is working properly](https://github.com/bobymicroby/node-redis/blob/b33a662e501fcea56aa27d7e90ff604b9f965183/packages/client/lib/commander.ts#L49). The problem is that it is executed only for module commands. And we have XREAD and XREADGROUP as part of the main client. So we should add the same logic in the place we attach the core functions as well

bobymicroby (approved):
lgtm

packages/client/lib/commander.ts
bobymicroby: got it

PR #3006, opened 2025-06-25
fix(commands): sPopCount return Array<string>

### LINKED ISSUE #3004: [v5]spop and sPopCount types are incorrect
### Description

Hi, I think the type returned by functions `spop` and `spopCount` are invalid. Actually the type return : 

```
Promise<string | null>
```

But this functions return an array of string. The return type should be `Promise<string[]>` instead.

### Node.js Version

22.16.0

### Redis Server Version

_No response_

### Node Redis Version

5.5.6

### Platform

macOS

### Logs

```bash

```

nkaradzhov:
fix sPopCount return type - should be an array instead of string | null

Also, touch the tests for sPop and sPopCount to use the new parseCommand API

fixes #3004

### Checklist

### DIFF
--- a/packages/client/lib/commands/SPOP_COUNT.ts
+++ b/packages/client/lib/commands/SPOP_COUNT.ts
@@ -1,5 +1,5 @@
 import { CommandParser } from '../client/parser';
-import { RedisArgument, BlobStringReply, NullReply, Command } from '../RESP/types';
+import { RedisArgument, Command, ArrayReply } from '../RESP/types';
 
 export default {
   IS_READ_ONLY: false,
@@ -16,5 +16,5 @@ export default {
     parser.pushKey(key);
     parser.push(count.toString());
   },
-  transformReply: undefined as unknown as () => BlobStringReply | NullReply
+  transformReply: undefined as unknown as () => ArrayReply<string>
 } as const satisfies Command;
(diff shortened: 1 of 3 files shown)


packages/client/lib/commands/SPOP_COUNT.ts
```diff
@@ -16,5 +16,5 @@ export default {
     parser.pushKey(key);
     parser.push(count.toString());
   },
-  transformReply: undefined as unknown as () => BlobStringReply | NullReply
+  transformReply: undefined as unknown as () => ArrayReply<string>
```
htemelski-oss, line 19: Isn't the proper return type `ArrayReply<string> | NullReply` ?
Based on the [docs](https://redis.io/docs/latest/commands/spop/) we can get Nil/Null if the key doesn't exist

bobymicroby: according to docs ( with count  parameter ) :

One of the following:
[Nil reply](https://redis.io/docs/latest/develop/reference/protocol-spec#bulk-strings): if the key does not exist.
[Array reply](https://redis.io/docs/latest/develop/reference/protocol-spec#arrays): when called with the count argument, a list of the removed members.

I agree with Hristo  about the return type : `ArrayReply<string> | NullReply`

nkaradzhov: resolved offline, if count is provided, the server always returns an array

PR #3019, opened 2025-07-17
Add Redis transparent proxy test utilities

bobymicroby:
The RedisProxy component is a transparent TCP proxy designed for Redis integration testing that sits between Redis clients and servers, enabling sophisticated test scenarios through arbitrary RESP response injection. 

The proxy maintains full protocol compatibility while providing the ability to intercept all bidirectional traffic, monitor connection events, and most importantly, inject custom RESP messages directly to client connections. This allows developers to simulate complex Redis behaviors like cluster redirections (`MOVED`/`ASK` responses), RESP3 push notifications, error conditions, and server-side events that would be difficult to reproduce with a real Redis server.

### DIFF
--- a/packages/client/lib/test-utils.ts
+++ b/packages/client/lib/test-utils.ts
@@ -93,6 +93,12 @@ export const GLOBAL = {
         password: 'password'
       }
     },
+    OPEN_RESP_3: {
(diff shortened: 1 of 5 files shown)


nkaradzhov (approved):
CLGTM

PR #3041, opened 2025-07-30
feat: add EPSILON parameter support to VSIM command

PavelPashov:
### Description

add EPSILON parameter support to VSIM command

---

### Checklist

### DIFF
--- a/packages/client/lib/commands/VSIM.ts
+++ b/packages/client/lib/commands/VSIM.ts
@@ -4,6 +4,7 @@ import { transformDoubleArgument } from './generic-transformers';
 
 export interface VSimOptions {
   COUNT?: number;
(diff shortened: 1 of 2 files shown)


packages/client/lib/commands/VSIM.ts
```diff
@@ -4,6 +4,7 @@ import { transformDoubleArgument } from './generic-transformers';
 
 export interface VSimOptions {
   COUNT?: number;
+  EPSILON?: number;
```
htemelski-redis, line 7: Wouldn't that be a breaking change in the function signature?

htemelski-redis (changes requested):

htemelski-redis (approved):

PR #3045, opened 2025-08-05
fix(commands): expire, expireAt are not readonly

### LINKED ISSUE #3044: getting MOVED errors on expire command
### Description

we have 3 shards elasticache cluster (3 maters + a replica for each master). I occasionally/rarely getting MOVED errors on my expire commands. The cluster is stable and no slot redistribution is happening at all but yet I am getting this. I dug a bit into the code and I noticed you flag the expire command as readonly (see image below).

So my guess (still havent dug deeper into it). is it is routing the command to the correct shard but to the replica and getting a MOVED (since the replica probably cant update the expiry) and it then reforwards to the replica 16 times in a row exceeding the maxRedirection. And rarely cause it has to pick the replica 16 times in a row ie the probability is 1 in 2^16 or one in 65,000

The main question is why am I getting the MOVED error and how to resolve it

A side question would be why is expires marked as read only as it clearly updates the records metadata

Image

### Node.js Version

20.18

### Redis Server Version

8.1

### Node Redis Version

5.7

### Platform

linux

### Logs

```bash

```

nkaradzhov:
fixes #3044

`expire` and `expireAt` commands are not readonly because they modify the server's state.

### DIFF
--- a/packages/client/lib/commands/EXPIREAT.ts
+++ b/packages/client/lib/commands/EXPIREAT.ts
@@ -3,7 +3,6 @@ import { RedisArgument, NumberReply, Command } from '../RESP/types';
 import { transformEXAT } from './generic-transformers';
 
 export default {
-  IS_READ_ONLY: true,
   /**
    * Sets the expiration for a key at a specific Unix timestamp
    * @param parser - The Redis command parser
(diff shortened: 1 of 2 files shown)


packages/client/lib/commands/EXPIREAT.ts
```diff
@@ -3,7 +3,6 @@ import { RedisArgument, NumberReply, Command } from '../RESP/types';
 import { transformEXAT } from './generic-transformers';
 
 export default {
-  IS_READ_ONLY: true,
```
maa105, line 6: There is also EXPIRETIME command no?

nkaradzhov: [EXPIRETIME is actually readonly](https://redis.io/docs/latest/commands/expiretime/)

PR #3053, opened 2025-08-12
Clustering sendCommand docs

nfriedly:
### Description

We noticed that `sendCommand()` takes different arguments for clusters vs clients, and I wanted to document the differences. I think I got it correct, but please review closely just to be sure.

It might also be worth adding a note to [the readme](https://github.com/redis/node-redis/blob/2f106324507eec905b8fe7691ba11179acdeeca7/README.md#L136-L144) also, since this is a somewhat unexpected difference, what do you think?

Relates to https://github.com/express-rate-limit/rate-limit-redis/issues/207 & https://github.com/express-rate-limit/rate-limit-redis/pull/208

---

### Checklist

- [ ] ~~Is the new or changed code fully tested?~~ (N/A)

### DIFF
--- a/docs/clustering.md
+++ b/docs/clustering.md
@@ -38,6 +38,25 @@ await cluster.close();
 | scripts                |         | Script definitions (see [Lua Scripts](./programmability.md#lua-scripts))                                                                                                                                                                                                                                                    |
 | functions              |         | Function definitions (see [Functions](./programmability.md#functions))                                                                                                                                                                                                                                                      |
 
+## Usage
+
+Most redis commands are the same as with individual clients.
+
+### Unsupported Redis Commands
+
+If you want to run commands and/or use arguments that Node Redis doesn't know about (yet!) use `.sendCommand()`.
+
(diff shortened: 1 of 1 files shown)


docs/clustering.md
```diff
@@ -38,6 +38,25 @@ await cluster.close();
 | scripts                |         | Script definitions (see [Lua Scripts](./programmability.md#lua-scripts))                                                                                                                                                                                                                                                    |
 | functions              |         | Function definitions (see [Functions](./programmability.md#functions))                                                                                                                                                                                                                                                      |
 
+## Usage
+
+Most redis commands are the same as with individual clients.
```
nfriedly, line 43: I *hope* this is true. Are there other differences that should be mentioned here?

nkaradzhov: Correct

nkaradzhov (approved):
Thanks @nfriedly, your comments are spot on. Its nice to have this explained

PR #3068, opened 2025-08-25
WIP: unrestricted timeouts

florian-schunk:
### Description

The purpose of this PR is to remove the restriction that command timeouts only apply while the command is still waiting to be sent, so that a slow answer from redis also triggers the timeout.

However testing this code, I noticed some discrepancies between running the tests on my local machine and the pipeline, so I'm opening this PR now, before the development is fully finished to see how the pipeline processes the tests so far.

### Checklist

### DIFF
--- a/packages/client/lib/client/commands-queue.ts
+++ b/packages/client/lib/client/commands-queue.ts
@@ -27,10 +27,6 @@ export interface CommandToWrite extends CommandWaitingForReply {
     signal: AbortSignal;
     listener: () => unknown;
   } | undefined;
-  timeout: {
-    signal: AbortSignal;
-    listener: () => unknown;
-  } | undefined;
 }
 
 interface CommandWaitingForReply {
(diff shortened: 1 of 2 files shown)


nkaradzhov:
Hi @florian-schunk, apologies for the late reply. Im still loaded on other work and Im not sure when i will be able to take more thorough look at this. 

Here is a scenario where this approach breaks: You send two commands (cmdA, cmdB). cmdA expires, server then replies for cmdA:

```
  +--------+             +--------+
  | Client |             | Redis  |
  +--------+             +--------+
       |                      |
       |------ cmdA --------->|
       |                      |
       |------ cmdB --------->|
       |                      |
       |↺ cmdA expired        |
       |                      |
       |<---- cmdA reply -----|
       |                      |
```

When we expire a command, we just remove it from the queue. That means we will have only cmdB remaining in the queue when we receive the response for cmdA. 

And look at how we handle a response:
```ts
  #onReply(reply: ReplyUnion) {
    this.#waitingForReply.shift()!.resolve(reply);
  }
```

This means we will wrongly think the response for cmdA is the response for cmdB


Im not sure how many other corner cases will pop out, but we need to very carefully consider all posibilities as this has the potential to break stuff in a very bad way. ( hence why I was reluctant to work on this :D )

florian-schunk:
Yes, I also realized that this was an issue, so I decided to instead work on a timeout [here](https://github.com/animir/node-rate-limiter-flexible/pull/335) instead.
I tried to solve this here as well, didn't quite get it to work. I pushed my work in progress now, as a reference if anyone wants to tackle this in the future.

PR #3074, opened 2025-09-06
Fix client kill enums

### LINKED ISSUE #2805: Command parameter enums not exported
### Description

I am trying to invoke the `clientKill()` command on a redis client object. This command takes a parameter object containing a field of type `enum ClientKillFilters`. But this enum is not exported by the `redis` or `@redis/client` packages, which means there is no way to supply a valid parameter object for this command.

The same is true for other commands that take an enum parameter, of which there are a handful. A quick grep shows:
- COMMAND_LIST
- CLIENT_KILL
- CLUSTER_FAILOVER
- CLUSTER_NODES
- CLUSTER_SETSLOTS
- FLUSHALL

I have found a workaround by importing the enum from the `'@redis/client/dist/lib/commands/CLIENT_KILL'` package, but obviously that is not a sustainable solution. The `@redis/client` package should either export those enums, or replace the enums altogether with one-of-several-strings parameters.

### Node.js Version

v20.14.0

### Redis Server Version

_No response_

### Node Redis Version

4.6.14

### Platform

_No response_

### Logs

_No response_

watersRand:
This pull request resolves the issue where certain command parameter enums, specifically ClientKillFilters for the CLIENT KILL command, were not publicly exported by the @redis/client package. The purpose of this change is to make the ClientKillFilters enum accessible to all package consumers, eliminating the need for users to rely on an unstable internal import path.

This matters because it provides a reliable and supported way for developers to use commands that require these enums, improving the package's usability and API stability.

Closes #2805


nkaradzhov:
Hi @watersRand, sorry for the late response and thanks for the contribution! This looks good, but we can do a little better here by exporting those to the top level so users can `import { CLIENT_KILL_FILTERS } from 'redis'`.[ I made a PR](https://github.com/watersRand/node-redis1/pull/1) to your branch, so you can just take the changes and we can progress from there

PR #3076, opened 2025-09-09
Emit invalidate events from client

sastcsgh:
### Description

This PR adds an option to the client (`emitInvalidate`) that allows the client to emit `invalidate` events when it receives invalidation messages from the Redis server. This is similar to what the Client Side Cache does, except it doesn't require the cache to be enabled. This is useful when you want to perform certain transformations on the data from Redis and cache the result of these manually, while still being able to automatically receive invalidation events for the data you pull from the server.

---

### Checklist

### DIFF
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -774,9 +774,8 @@ export default class RedisClient<
       commands.push({cmd: ['CLIENT', 'TRACKING', 'ON']});
     }
     
-    const { tls, host } = this.#options!.socket as RedisTcpSocketOptions;
-    const maintenanceHandshakeCmd = await EnterpriseMaintenanceManager.getHandshakeCommand(!!tls, host!, this.#options!);
     const maintenanceHandshakeCmd = await EnterpriseMaintenanceManager.getHandshakeCommand(this.#options);
+    
     if(maintenanceHandshakeCmd) {
       commands.push(maintenanceHandshakeCmd);
     };


nkaradzhov:
Hi @sastcsgh, thanks for the PR, much appreciated! Im so sorry, we just merged a PR that touches on how you interact with the commands-queue. Those changes are in a bit of a conflict with your changes. It would be great if you fix the conflicts, then I can review more thoroughly.

Just to give you some hints:
The new way we hook to push events on the commands-queue is via [addPushHandler](https://github.com/redis/node-redis/blob/e2702b63f2f7317f1745216ceebe3d1686c17dd0/packages/client/lib/client/commands-queue.ts#L185), also take a look at the [PushHandler definition](https://github.com/redis/node-redis/blob/e2702b63f2f7317f1745216ceebe3d1686c17dd0/packages/client/lib/client/commands-queue.ts#L55). Keep in mind that currently, client-side caching is fully consuming the `invalidate` push message by returning `true` ( maybe you can make it return false, then return true from your new handler )

sastcsgh:
No worries @nkaradzhov, the new pushHandler logic looks much more robust than the old solution anyway. 
I re-wrote the logic for `emitInvalidate` in the new push handler style. (Like before, it's basically the same logic as for the cache, except it emits the invalidate event instead of calling the `invalidate` method of the `clientSideCache`).
I left the push handler for the cache as it was, since the cache itself emits the invalidate event anyway, so when the cache *is* used, we can listen for the event there.

nkaradzhov:
@sastcsgh So sorry, i tried to fix the merge conflicts via the web interface, but actually made things worse... Can you please remove my last commit from your branch and fix the merge conflicts so we can run the tests again

sastcsgh:
As far as I can tell, the only remaining issue was the changed `maintenanceHandshakeCmd` logic. Should be fixed now!

nkaradzhov:
@sastcsgh thank you, this looks good to go. Just want to have @bobymicroby give one final blessing

bobymicroby (approved):
lgtm

PR #3081, opened 2025-09-18
perf(ts)!: move multi() away from RedisClient

nkaradzhov:
RedisClientMultiCommandType is slow to typecheck because of its recursive nature and the fact that it has many ( around 700 ) methods. Most of the users dont use multi(), so there is no point to incur the ts costs to them. One possible solution is to move the multi() method away from the main RedisClient, thus typechecking will not apply to RedisClientMultiCommandType.

This is a breaking change.

Usage before:

```ts
import { createClient } from 'redis';
cont result = await client.multi()
  .set('foo', 3)
  .get('bar')
  .execTyped();
```

Usage after:

```ts
import { createClient, multi } from 'redis';
cont result = await multi(client)
  .set('foo', 3)
  .get('bar')
  .execTyped();
```

### Description

---

### Checklist

### DIFF
--- a/packages/client/index.ts
+++ b/packages/client/index.ts
@@ -12,7 +12,9 @@ export { VerbatimString } from './lib/RESP/verbatim-string';
 export { defineScript } from './lib/lua-script';
 export * from './lib/errors';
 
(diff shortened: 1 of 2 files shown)


nkaradzhov:
closing in favor of https://github.com/redis/node-redis/pull/3084

Issue #1855, opened 2022-01-20
Redis cluster not emitting events


danish241194:
```
 const Redis  = require("redis");

  const cluster = Redis.createCluster({
      rootNodes: [
        {
          url: 'redis://localhost:7000',
          password: 'myredis'
        }]
    });

    cluster.on('ready', () => console.log('Redis Cluster ready'));

   cluster.on('connect', () => console.log('Redis Cluster connect'));

    cluster.on("reconnecting", () => console.log('Redis Cluster reconnecting '));
    cluster.on('error', (err : any) => console.log('Redis Cluster Error', err));
    cluster.connect();
```
With the above code, it connects to redis server perfectly and if i set key, value that also works fine but i can see 'ready' , 'connect' ,'reconnecting' events are not getting trigerred . 

If redis client is connected to redis server and i stop redis server i can see 'error' event getting trigerred mutiple times without emitting reconencting event.

Is there any other way to get ready , connect and reconnecting event when dealing with cluster ? 
**Environment:**
 - "node": ">=12.22.7"
 - Redis Server Version: redis_version:6.2.5
 - node "redis": "^4.0.0",
 - Platform: mac os 10.15.7 , also tried on Amazon Linux 2

leibale:
Currently the cluster only emit errors, wanna share the use case of those events in a cluster?

danish241194:
I want to raise some kind of metric on reconnecting and and ready event. Also i more thing when i terminated my server i can see the cluster.on('error', (err : any) => console.log('Redis Cluster Error', err)); is not catching the connection error issue , it causes app to crash.

mapsam:
👍 I have a potentially very specific use-case for wanting cluster "ready" events:

Cluster events like "ready" would be helpful when managing connections within a long-running application. If we can listen to "ready" and "error" events, we can effectively flip between using the cluster client or not. For example:

```js
const cluster = createCluster();
const isConnected = false;

cluster.on('ready', () => {
  isConnected = true;
});

async function get(id) => {
  if (isConnected) {
    const val = await cluster.get(id);
    return val;
  } else {
     // fallback to alternative data source
     // and try to reconnect redis in the background
     cluster.connect();
     const val = await getFromSomethingElse(id);
     return val;
  }
}
```

This allows an application to avoid constantly attempting to connect to redis before falling back to the alternative datasource with a `try/catch` block, which can lead to latency and unnecessary network I/O when the cluster is out of commission.

[...]

PR #3083, opened 2025-09-18
cluster/node events (#1855)

NaughtySora:
### Description
Adding events for cluster client, some people were interested in it.
I have read the issue but didn't see any concrete approved suggestions so i come up with some.

for cluster:
1. connect - when cluster ready.
2. disconnect - when cluster is closed.
3. error (already exists).

for nodes:
 1. node-connect - when tcp socket connected.
 2. node-ready - when tcp socket are ready.
 3. node-reconnecting - when tcp socket is reconnecting.
 4. node-disconnect - when tcp socket disconnected.
 5. node-error - when tcp socket errored.
 
Node events provide host/port to identify a particular node.

I would like to hear suggestions or opinion if this feature needed.

issue #1855 

### Checklist

I tried to run tests via npm run build and npm test but it fails with timeout "before hook" after 30000ms after 1 hour of going a couple of time. But git checks are fine i guess.
 
I haven't seen any tests for emitting events, i have tested it only locally.

Can't find where i can update desctiption about cluster events. 
I saw small section of client events on redis website.
Can't update that.

### DIFF
--- a/packages/client/lib/cluster/index.spec.ts
+++ b/packages/client/lib/cluster/index.spec.ts
@@ -343,12 +343,6 @@ describe('Cluster', () => {
   describe('clusterEvents', () => {
     testUtils.testWithCluster('should fire events', async (cluster) => {
       const log: string[] = [];
-      const numberOfMasters = 2;
-      const nodeConnect = numberOfMasters;
-      const nodeReady = nodeConnect + numberOfMasters;
-      const connect = nodeReady + 1;
-      const nodeDisconnect = connect + numberOfMasters;
-      const disconnect = nodeDisconnect + 1;
 
       cluster
         .on('connect', () => log.push('connect'))
@@ -363,33 +357,16 @@ describe('Cluster', () => {
       await cluster.connect();
       cluster.destroy();
 
-      assert.equal(log.length, disconnect);
-
-      assert.deepEqual(
-        log.slice(0, nodeConnect),
-        new Array(numberOfMasters).fill('node-connect'),
-      );
-      assert.deepEqual(
-        log.slice(nodeConnect, nodeReady),
-        new Array(numberOfMasters).fill('node-ready'),
-      );
-      assert.deepEqual(
-        log.slice(nodeReady, connect),
-        new Array(1).fill('connect'),
-      );
-      assert.deepEqual(
-        log.slice(connect, nodeDisconnect),
-        new Array(numberOfMasters).fill('node-disconnect'),
-      );
-      assert.deepEqual(
-        log.slice(nodeDisconnect, disconnect),
-        new Array(1).fill('disconnect'),
-      );
-
-      assert.equal(log.includes('error'), false);
-      assert.equal(log.includes('node-error'), false);
-      assert.equal(log.includes('node-reconnecting'), false);
-
+      assert.deepEqual(log, [
+        'node-connect',
+        'node-connect',
+        'node-ready',
+        'node-ready',
+        'connect',
+        'node-disconnect',
+        'node-disconnect',
+        'disconnect',
+      ]);
     }, {
       ...GLOBAL.CLUSTERS.OPEN,
       disableClusterSetup: true,


nkaradzhov:
@NaughtySora hi, sorry for big delay! This looks good! Could I ask you to:
1. Add some tests [here](https://github.com/redis/node-redis/blob/master/packages/client/lib/cluster/index.spec.ts) that verify the events are properly fired? I have made a [PR to your branch with a template of the test](https://github.com/NaughtySora/node-redis/pull/1)
2. Document the events in a new section called `Events` [here](https://github.com/redis/node-redis/blob/master/docs/clustering.md). You can take inspiration from [here](https://github.com/redis/node-redis/blob/master/README.md#events)

NaughtySora:
Hi, thank you for the reply @nkaradzhov . I've added the test example from your PR, and I will proceed with the steps above .

NaughtySora:
I added documentation inspired by standalone client.
Test is also included but I have 2 concerns.
1. I found only one way so far to get cluster's amount of nodes for testing. I'm looking for it in cluster._options, it can easily break encapsulation and cause troubles in the future. 
2. I have tried to add negative routes for tests but faced the fact I have no idea how to force cluster node to reconnect in test framework. So the events 'error', 'node-error' and 'node-reconnecting' haven't tested yet.

nkaradzhov:
> I added documentation inspired by standalone client. Test is also included but I have 2 concerns.
> 
> 1. I found only one way so far to get cluster's amount of nodes for testing. I'm looking for it in cluster._options, it can easily break encapsulation and cause troubles in the future.
> 2. I have tried to add negative routes for tests but faced the fact I have no idea how to force cluster node to reconnect in test framework. So the events 'error', 'node-error' and 'node-reconnecting' haven't tested yet.

1. Take a look [here](https://github.com/redis/node-redis/blob/73413e086c26c33cfa0575a00c27ef367e31aa77/packages/client/lib/test-utils.ts#L127) - you can specify how many masters/replicas you want your cluster. Essentially: 
```ts
  testUtils.testWithCluster('foo', async cluster => {
    //tests
  }, {
    serverArguments: [],
    numberOfMasters: 3,
    numberOfReplicas: 3
  });

```
2. I think its fine to leave those untested for now

NaughtySora:
I've found the way to make test more safe and changed docs a bit for more clear description.

NaughtySora:
Hi, facing weird test behavior. 
Locally I rise 6 standalone redis instances with 
redis-server redis.conf with ports 7000 - 7005, and after cluster via
redis-cli --cluster create 127.0.0.1:7000 127.0.0.1:7001 127.0.0.1:7002 127.0.0.1:7003 127.0.0.1:7004 127.0.0.1:7005 --cluster-replicas 1
the same test only changing master count to 3 instead of 2. Everything works properly.
I see tests error and realize that it fails on the first assert when 2 !== 8, so the actual event count in log array is 2, i suspect its 'connect' and 'disconnect' so the node-* events aren't fired or fired before/after the assert.
I was wondering maybe I don't know how test framework works?
Can you please give me any advice ?

nkaradzhov:
@NaughtySora it looks like we are using an optimization for testing, which tries to minimize connections if there are no commands yet. I made a PR to your repo that fixes the test.

packages/client/lib/cluster/index.spec.ts
```diff
@@ -339,4 +339,63 @@ describe('Cluster', () => {
...
+        new Array(numberOfMasters).fill('node-disconnect'),
+      );
+      assert.deepEqual(
+        log.slice(nodeDisconnect, disconnect),
+        new Array(1).fill('disconnect'),
+      );
+
+      assert.equal(log.includes('error'), false);
+      assert.equal(log.includes('node-error'), false);
+      assert.equal(log.includes('node-reconnecting'), false);
```
nkaradzhov: I think we can simplify this by just asserting the log is equal to the expected log. For example:

```ts
assert.deepEqual(log, [
  'node-connect',
  'node-connect',
  'node-connect',
  'node-ready',
  'node-ready',
  'node-ready',
  'connect',
  'node-disconnect',
  'node-disconnect',
  'node-disconnect',
  'disconnect'
]
```

NaughtySora: Total agree with it. I will make a change.
If I get it correctly, if we have 2 masters it will be 2 event for each node-* event. 
So the test should be 
[
        'node-connect',
        'node-connect',
        'node-ready',
        'node-ready',
        'connect',
        'node-disconnect',
        'node-disconnect',
        'disconnect',
]
Correct me please if i'm wrong.

nkaradzhov:
@NaughtySora good work! Thanks again for the contribution!

NaughtySora:
Thank you!

Issue #3062, opened 2025-08-19
Unhandled promise rejection triggered in DoublyLinkedList after executeTask()


udnes99:
### Description

We just detected an unhandled promise rejection that was triggered while issuing a command or a `.execute()` using a  pool instance. Due to limited stack trace, I am not sure if it was a dedicated client obtained by `.execute()` or just a client retrieved for the execution of a single command. I am not that familiar with the internals of the Redis SDK, but I am guessing some kind of race condition within the library related to the acquisition and release of client instances for tasks, or some bug in the linked list.

### Node.js Version

22.18.0

### Redis Server Version

7.4.0

### Node Redis Version

5.8.0

### Platform

Linux

### Logs

```bash
TypeError: Cannot set properties of undefined (setting 'next')
  File "/app/node_modules/.pnpm/@redis+client@5.8.0/node_modules/@redis/client/dist/lib/client/linked-list.js", line 76, col 32, in DoublyLinkedList.remove
    node.previous.next = node.next;
  File "/app/node_modules/.pnpm/@redis+client@5.8.0/node_modules/@redis/client/dist/lib/client/pool.js", line 262, col 28, in #returnClient
    this.#clientsInUse.remove(node);
  File "/app/node_modules/.pnpm/@redis+client@5.8.0/node_modules/@redis/client/dist/lib/client/pool.js", line 248, col 50
    .finally(() => this.#returnClient(node));
  File "<anonymous>"
```

nkaradzhov:
Hi @udnes99, thanks for raising this issue. I can easily patch the current TypeError that you are experiencing, but i have the suspicion that there is another problem that is hiding here. And that problem is gonna be hard to find tbh. If you are able to narrow it down somehow on your side, that would be great.

udnes99:
> Hi [@udnes99](https://github.com/udnes99), thanks for raising this issue. I can easily patch the current TypeError that you are experiencing, but i have the suspicion that there is another problem that is hiding here. And that problem is gonna be hard to find tbh. If you are able to narrow it down somehow on your side, that would be great.

I understand! Like I said, I am not that familiar with the internals of this library. I am not sure what the conditions are for triggering it; it seems sporadic on our side. I suspect there is some sort of race condition. I can try to look through the code, though.

michal-billtech:
I get the same error consistently for the following code:

```typescript
//repro.ts
import { createClientPool } from "redis";

const client = createClientPool({ url: "redis://bad-host:6379" });
[...]

PR #3085, opened 2025-09-21
Fix: resolve doubly linked list push issue

ntvviktor:
### Description

- Fixed initialization problem in DoublyLinkedList.push() for an empty doubly linked list
- Resolves issue with client pool management
- An attempt to fix this issue https://github.com/redis/node-redis/issues/3062.

Now the client pool properly shut down after getting a bad connection.
---

### Checklist

- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? Just a Bug fix.

### DIFF
--- a/packages/client/lib/client/linked-list.spec.ts
+++ b/packages/client/lib/client/linked-list.spec.ts
@@ -6,7 +6,7 @@ import {
 import { equal, deepEqual } from "assert/strict";
 
 describe("DoublyLinkedList", () => {
-  const list = new DoublyLinkedList();
+  const list = new DoublyLinkedList<number>();
 
   it("should start empty", () => {
     equal(list.length, 0);
@@ -95,6 +95,38 @@ describe("DoublyLinkedList", () => {
       count++;
     }
     equal(count, 6);
+  });
+
+  it("should handle remove on empty list", () => {
+    list.reset();
+    const node = list.push(1);
+    list.remove(node);
+    equal(list.length, 0);
+    deepEqual(Array.from(list), []);
+    list.remove(node);
+    equal(list.length, 0);
+    deepEqual(Array.from(list), []);
+  });
+
+
+  it("should safely remove nodes while iterating", () => {
+    list.reset();
+    list.push(1);
+    list.push(2);
+    list.push(3);
+    list.push(4);
+    list.push(5);
+    
+    const visited: number[] = [];
+    for (const node of list.nodes()) {
+      visited.push(node.value);
+      if (node.value % 2 === 0) {
+        list.remove(node);
+      }
+    }
+    deepEqual(visited, [1, 2, 3, 4, 5]);
+    equal(list.length, 3);
+    deepEqual(Array.from(list), [1, 3, 5]);
   });
 });
 
--- a/packages/client/lib/client/linked-list.ts
+++ b/packages/client/lib/client/linked-list.ts
@@ -29,11 +29,16 @@ export class DoublyLinkedList<T> {
     ++this.#length;
 
     if (this.#tail === undefined) {
-      return this.#tail = this.#head = {
+      this.#head = {
+        previous: undefined,
+        next: this.#tail,
+        value
+      }
+      return this.#tail = {
         previous: this.#head,
         next: undefined,
         value
-      };
+      }
     }
 
     return this.#tail = this.#tail.next = {
--- a/packages/client/lib/client/linked-list.ts
+++ b/packages/client/lib/client/linked-list.ts
@@ -83,6 +83,7 @@ export class DoublyLinkedList<T> {
   }
 
   remove(node: DoublyLinkedNode<T>) {
+    if (this.#length === 0) return;
     --this.#length;
 
     if (this.#tail === node) {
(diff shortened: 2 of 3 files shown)


nkaradzhov:
@ntvviktor looks like you broke the tests

ntvviktor:
> @ntvviktor looks like you broke the tests

hey [nkaradzhov](https://github.com/nkaradzhov), I am considering, even though fixing the code logic of the DoublyLinkedList will properly shut down the bad host request for redis connection, but it will break the sentinel client tests. I believe it has been somehow coded to pass with the error of issue:
```shell
TypeError: Cannot set properties of undefined (setting 'next')
  File "/app/node_modules/.pnpm/@redis+client@5.8.0/node_modules/@redis/client/dist/lib/client/linked-list.js", line 76, col 32, in DoublyLinkedList.remove
    node.previous.next = node.next;
```
I will investigate further to try to triage and fix the root cause.

ntvviktor:
In the case of the issue above, seems like the `RedisClientPool` use the Doubly Linked List for a client in use store, so when there is an error such as bad host request as the issue described, the bang mark `node.previous!.next` in the `remove()` function in  DoublyLinkedList lead to the TypeError since it calls `remove()` twice, in the catch, and in the `#returnClient`.
```ts
async #create() {
    const node = this._self.#clientsInUse.push(
      this._self.#clientFactory()
        .on('error', (err: Error) => this.emit('error', err))
    );

    try {
      const client = node.value;
      await client.connect();
    } catch (err) {
      this._self.#clientsInUse.remove(node);
      throw err;
    }

    this._self.#returnClient(node);
  }
```
I fix using the if statement to check for the previous is not enough, and found out a root cause from RedisCommandQueue also using Doubly Linked List. 
It uses a for loop statement over an iterator `*nodes()` function while using `remove()` in the loop, leading to a TimeOut command execution in the Test, so in this PR I also trying to fix that by changing the iterator function `*nodes()` to make sure setting `node.next` to `undefined` at `shift()`, `unshift()`, and `remove()` will be safe.

I ran all tests locally and it currently works as expected. @nkaradzhov  Please help me enable the pipeline to see whether my fix break anything. Thanks!

ntvviktor:
Hi @nkaradzhov, please help me review the code, thanks in advance

nkaradzhov:
@ntvviktor the PR looks to be damaged somehow, it looks like ther are some unwanted changes after a merge?

ntvviktor:
> @ntvviktor the PR looks to be damaged somehow, it looks like ther are some unwanted changes after a merge?

Hi @nkaradzhov thanks for feedback, I have made changes, my bad on the rebase master branch into this.

nkaradzhov:
@ntvviktor sorry for being paranoid here, but could you please provide a test case for the DoublyLinkedList [somewhere here](https://github.com/redis/node-redis/blob/master/packages/client/lib/client/linked-list.spec.ts)
Usually, when we find a bug, we write a test that proves the bug, then do a fix and the test starts passing. We end up having our test suite growing stronger

ntvviktor:
> @ntvviktor sorry for being paranoid here, but could you please provide a test case for the DoublyLinkedList [somewhere here](https://github.com/redis/node-redis/blob/master/packages/client/lib/client/linked-list.spec.ts) Usually, when we find a bug, we write a test that proves the bug, then do a fix and the test starts passing. We end up having our test suite growing stronger

Thanks @nkaradzhov, I have added the according tests.

nkaradzhov:
Looks like this causes #3127

nkaradzhov:
@ntvviktor, sorry, what was the reason for the addition of `#handleSentinelFailure`?

ntvviktor:
Hey @nkaradzhov, it was because during this test case, after a master is shutdown and the test trying to reconnect, the sentinelSentinels or some other utils could use the first item [0] in the list, but it already shutdown, hence the `topology-change` was never reached or emitted, only the master-change was? hence, I made changes by remove the dead node before the #reset function so that it could has the correct event emitted, I personally think the #reset function would bring back the re-connecting for all of the redis sentinels, but perhaps I made a wrong assumption. Let me know what you think
```
it('shutdown sentinel node', async function () {
      this.timeout(60000);
      sentinel = frame.getSentinelClient();
      sentinel.setTracer(tracer);
      sentinel.on("error", () => { });
      await sentinel.connect();
      tracer.push("connected");

      let sentinelChangeResolve;
      const sentinelChangePromise = new Promise((res) => {
        sentinelChangeResolve = res;
      })

      const sentinelNode = sentinel.getSentinelNode();
      tracer.push(`sentinelNode = ${sentinelNode?.port}`)

      sentinel.on('topology-change', (event: RedisSentinelEvent) => {
        tracer.push(`got topology-change event: ${JSON.stringify(event)}`);
        if (event.type === "SENTINEL_CHANGE") {
          tracer.push("got sentinel change event");
          sentinelChangeResolve(event.node);
        }
      });

      tracer.push("Stopping sentinel node");
      await frame.stopSentinel(sentinelNode!.port.toString());
      tracer.push("Stopped sentinel node and waiting on sentinel change promise");
      const newSentinel = await sentinelChangePromise as RedisNode;
      tracer.push("got sentinel change promise");
      assert.notEqual(sentinelNode!.port, newSentinel.port);
    });
```
when the frame trying to get

PR #3097, opened 2025-10-07
chore(build): fix npm run build command

nkaradzhov:
### Description
importing from 'package.json' messes typescript, as it starts thinking it depends on the files it wants to write, resulting in build errors if the dist folders are not deleted. ultimately it was very annoying to delete all dist folders and then build when you just want to build. and its also much slower because you loose all your cache

### Checklist


nkaradzhov:
closing this as it turns out the new dependency on package.json might not be the only problem. needs further investigation

Issue #1955, opened 2022-02-09
Add support for redis command: LATENCY HISTOGRAM


chayim:

nkaradzhov:
Implemented — available as `latencyHistogram` (`LATENCY HISTOGRAM`) in v5, with RESP2/RESP3 reply handling. Closing as done.

PR #3099, opened 2025-10-10
Add latency histogram

NaughtySora:
Adding LATENCY HISTOGRAM command.
I personally will use this in my current and future projects.
#1955 

I have tried to run all the tests locally but facing the same issue with timeout as before.
Tests are present.
I think documentation about commands located in redis official website.


NaughtySora:
Hi, thank you so much for the amends.
I will be more careful next time with REST 2/3 and redis documentation.

NaughtySora:
I'm i right to assume that according to the redis documentation i need also implement RESP 2 transform function with histogram usec tuple?
I'm a bit confused after merging the PR.

nkaradzhov:
@NaughtySora sorry, i didnt quite get what you mean.

NaughtySora:
in transformReply object there are 2 options for RESP 2 and 3.
 RESP 2 has undefined value, and 3 has an implementation according to redis documentation.
 But redis documentation also says that RESP 2 has its own contract.
I thought i need to implement it as well.

PR #3105, opened 2025-10-17
Extend proxy

nkaradzhov:
The purpose for this PR is to extend the proxy to provide couple of new functionalities:

- [ ] request/response
- [x] response transformations
- [ ] resp support

### DIFF
--- a/packages/test-utils/lib/redis-proxy.ts
+++ b/packages/test-utils/lib/redis-proxy.ts
@@ -59,7 +59,7 @@ export class RedisProxy extends EventEmitter {
   public readonly config: Required<ProxyConfig>;
   private readonly connections: Map<string, ActiveConnection>;
   private isRunning: boolean;
-  private interceptorInitializer?: InterceptorInitializer;
+  private interceptorInitializer: InterceptorInitializer = (init) => init;
 
   constructor(config: ProxyConfig) {
     super();
@@ -261,11 +261,6 @@ export class RedisProxy extends EventEmitter {
 
     clientSocket.on('data', async (data) => {
       this.emit('data', connectionId, 'client->server', data);
-
-      if(!this.interceptorInitializer) {
-        serverSocket.write(data);
-        return;
-      }
 
       connectionInfo.inflightRequestsCount++;
 


nkaradzhov:
#jit_bypass_commit

packages/test-utils/lib/redis-proxy.ts
```diff
@@ -243,12 +259,38 @@ export class RedisProxy extends EventEmitter {
       this.emit('connection', connectionInfo);
     });
 
-    clientSocket.on('data', (data) => {
+    clientSocket.on('data', async (data) => {
       this.emit('data', connectionId, 'client->server', data);
-      serverSocket.write(data);
+
+      if(!this.interceptorInitializer) {
```
bobymicroby: I'm wondering if we can eliminate the branch prediction (`if(!this.interceptorInitializer`)) entirely by implementing a default "socket write/noop" interceptor that simply performs the socket write when no initializer is provided.

packages/test-utils/lib/redis-proxy.ts
```diff
@@ -20,6 +20,7 @@ interface ConnectionInfo {
 interface ActiveConnection extends ConnectionInfo {
   readonly clientSocket: net.Socket;
   readonly serverSocket: net.Socket;
+  inflightRequestsCount: number
```
bobymicroby, line 23: What do you think about borrowing from node-redis's cache stats [approach](https://github.com/bobymicroby/node-redis/blob/d6d8d8e8ed98f401d1fb69b97b08a767ee9de166/packages/client/lib/client/cache.ts#L30) ? We could start by tracking inflightRequestsCount initially and add more metrics later. This design integrates well if you want to hook in traces, metrics, or logs.

bobymicroby (approved):
LGTM. I've left some minor improvement suggestions.

packages/test-utils/lib/redis-proxy.ts
nkaradzhov: this is only to distinguish between request/response and push for now. it will go away in nest iteration

packages/test-utils/lib/redis-proxy.ts
nkaradzhov: yes, we can totally do this, in fact it was like that before, but i thought this is better. i will move it back

PR #3110, opened 2025-10-24
add OpenTelemetry metrics instrumentation

PavelPashov:
### Description

TODO 
- [x] Add docs
- [x] Add examples

---

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core request/connection paths (`RedisClient`, `RedisSocket`, queues, cluster redirection) to emit metrics and maintain per-client identity/registry, so mistakes could add overhead or skew counts. Functional behavior should be unchanged, but instrumentation around errors/closures and cache invalidation needs careful review.
> 
> **Overview**
> Adds first-class OpenTelemetry metrics support via a new `OpenTelemetry.init()` entrypoint and an internal metrics/registry layer that can be enabled before client creation.
> 
> Instruments core client operations to emit metrics for command durations (including `MULTI`/`PIPELINE` batches), connection lifecycle (create time, active count, close reasons, relaxed timeouts, handoff), resiliency/errors (including cluster redirections and maintenance notifications), pub/sub message in/out, stream lag from `XREAD`/`XREADGROUP`, and client-side caching hits/misses/evictions.
> 
> Introduces stable per-client identity (`_clientId`) and a `ClientRegistry` so observable gauges and metric attribution can track standalone clients, pools, and cluster nodes; updates pool/cluster/node creation to set roles/parent IDs and unregisters clients on `close`/`destroy`/`quit`. Documentation and an `otel-metrics.js` example are added, along with new OpenTelemetry dev dependencies and extensive tests for the metrics implementation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6b0433169d9dc7cd0f337e89695662cb30b99889. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -481,11 +481,11 @@ export default class RedisClient<
    * Sets the client identity. Used by pool/cluster/sentinel when creating child clients.
    */
   _setIdentity(role: ClientRole, parentId?: string): void {
-    this._self.#clientIdentity = {
-      ...this._self.#clientIdentity,
-      role,
-      parentId
-    };
+    this._self.#clientIdentity.role = role;
+
+    if (parentId) {
+      this._self.#clientIdentity.parentId = parentId;
+    }
   }
 
   /**
--- a/packages/client/lib/client/identity.ts
+++ b/packages/client/lib/client/identity.ts
@@ -29,8 +29,8 @@ export enum ClientRole {
  */
 export interface ClientIdentity {
   readonly id: string;
-  readonly role: ClientRole;
-  readonly parentId?: string;
+  role: ClientRole;
+  parentId?: string;
 }
 
 /**
--- a/packages/client/lib/opentelemetry/client-registry.spec.ts
+++ b/packages/client/lib/opentelemetry/client-registry.spec.ts
@@ -126,7 +126,7 @@ describe("ClientRegistry Unit Tests", () => {
 
   it("should handle unix socket clients", () => {
     ClientRegistry.init();
-    
+
     createClient({
       socket: {
         path: "/tmp/redis.sock",
@@ -147,6 +147,25 @@ describe("ClientRegistry Unit Tests", () => {
     assert.strictEqual(attributes.port, undefined);
     assert.strictEqual(attributes.db, 0);
     assert.strictEqual(attributes.clientId, handle.identity.id);
+  });
+
+  it("should reflect identity updates after _setIdentity", () => {
+    ClientRegistry.init();
+
+    const client = createClient();
+    const originalId = client._clientId;
+
+    client._setIdentity(ClientRole.POOL_MEMBER, "pool-1");
+
+    const handle = ClientRegistry.instance.getById(originalId);
+    assert.ok(handle);
+    assert.strictEqual(handle.identity.id, originalId);
+    assert.strictEqual(handle.identity.role, ClientRole.POOL_MEMBER);
+    assert.strictEqual(handle.identity.parentId, "pool-1");
+
+    const attributes = handle.getAttributes();
+    assert.strictEqual(attributes.clientId, originalId);
+    assert.strictEqual(attributes.parentId, "pool-1");
   });
 });
 


elimelt:
This is an exciting feature, I hope it makes it to main!

packages/client/lib/client/cache.ts
```diff
@@ -551,21 +553,41 @@ export class BasicClientSideCache extends ClientSideCacheProvider {
...
     // "2"
     let cacheEntry = this.get(cacheKey);
+
     if (cacheEntry) {
       // If instanceof is "too slow", can add a "type" and then use an "as" cast to call proper getters.
       if (cacheEntry instanceof ClientSideCacheEntryValue) { // "2b1"
         this.#statsCounter.recordHits(1);
+        OTelMetrics.instance.clientSideCacheMetrics.recordCacheRequest(
+          CSC_RESULT.HIT,
+          client._clientId,
```
vladvildanov, line 563: We don't have this in the design, is it something that we discussed and should do across clients?

packages/client/lib/client/cache.ts
```diff
@@ -551,21 +553,41 @@ export class BasicClientSideCache extends ClientSideCacheProvider {
...
+        );
+        // Estimate bytes saved by avoiding network round-trip
+        // Note: JSON.stringify approximation; actual RESP wire size may differ (especially for Buffers)
+        const bytesEstimate = JSON.stringify(cacheEntry.value).length;
```
vladvildanov: What do you think about wrapping it with `Buffer.byteLength(JSON.stringify(cacheEntry.value), 'utf8');`? If data to be stored isn't just ASCII it makes sense to measure UTF-8 bytes count

packages/client/lib/client/cache.ts
```diff
@@ -551,21 +553,41 @@ export class BasicClientSideCache extends ClientSideCacheProvider {
...
+        OTelMetrics.instance.clientSideCacheMetrics.recordNetworkBytesSaved(
+          bytesEstimate,
+          client._clientId,
+        );
 
         return structuredClone(cacheEntry.value);
       } else if (cacheEntry instanceof ClientSideCacheEntryPromise) { // 2b2
         // This counts as a miss since the value hasn't been fully loaded yet.
         this.#statsCounter.recordMisses(1);
+        OTelMetrics.instance.clientSideCacheMetrics.recordCacheRequest(
```
vladvildanov, line 574: I think it makes sense to move this call inside of the function above `this.#statsCounter.recordMisses(1)` , in this case you'll avoid code repetitions

packages/client/lib/client/index.ts
```diff
@@ -1080,21 +1163,44 @@ export default class RedisClient<
     args: ReadonlyArray<RedisArgument>,
     options?: CommandOptions
   ): Promise<T> {
+    const recordOperation = OTelMetrics.instance.commandMetrics.createRecordOperationDuration(
```
vladvildanov, line 1169: Why does operation duration recording differs from other recorders? Instead of calling recorder function directly you're using a callback factory method?

packages/client/lib/opentelemetry/metrics.ts
```diff
@@ -0,0 +1,967 @@
...
+    this.#options = options;
+    this.#instruments = instruments;
+  }
+
+  public createRecordConnectionCreateTime(clientId?: string): () => void {
+    const startTime = performance.now();
+    const clientAttributes = resolveClientAttributes(clientId);
+
+    return () => {
+      this.#instruments.dbClientConnectionCreateTime.record(
```
vladvildanov, line 192: Nice place for closure!

vladvildanov (changes requested):

packages/client/lib/opentelemetry/noop-metrics.ts
```diff
@@ -0,0 +1,121 @@
...
+  IOTelStreamMetrics,
+  IOTelClientSideCacheMetrics,
+  IOTelResiliencyMetrics,
+  IOTelConnectionAdvancedMetrics,
+  IOTelCommandMetrics,
+  IOTelConnectionBasicMetrics,
+} from "./types";
+import { noopFunction } from "./utils";
+
+export class NoopCommandMetrics implements IOTelCommandMetrics {
```
vladvildanov, line 18: Why do you need this object? Just to ensure that no commands will be actually send to a collector?

packages/client/lib/client/cache.ts
PavelPashov: We have it in the design as `redis.client.csc.requests` unless it's implemented incorrectly here

packages/client/lib/client/cache.ts
PavelPashov: I’d prefer to keep it simpler for now. Since `StatsCounter` is an interface, it could be swapped out for a different implementation, and that would mean we’d need to support this across the various `StatsCounter` types.

packages/client/lib/opentelemetry/noop-metrics.ts
PavelPashov: Because we have multiple metric groups and don’t know what the user will opt into, I kept each group in its own class and defaulted them to noop. When the user initializes metrics, we swap in the real implementations based on config, so all checks happen once during `OpenTelemetry.init(...)`. This keeps runtime code simple (no repeated conditional checks) and ensures disabled metrics have near-zero overhead

packages/client/lib/client/cache.ts
PavelPashov: That's a good catch!

packages/client/lib/client/cache.ts
vladvildanov: @PavelPashov I mean clientId attribute, this is not in the design

packages/client/lib/client/cache.ts
vladvildanov: In this case proxy pattern works the best, every implementation should go through Proxy object that adds metric recording. But I agree, it might be an overkill for now

packages/client/lib/opentelemetry/noop-metrics.ts
vladvildanov: So this is basically Null object pattern where in case if Otel is disabled empty methods are called?

packages/client/lib/client/cache.ts
PavelPashov: `recordCacheRequest` is more of a helper function not the actual instrument. We use the `clientId` to get the client attributes from the `ClientRegistry`, but `clientId` is also the value for `db.client.connection.pool.name`.
image

vladvildanov: Okay

packages/client/lib/opentelemetry/noop-metrics.ts
PavelPashov: Yes, exactly

packages/client/lib/cluster/index.ts
```diff
@@ -459,6 +490,12 @@ export default class RedisCluster<
         }
 
         if (err.message.startsWith('MOVED')) {
+          OTelMetrics.instance.resiliencyMetrics.recordClientErrors(
+            err,
+            true,
+            client._clientId,
+            i
+          );
```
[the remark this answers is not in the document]
PavelPashov: @nkaradzhov I think it's worth discussing this, because there seem to be trade offs to the possible solutions

packages/client/lib/client/socket.ts
```diff
@@ -60,6 +61,7 @@ export default class RedisSocket extends EventEmitter {
   readonly #reconnectStrategy;
   readonly #socketFactory;
   readonly #socketTimeout;
+  readonly #clientId: string;
```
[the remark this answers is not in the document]
PavelPashov: @nkaradzhov we can discuss this as well

packages/client/lib/RESP/types.ts
```diff
@@ -293,6 +293,7 @@ export type Command = {
   TRANSFORM_LEGACY_REPLY?: boolean;
   transformReply: TransformReply | Record<RespVersions, TransformReply>;
   unstableResp3?: boolean;
+  onSuccess?: (args: ReadonlyArray<RedisArgument>, reply: unknown, clientId: string) => void;
```
PavelPashov: @nkaradzhov we might want to rename `onSuccess` to something more metrics/otel related

packages/client/lib/client/index.ts
```diff
@@ -1042,10 +1122,13 @@ export default class RedisClient<
     } else {
       const reply = await fn();
 
-      if (transformReply) {
-        return transformReply(reply, parser.preserve, commandOptions?.typeMapping);
+      const finalReply = transformReply ? transformReply(reply, parser.preserve, commandOptions?.typeMapping) : reply;
+
+      if (command.onSuccess) {
+        command.onSuccess(parser.redisArgs, finalReply, this._self._clientId);
```
PavelPashov: @nkaradzhov there might be a better way to record specific command related metrics that require the server response

packages/client/lib/opentelemetry/metrics.ts
```diff
@@ -0,0 +1,967 @@
...
+
+    // Build the appropriate function based on options
+    if (options.hasIncludeCommands || options.hasExcludeCommands) {
+      // Version with filtering
+      this.createRecordOperationDuration = this.#createWithFiltering.bind(this);
+    } else {
+      this.createRecordOperationDuration =
+        this.#createWithoutFiltering.bind(this);
+    }
+  }
```
nkaradzhov, line 81: why have two fns?

packages/client/lib/client/index.ts
nkaradzhov: Im wondering as well. Is it only to capture the start time?

packages/client/lib/RESP/types.ts
nkaradzhov: lets see if we can remove it altogether and wether its worth it

packages/client/lib/opentelemetry/metrics.ts
PavelPashov: @nkaradzhov to make sure we don't do unnecessary checks after initiating the metrics singleton if the user hasn't provided any included or excluded commands

packages/client/lib/client/cache.ts
nkaradzhov: TODO: cacheEntry.value could be many things - string, number, boolean, etc.

packages/client/lib/client/socket.ts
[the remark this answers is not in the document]
nkaradzhov: leave as is, If we need to fix something, it's going to be in the swap logic.

packages/client/lib/client/cache.ts
PavelPashov: As agreed with @elena-kolevska & @nkaradzhov, this metric is now a noop, until we can measure the actual size from the received buffer

packages/client/lib/opentelemetry/metrics.ts
```diff
@@ -0,0 +1,975 @@
...
+      [OTEL_ATTRIBUTES.redisClientCscReason]: reason,
+    });
+  }
+
+  public recordNetworkBytesSaved(value: unknown, clientId?: string) {
+    // TODO: `redis.client.csc.network_saved` cannot be computed correctly here
+    // because CSC stores transformed replies and does not retain raw byte size.
+    // Implement this at a lower protocol/parsing layer where response bytes are known.
+  }
+}
```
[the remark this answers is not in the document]
PavelPashov: As agreed with @elena-kolevska & @nkaradzhov, this metric is now a noop, until we can measure the actual size from the received buffer

PavelPashov:
@nkaradzhov & @vladvildanov 

Created a separated [PR](https://github.com/PavelPashov/node-redis/pull/6) for conversion of ObservableGauges to up UpDownCounters.

After the current PR is released I'll need to create 3 issues for:

1. `redis.client.csc.network_saved` - disabled, due to incorrect values computed from inside the metrics class
2. `db.client.connection.count` - state is always `used`, this should be renamed and converted to ObservableGauge in order to be more accurate.
3. `db.client.connection.pending_requests` - disabled, should be renamed and remain a ObservableGauge, as converting it to UpDownCounter will likely result in a performance hit when enabled.

PR #3120, opened 2025-10-30
Rebuild after release - fix #3118

### LINKED ISSUE #3118: Client version in new release marked as beta in NPM registry
### Description

Hello,
First of all, thank you for all the hard work and dedication you've put into maintaining Node-Redis!

I noticed that the redis client incorrectly reports its version as `5.9.0-beta.3` in the new release `5.9.0`.  
I'm wondering if there might have been a mistake with the deployment or a caching issue.

```bash
# In a fresh Node.js project
npm install node-redis@5.9.0
grep "version"  node_modules/@redis/client/dist/package.json # "version": "5.9.0-beta.3",
```

In the parent `package-lock.json`:

```json
"node_modules/@redis/client": {
  "version": "5.9.0",
  "resolved": "https://registry.npmjs.org/@redis/client/-/client-5.9.0.tgz",
  "integrity": "sha512-EI0Ti5pojD2p7TmcS7RRa+AJVahdQvP/urpcSbK/K9Rlk6+dwMJTQ354pCNGCwfke8x4yKr5+iH85wcERSkwLQ==",
  "license": "MIT",
  "dependencies": {
    "cluster-key-slot": "1.1.2"
  },
  "engines": {
    "node": ">= 18"
  }
},
```

It claims that the client is on version `5.9.0`, but when I actually download the package from the registry:

```bash
wget https://registry.npmjs.org/@redis/client/-/client-5.9.0.tgz
mkdir client-5.9.0
tar -xzvf client-5.9.0.tgz -C client-5.9.0
grep "version" client-5.9.0/package.json # "version": "5.9.0",
# but inside the dist/
grep "version" client-5.9.0/dist/package.json # "version": "5.9.0-beta
```

I could guess that this may be due to caching if `5.9.0-beta.3` and `5.9.0` share the same content.

Best,
Jannis

### Node.js Version

v22.20.0

### Redis Server Version

_No response_

### Node Redis Version

5.9.0

### Platform

Linux

### Logs

```bash

```

zingazzi:
### Description

wget https://registry.npmjs.org/@redis/client/-/client-5.9.0.tgz
mkdir client-5.9.0
tar -xzvf client-5.9.0.tgz -C client-5.9.0
grep "version" client-5.9.0/package.json # "version": "5.9.0",
# but inside the dist/
grep "version" client-5.9.0/dist/package.json # "version": "5.9.0-beta

The release workflow has a timing issue where:
- The build process (npm run build) runs before the version bump
This creates the dist/package.json with the beta version (5.9.0-beta.3)

- Then release-it updates the source package.json to the final version (5.9.0)
But the dist/package.json is never updated, so it still contains the beta version

The Solution:
The release workflow needs to rebuild the packages after the version bump but before publishing to npm.

---

### Checklist

- [ x] Does `npm test` pass with this change (including linting)?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,7 +49,3 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-      # Rebuild packages after version bump to ensure dist/package.json is updated
-      - name: Rebuild packages after version bump
-        run: npm run build
--- a/packages/client/.release-it.json
+++ b/packages/client/.release-it.json
@@ -9,5 +9,8 @@
     "commitMessage": "Release ${tagName}",
     "tagAnnotation": "Release ${tagName}",
     "commitArgs": "--all"
+  },
+  "hooks": {
+    "before:release": "npm run build"
   }
 }


.github/workflows/release.yml
```diff
@@ -49,3 +49,7 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      # Rebuild packages after version bump to ensure dist/package.json is updated
```
jannis-adamek-dt: Hi :) Isn't the package then already on the NPM registry with the previous `npm run releases` step?

zingazzi: You are absoluty right! Sorry my fault. I think another solution to fix this error can be add an Hook in release-it.json.

Something like this

```
"hooks": {
    "before:release": "npm run build"
  }
```

What do you think about this solution?

jannis-adamek-dt: I'm not very familiar with these hooks but it sounds reasonable to me.

zingazzi: I've update the hook. Feel free to merge it if you consider it a valid solution.
Thank you for your patience.

nkaradzhov:
Hi @zingazzi, this has been addressed in #3125, so i will close this one for now

PR #3122, opened 2025-10-31
feat(xreadgroup): add claim attribute

nkaradzhov:
the CLAIM attribute can be used to instruct redis to return PEL ( Pending Entries List ) entries with their respective deliveries and ms since last delivery


packages/client/lib/commands/XREADGROUP.ts
```diff
@@ -5,7 +5,7 @@ import { transformStreamsMessagesReplyResp2 } from './generic-transformers';
 
 /**
  * Options for the XREADGROUP command
- * 
+ *
  * @property COUNT - Limit the number of entries returned per stream
  * @property BLOCK - Milliseconds to block waiting for new entries (0 for indefinite)
  * @property NOACK - Skip adding the message to the PEL (Pending Entries List)
```
PavelPashov, line 11: Do we need to add the `CLAIM` property description here?

PavelPashov (approved):
LGTM

packages/client/lib/commands/XREADGROUP.ts
nkaradzhov: Thanks, added

Issue #3108, opened 2025-10-23
Redis client stucks in reconnection loop


sergeyzwezdin:
### Description

```typescript
import { createClient } from 'redis'

// ...

        this.redisClient = createClient({
            url: 'redis://localhost:6379',
            disableOfflineQueue: true,
            socket: {
                connectTimeout: 1000,
            },
        })
            .on('error', () => {
                console.log('🙉 REDIS ERROR')
            })
            .on('connect', () => {
                console.log('🙃 REDIS CONNECT')
            })
            .on('ready', () => {
                console.log('😎 REDIS READY')
            })
            .on('end', () => {
                console.log('👋 REDIS END')
            })
            .on('close', () => {
                console.log('👊 REDIS CLOSE')
            })
            .on('reconnecting', () => {
                console.log('🫠  REDIS RECONNECTING')
            })
            .on('reconnect', () => {
                console.log('🤗 REDIS RECONNECT')
            })
            .on('invalidate', () => {
                console.log('🫡 REDIS INVALIDATE')
            })
```

Steps to reproduce:
1. Start Redis
2. Run code listed above
3. Terminal output:
```
🙃 REDIS CONNECT
😎 REDIS READY
```
4. Stop Redis
5. Terminal output:
```
🙉 REDIS ERROR
🫠  REDIS RECONNECTING
🙃 REDIS CONNECT
🙉 REDIS ERROR
😎 REDIS READY
```

Notice last `REDIS READY`.
Next, once Redis server restarted it could not reconnect.

The bug was introduced starting `5.0.0`. Works okay on `4.7.0`

### Node.js Version

24.7.0

### Redis Server Version

7.4.6

### Node Redis Version

5.9.0

### Platform

macOS

### Logs

```bash

```

nkaradzhov:
@sergeyzwezdin thanks for raising this, will take a look

nkaradzhov:
@sergeyzwezdin how exactly are you starting/stopping the redis server?

This is my code:
```
import { createClient } from "redis";

const client = createClient({
  url: `redis://localhost:6379`,
  disableOfflineQueue: true,
  socket: {
    connectTimeout: 1000,
  },
})
  .on("error", () => {
    console.log("🙉 REDIS ERROR");
  })
  .on("connect", () => {
    console.log("🙃 REDIS CONNECT");
  })
  .on("ready", () => {
    console.log("😎 REDIS READY");
  })
  .on("end", () => {
    console.log("👋 REDIS END");
[...]

PR #3124, opened 2025-11-02
socket_reconnection_guard

zingazzi:
### Description

Seems there was an issue related to reconnection #3108 
The issue was related (I think) by a race condition on socket connect()

adding a guard to prevent multple connecton at the same time the issue seems resolved.
Because is a core component double check it. 

Maybe i can add also a few test in test suite.

---

### Checklist

### DIFF
--- a/packages/client/lib/client/socket.ts
+++ b/packages/client/lib/client/socket.ts
@@ -74,6 +74,10 @@ export default class RedisSocket extends EventEmitter {
   #isReady = false;
 
   get isReady() {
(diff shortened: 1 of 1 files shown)


nkaradzhov:
hi @zingazzi, thanks for the contribution. i will have to take more thorough look into this as its in one of the most critical part of the driver.

nkaradzhov:
@zingazzi, thanks again! This problem was adressed in #3128

PR #3128, opened 2025-11-05
fix(socket): prevent false-ready state when socket errors during handshake

nkaradzhov:
### Description
Fixes race condition where async socket errors during connection handshake don't trigger reconnection. Validates socket state after initiator completes to catch errors swallowed by command handlers.

fixes: #3108

### Checklist

### DIFF
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -1041,8 +1041,6 @@ describe('Client', () => {
       teardown();
     });
 
-    //utils
-
     async function setup(
       socketOptions: Partial<RedisSocketOptions>,
       dropCount: number,
@@ -1087,15 +1085,11 @@ describe('Client', () => {
     // - But socket is immediately destroyed, causing ECONNRESET during handshake
     function setupMockServer(dropImmediately: number) {
       const server = net.createServer(async (socket) => {
-        // socket.on("data", (data) =>
-        //   console.log(data.toString().replaceAll("\r\n", " ")),
-        // );
         if (dropImmediately > 0) {
           dropImmediately--;
           socket.destroy();
         }
         socket.write("+OK\r\n+OK\r\n");
-        // socket.write("*7\r\n$6\r\nserver\r\n$5\r\nredis\r\n$7\r\nversion\r\n$5\r\n7.2.0\r\n$5\r\nproto\r\n:3\r\n$2\r\nid\r\n:1\r\n")
       });
       return server;
     }


packages/client/lib/client/index.spec.ts
```diff
@@ -1008,6 +1012,95 @@ describe('Client', () => {
...
+      const server = net.createServer(async (socket) => {
+        // socket.on("data", (data) =>
+        //   console.log(data.toString().replaceAll("\r\n", " ")),
+        // );
+        if (dropImmediately > 0) {
+          dropImmediately--;
+          socket.destroy();
+        }
+        socket.write("+OK\r\n+OK\r\n");
+        // socket.write("*7\r\n$6\r\nserver\r\n$5\r\nredis\r\n$7\r\nversion\r\n$5\r\n7.2.0\r\n$5\r\nproto\r\n:3\r\n$2\r\nid\r\n:1\r\n")
```
PavelPashov: Should we uncomment/delete this?

packages/client/lib/client/index.spec.ts
```diff
@@ -1008,6 +1012,95 @@ describe('Client', () => {
...
+      return log;
+    }
+
+    // Create a TCP server that accepts connections but immediately drops them <dropImmediately> times
+    // This simulates what happens when Docker container is stopped:
+    // - TCP connection succeeds (OS accepts it)
+    // - But socket is immediately destroyed, causing ECONNRESET during handshake
+    function setupMockServer(dropImmediately: number) {
+      const server = net.createServer(async (socket) => {
+        // socket.on("data", (data) =>
```
PavelPashov: Should we delete this?

packages/client/lib/client/index.spec.ts
```diff
@@ -1008,6 +1012,95 @@ describe('Client', () => {
...
+          "reconnecting",
+          "connect",
+          "ready",
+        ],
+        log,
+      );
+      teardown();
+    });
+
+    //utils
```
PavelPashov: Should we delete this?

PavelPashov (approved):

packages/client/lib/client/index.spec.ts
nkaradzhov: thanks, fixed

packages/client/lib/client/index.spec.ts
nkaradzhov: thanks, fixed

packages/client/lib/client/index.spec.ts
nkaradzhov: thanks, fixed

Issue #2999, opened 2025-06-17
scanIterator on RedisSentinel


benoitvidis:
### Motivation

Currently, `RedisSentinel` does not implement scan iterators (tested on `@redis/client 5.5.6`).

The iterator is not available from a master lease either.

Even if the code is easy to replicate, this could be a nice addition.

Thank you!

### Basic Code Example

```JavaScript

```

nkaradzhov:
Hi @benoitvidis, thanks for flagging this, we will consider adding it. Would you be willing to contribute maybe?

harshrai654:
Hello, I'm very interested in implementing scanIterator support for the `RedisSentinel` client and would like to contribute to this feature. It's a key requirement for us as we're migrating our application from a single Redis master to a high-availability Sentinel setup, and we rely on this functionality extensively.
My initial investigation into the codebase suggests that using [acquire()](https://github.com/redis/node-redis/blob/master/docs/sentinel.md#master-client-lease) could be a good way to ensure the iteration happens against a stable master node. My proposed solution would look something like this conceptually:
```
async* scanIterator(options) {
  // Acquire a client lease to ensure all SCAN commands go to the same master
  const masterClient = await this.acquire();
  try {
    let cursor = '0';
    do {
      const reply = await masterClient.scan(cursor, options);
      cursor = reply.cursor;
      yield reply.keys;
    } while (cursor !== '0');
  } finally {
    // Ensure the lease is always released
    masterClient.release();
  }
}
```

As I continue to explore the source code to fully understand the implications of this approach, I'm focusing on a few key areas. I will continue to dig into these, but I was hoping you might be able to provide some guidance if you have answers to below questions:
**Failover Behaviour:** How a client lease from `acquire()` behaves during a master failover. For example, if the master goes down mid-iteration, will the leased client throw an error, (assuming it will) or is there a mechanism that would allow it to gracefully reconnect to the newly promoted master?
**Replica Reads:** A huge benefit would be to offload these read-only `SCAN` operations to replicas. For this to work with an iterator, a lease would need to be "sticky" to a single replica for the duration of the scan. I'm looking into how replica clients are managed in the connection pool to see if acquiring a stable lease on a specific replica is currently possible.

Thanks!

nkaradzhov:
@harshrai654 , 

- Failover behavior of `acquire()`: the lease survives a master failover. This sounds good, but a scan cursor is node-specific, meaning a cursor created on one node doesnt work on another node.
[...]

PR #3141, opened 2025-11-26
Sentinel scan iterator

harshrai654:
### Description

This PR implements the scanIterator method for the `RedisSentinel` client.
Previously, scanIterator was not exposed on the Sentinel client, limiting the ability to iterate over keys on the master node directly through the Sentinel interface. This implementation adds that capability with  handling for failovers.

Issue: #2999 

Changes:
- **Automatic Failover Handling:** If a master failover occurs during iteration, the iterator detects the `MASTER_CHANGE` topology event and automatically restarts the scan from the beginning on the new master. This ensures all keys are eventually covered, even if the topology changes mid-scan.
- **Reset-on-Failover Strategy**: Since Redis cursors are stateful and specific to a node instance, the iterator deliberately restarts from cursor 0 (the beginning) upon detecting a master failover. This guarantees data completeness but may result in duplicate keys being yielded, as documented.
---

### Checklist

---

> [!NOTE]
> **Medium Risk**
> New master-key iteration API with explicit failover semantics and pool/lease behavior; incorrect caller handling could miss keys or retry incorrectly, but changes are localized to Sentinel scanning.
> 
> **Overview**
> Adds **`scanIterator`** on the **Redis Sentinel** client so callers can page over keys on the current master with the same **SCAN** options as standalone clients (including an optional starting **`cursor`**).
> 
> Each page runs **`SCAN`** through **`_execute`**, so the master lease is held only for that call and released before **`yield`**—avoiding deadlocks when the loop body runs other commands (e.g. **`mGet`**) with **`masterPoolSize: 1`** or **`reserveClient: true`**.
> 
> On **`MASTER_CHANGE`** while another page is still needed (cursor not yet **`0`**), iteration **stops** with a new **`ScanIteratorInterruptedError`** instead of reusing a node-local cursor on a new master; plain connection errors are **not** wrapped. Termination compares the cursor with **`cursor.toString() !== '0'`** so **Buffer** cursors from type mapping do not loop forever.
> 
> **`ScanIteratorOptions`** is **exported** from the client package. **Sentinel docs** describe iterator usage, failover semantics, and pool behavior. **Tests** cover happy path, **`MATCH`**, synthetic and real failover, race inside **`_execute`**, listener cleanup, and regressions above.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0cae392d32fefdc29ca029c665c63dad297d8860. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/sentinel.md
+++ b/docs/sentinel.md
@@ -160,3 +160,27 @@ try {
   clientLease.release();
 }
 ```
+
+## Scan Iterator
+
+The sentinel client supports `scanIterator` for iterating over keys on the master node:
+
+```javascript
+for await (const keys of sentinel.scanIterator()) {
+  // ...
+}
+```
+
+If a failover occurs during the scan, the iterator will automatically restart from the beginning on the new master to ensure all keys are covered. This may result in duplicate keys being yielded. If your application requires processing each key exactly once, you should implement a deduplication mechanism (like a `Set` or Bloom filter).
+
+```javascript
+const processed = new Set();
+for await (const keys of sentinel.scanIterator()) {
+  for (const key of keys) {
+    if (processed.has(key)) continue;
+    processed.add(key);
+
+    // process key
+  }
+}
+```
--- a/docs/sentinel.md
+++ b/docs/sentinel.md
@@ -175,17 +175,21 @@ for await (const keys of sentinel.scanIterator()) {
 
 SCAN cursors are node-local — a cursor returned by one Redis instance is meaningless on any other instance. Because of this, the sentinel iterator cannot transparently survive a master failover: the in-flight cursor cannot be resumed on the promoted replica, and silently restarting from cursor `0` on the new master would hide both duplicate keys (already yielded from the old master) and data loss (writes that had not yet replicated before the failover).
 
-Instead, if the master changes while an iteration is in progress, the iterator throws `SentinelMasterChangeError`. The caller decides whether to retry the iteration from scratch, accept the partial result, or fail the surrounding operation.
+Instead, if the iterator observes a `MASTER_CHANGE` topology event while an iteration is in progress, it throws `ScanIteratorInterruptedError`. The caller decides whether to retry the iteration from scratch, accept the partial result, or fail the surrounding operation.
+
+Connection-level errors raised by the underlying client (e.g. `SocketClosedUnexpectedlyError`, `SocketTimeoutError`, `ReconnectStrategyError`) are **not** wrapped. A dropped socket is not by itself evidence of a failover — it may also be a transient network blip on the same master, in which case the cursor is still valid and a higher-level retry policy is appropriate. The original error is propagated as-is, and the caller can distinguish failover from a blip by checking for `ScanIteratorInterruptedError` versus other error types.
+
+In a real failover the dropped socket often precedes the Sentinel `MASTER_CHANGE` event (gated by `down-after-milliseconds`), so callers that want to treat both signals uniformly should catch both `ScanIteratorInterruptedError` **and** connection-class errors:
 
 ```javascript
-import { SentinelMasterChangeError } from '@redis/client';
+import { ScanIteratorInterruptedError } from '@redis/client';
 
 try {
   for await (const keys of sentinel.scanIterator()) {
     // ...
   }
 } catch (err) {
-  if (err instanceof SentinelMasterChangeError) {
+  if (err instanceof ScanIteratorInterruptedError) {
     // master failed over mid-iteration; restart from the beginning if desired
   } else {
     throw err;
--- a/docs/sentinel.md
+++ b/docs/sentinel.md
@@ -171,7 +171,15 @@ for await (const keys of sentinel.scanIterator()) {
 }
 ```
 
-If a failover occurs during the scan, the iterator will automatically restart from the beginning on the new master to ensure all keys are covered. This may result in duplicate keys being yielded. If your application requires processing each key exactly once, you should implement a deduplication mechanism (like a `Set` or Bloom filter).
+### Semantics and differences from standalone `scanIterator`
+
+The standalone `RedisClient.scanIterator()` inherits SCAN's documented guarantees: a full iteration returns every key present from start to end, and may return a key multiple times. See the [SCAN guarantees](https://redis.io/docs/latest/commands/scan/#scan-guarantees) page.
+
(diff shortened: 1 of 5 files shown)


nkaradzhov:
Hi @harshrai654, thanks for the contribution! On first glance, this looks good, gut it looks like there are some failing tests. Can you please resolve those, so we can have more thorough look on the final state of the code

harshrai654:
Hey @nkaradzhov I have fixed the tests. A stopped node in previous test of the describe block was causing issues in the next test which requires healthy nodes in the sentinel. Added cleaning up of nodes of frame between these two tests.

docs/sentinel.md
```diff
@@ -101,3 +101,27 @@ try {
...
+
+The sentinel client supports `scanIterator` for iterating over keys on the master node:
+
+```javascript
+for await (const keys of sentinel.scanIterator()) {
+  // ...
+}
+```
+
+If a failover occurs during the scan, the iterator will automatically restart from the beginning on the new master to ensure all keys are covered. This may result in duplicate keys being yielded. If your application requires processing each key exactly once, you should implement a deduplication mechanism (like a `Set` or Bloom filter).
```
elimelt: nit: Bloom filter is not really a sound way to deduplicate on its own, so I'm not sure about the phrasing "`Set` or Bloom filter".

I'm not an expert on Sentinel or Redis replication, but is there any chance the two values differ between old and new master?

nkaradzhov:
@harshrai654 thank you so much! I will need to take a thorough look into this before passing it, and I am extremely busy on a big change that needs to happen very soon. So, apologies, but I will have to delay my full review here.

docs/sentinel.md
harshrai654: - I agree, A Bloom filter isn’t a perfect deduplication mechanism.
The reason I mentioned it is because it provides a very low-memory way to reduce duplicates for large keyspaces, even though it can still produce false positives and **skip some unseen keys.**
I’ll update the documentation to clarify this tradeoff: using a Set gives strict, 100% deduplication, while a Bloom filter is only suitable when occasional false positives are acceptable in exchange for significantly lower memory usage.

- Yes, the results may differ between the old and new master.
Redis replication is asynchronous, so a replica promoted to master may not contain exactly the same
keyspace as the previous master at the moment of failover. Some writes may not yet have replicated,
and new writes may occur on the promoted master after the failover.
Because the iterator restarts from cursor 0 on the new master, any additions or deletions that
happened around the failover will be reflected in the new scan.
Reference: https://redis.io/docs/latest/commands/scan/#scan-guarantees

harshrai654:
> @harshrai654 thank you so much! I will need to take a thorough look into this before passing it, and I am extremely busy on a big change that needs to happen very soon. So, apologies, but I will have to delay my full review here.

No worries, Thanks for the update and for letting me know!

docs/sentinel.md
elimelt: - Yeah I only called it out because of the implication you made, i.e. "application requires processing each key exactly once" -> "use set or bloom filter"
- Difference in semantics between this and a regular scan is probably worth noting in the docs as well

TheIndra55:
This seems to cause a deadlock when the consumer tries to do another operation during the scan and there's not enough master clients available.

```js
for await (const keys of sentinel.scanIterator()) {
    console.log(keys)

    // Hangs forever
    const values = await sentinel.mGet(keys)
    console.log(values)
}
```

harshrai654:
Hey @TheIndra55, thanks for catching this! You are absolutely correct.

Acquiring the master connection for the entire lifespan of the iterator causes a resource starvation deadlock. Since `acquire` reserves a client from the pool and doesn't release it until the iterator finishes, any operation inside the consumer's loop (like your `mGet` example) that requires a connection will block indefinitely if the pool is exhausted (e.g., a single-connection pool).

Even with a larger pool, this implementation unnecessarily holds a connection idle while the user's code executes, reducing overall throughput.

The fix can be to do an **"Acquire->Scan->Release"** loop: we should acquire the connection only for the `scan` operation itself and release it immediately before yielding to the user. This ensures the pool is free for other operations during the consumer's processing phase.

Here is the proposed implementation:

```typescript
async *scanIterator(
  this: RedisSentinelType<M, F, S, RESP, TYPE_MAPPING>,
  options?: ScanOptions & ScanIteratorOptions
) {
  let cursor = options?.cursor ?? "0";
  let shouldRestart = false;

  // This listener persists for the entire iterator life to track failovers
  const handleTopologyChange = (event: RedisSentinelEvent) => {
    if (event.type === "MASTER_CHANGE") {
      shouldRestart = true;
    }
  };
  this.on("topology-change", handleTopologyChange);

  try {
    do {
      // 1. Acquire lease JUST for the scan command
      const masterClient = await this.acquire();
      
      let reply;
      try {
        // If master changed since last loop, reset cursor to 0
        if (shouldRestart) {
          cursor = "0";
          shouldRestart = false;
        }
        
        reply = await masterClient.scan(cursor, options);
        
      } finally {
        // 2. Release IMMEDIATELY so the pool is available for the user
        masterClient.release();
      }

      // Handle edge case: Topology changed *during* the scan
      if (shouldRestart) {
        // Data might be from the old master, so we discard this batch 
        // and restart from cursor 0 on the new master.
        cursor = "0"; 
        continue;
      }

      // 3. Yield to user
      // The connection is released, so `await sentinel.mGet()` will succeed
      cursor = reply.cursor;
      yield reply.keys;

    } while (cursor !== "0");

  } finally {
    this.removeListener("topology-change", handleTopologyChange);
  }
}
```

This ensures we respect the pool limits and handle topology changes correctly by resetting the cursor if a failover occurs between (or during) iterations.

What do you think?

nkaradzhov:
Hi @harshrai654 — first off, apologies for sitting on this for so long.
I pushed your Dec 9 proposal (acquire/release per scan page) onto the branch and rebased over master:

- Reworked the docs (`docs/sentinel.md`) to 
  - (a) make the `Set` vs Bloom filter trade-off explicit per @elimelt's nit, 
  - (b) call out the divergence from standalone SCAN guarantees, and 
  - (c) document the per-page lease behavior so users know commands inside the loop are safe.
- Added a regression test (`should not deadlock when consumer runs commands inside the loop with masterPoolSize 1`)

packages/client/lib/sentinel/index.ts
```diff
@@ -654,6 +656,71 @@ export default class RedisSentinel<
...
+          throw err;
+        }
+
+        cursor = reply.cursor;
+        yield reply.keys;
+        // Cursor may be a Buffer when a Blob String type mapping is in use;
+        // compare by string value so iteration actually terminates.
+      } while (cursor.toString() !== '0');
+    } finally {
+      this.removeListener('topology-change', handleTopologyChange);
```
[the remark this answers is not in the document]
nkaradzhov: doesnt matter, the user gets all keys BEFORE the failover

nkaradzhov (approved):

Issue #3144, opened 2025-12-05
Redis Sentinel ip address/hostname mapping config


elimelt:
### Motivation

It would be really helpful if we could map the ip/hostname sentinel responds with to a custom address for testing/proxying, like how redis cluster supports `nodeAddressMap`

I can open a PR if this is an acceptable feature to add

### Basic Code Example

```JavaScript
createSentinel({
  ...
  nodeAddressMap: {
    '10.0.0.1:30001': {
      host: 'external-host.io',
      port: 30001
    },
    '10.0.0.2:30002': {
      host: 'external-host.io',
      port: 30002
    }
  }
}
```

PavelPashov:
Thanks for the proposal. We can look into it, once #3145 is ready we’ll be happy to review it

elimelt:
cool, marked as ready for review

PR #3145, opened 2025-12-05
implement nodeAddressMap for sentinel

elimelt:
### Description

Adds support for nodeAddressMap to sentinel client options, equivalent to [cluster's nodeAddressMap feature](https://github.com/redis/node-redis/blob/master/docs/clustering.md#node-address-map).

I'd like to be able to use/test Sentinel from outside a k8s cluster, but returned addresses aren't resolvable to the client

Closes https://github.com/redis/node-redis/issues/3144

---

### Checklist

### DIFF
--- a/packages/client/lib/sentinel/index.ts
+++ b/packages/client/lib/sentinel/index.ts
@@ -5,7 +5,7 @@ import { CommandOptions } from '../client/commands-queue';
 import { attachConfig } from '../commander';
 import COMMANDS from '../commands';
 import { ClientErrorEvent, NamespaceProxySentinel, NamespaceProxySentinelClient, NodeAddressMap, ProxySentinel, ProxySentinelClient, RedisNode, RedisSentinelClientType, RedisSentinelEvent, RedisSentinelOptions, RedisSentinelType, SentinelCommander } from './types';
-import { clientSocketToNode, createCommand, createFunctionCommand, createModuleCommand, createNodeList, createScriptCommand, parseNode } from './utils';
+import { clientSocketToNode, createCommand, createFunctionCommand, createModuleCommand, createNodeList, createScriptCommand, getMappedNode, parseNode } from './utils';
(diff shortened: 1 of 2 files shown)


PavelPashov:
Thank you for the PR. I’ve added it to my review list.

elimelt:
@PavelPashov any way we could try and re-run the tests for 8.0.2 to see if this is a flake?

packages/client/lib/sentinel/index.ts
```diff
@@ -1426,6 +1442,16 @@ export class RedisSentinelFactory extends EventEmitter {
...
+  #getNodeAddress(address: string): RedisNode | undefined {
+    switch (typeof this.options.nodeAddressMap) {
+      case 'object':
+        return this.options.nodeAddressMap[address];
+
+      case 'function':
+        return this.options.nodeAddressMap(address);
+    }
+  }
+
```
nkaradzhov: Is this a duplicate?

packages/client/lib/sentinel/index.ts
```diff
@@ -716,16 +718,30 @@ class RedisSentinelInternal<
     );
   }
 
+  #getNodeAddress(address: string): RedisNode | undefined {
```
nkaradzhov: nit: we could take host port and make the string there instead of having to construct the key everywhere

nkaradzhov (changes requested):
@elimelt thanks for this, it looks good in general. I just left couple of nits

packages/client/lib/sentinel/index.ts
elimelt: Good point

packages/client/lib/sentinel/index.ts
elimelt: Yeah, we want both classes to be able to translate, but I can get rid of the duplication and just put this in a util function instead

elimelt:
@nkaradzhov fixed!

Issue #2987, opened 2025-06-04
Wrong return type of hGet


zdila:
### Description

```ts
export const redisClient = createClient(options).withTypeMapping({
  [RESP_TYPES.BLOB_STRING]: String,
});

// Type 'string | {}' is not assignable to type 'string'. Type '{}' is not assignable to type 'string'.ts(2322)
const res: string = await redisClient.hGet("foo", "bar");
```

### Node Redis Version

@redis/client 5.5.5

### Logs

```bash

```

charlie-harvey-beable:
Its not just `hGet()`. It is every single function. Why is every function returning `x | {}`?

jake-knerr:
I suspect it is due to String being a function and a Constructor.

jake-knerr:
This worked for me:

```js
.withTypeMapping({
    [RESP_TYPES.BLOB_STRING]: (val) => (val === null ? null : val.toString()),
    [RESP_TYPES.BLOB_ERROR]: (val) => (val === null ? null : val.toString()),
  })
```

Providing a custom decoder instructs Typescript to infer string | null instead of the string and the String constuctor object.

PR #3150, opened 2026-01-06
fix(client): unwrap constructors to primitives in type mapping

watersRand:
### Description

**Resolves: #2987**

This pull request fixes a type inference bug where using global constructors (like `String`, `Number`, or `Boolean`) in `.withTypeMapping()` resulted in incorrect return types, specifically appearing as `string | {}` or `number | {}` instead of the expected primitives.

**The Problem:**
When a user provides `String` (the `StringConstructor`) to the mapping, TypeScript's `infer` keyword in `ReplyWithTypeMapping` captures the **interface** `String` (the object wrapper) rather than the **primitive** `string`. Because this interface is a non-null object, it is often represented as `{}` in type unions, leading to the confusing `string | {}` signature for commands like `hGet`.

**The Solution:**
I introduced an `UnwrapConstructor<T>` utility type that explicitly maps global constructors to their primitive counterparts:

* `StringConstructor` → `string`
* `NumberConstructor` → `number`
* `BooleanConstructor` → `boolean`
* `BigIntConstructor` → `bigint`

This ensures that the `Extract` logic within the mapping correctly identifies the intended primitive type.

---

### Checklist

* [x] Does `npm test` pass with this change (including linting)?
* [x] Is the new or changed code fully tested?
* [x] Is a documentation update included? (N/A - Fixes existing type behavior)

### DIFF
--- a/packages/client/lib/RESP/types.spec.ts
+++ b/packages/client/lib/RESP/types.spec.ts
@@ -35,15 +35,7 @@ describe('Comprehensive RESP Type Mapping', () => {
             );
         });
 
-        it('DOUBLE: maps to number', async () => {
-            const res = await client
-                .withTypeMapping({
-                    [RESP_TYPES.DOUBLE]: Number
-                })
-                .hello();
 
-            assert.ok(res === null || typeof res === 'number');
-        });
     });
 
     describe('Complex Strings', () => {


nkaradzhov:
@watersRand thanks for this PR, much appreciated. I will need to check TS performance, as we already have a big issue, so hopefully this doesnt add to it. Will keep in touch.

nkaradzhov:
@watersRand i made a PR with some fixes, please review and merge it, so we can close this one :)

PR #3158, opened 2026-01-15
feat(client): add VRANGE command for vector sets

nkaradzhov:
Implement the VRANGE command which returns elements in a lexicographical range from a vector set. This provides a stateless iterator for elements inside a vector set.

The command supports:
- Lexicographical range with inclusive ([) and exclusive (() prefixes
- Special range markers: - (minimum) and + (maximum)
- Optional count parameter to limit results (negative returns all)

Added comprehensive tests including parseCommand unit tests and integration tests for both RESP2 and RESP3 protocols.

### Description

---

### Checklist


PavelPashov (approved):
LGTM

PR #3159, opened 2026-01-18
fix(client): return Promise consistently in cluster functions

orgads:
It is confusing for the same function to sometimes return an object and sometimes return a Promise.

### Checklist


orgads:
@sera bypass

nkaradzhov (changes requested):
@orgads thanks for this. In general, it looks good. Can you please explicitly annotate the return type of the updated methods?

orgads:
> @orgads thanks for this. In general, it looks good. Can you please explicitly annotate the return type of the updated methods?

Sure, done.

orgads:
@sera bypass

PR #3162, opened 2026-01-22
[8.6] Xinfostream - new idempotency fields

nkaradzhov:
### Description

New fields in response:

- idmp-duration - The duration value configured for the stream’s IDMP map.
- idmp-maxsize - The maxsize value configured for the stream’s IDMP map.
- pids-tracked - The number of idempotent pids currently tracked in the stream.
- iids-tracked - The number of idempotent ids currently tracked in the stream - This 	count reflects active iids that haven't expired or been evicted yet.
- iids-added - The count of all entries with an idempotent iid added to the stream during its lifetime - This is a cumulative counter that increases with each idempotent entry added.
- iids-duplicates - The count of all duplicate iids (for all pids) detected during the stream's lifetime - This is a cumulative counter that increases with each duplicate iid. 

https://docs.google.com/document/d/1vUj47wK8lIvE294IOG8BA5FlU5Jj9j3KHOD4lFERnEk

---

### Checklist

### DIFF
--- a/packages/client/lib/commands/XINFO_STREAM.spec.ts
+++ b/packages/client/lib/commands/XINFO_STREAM.spec.ts
@@ -29,6 +29,14 @@ describe('XINFO STREAM', () => {
         'entries-added': 0,
         'recorded-first-entry-id': '0-0',
       },
+      ...testUtils.isVersionGreaterThan([8, 6]) && {
+        'idmp-duration': 100,
+        'idmp-maxsize': 100,
+        'pids-tracked': 0,
+        'iids-tracked': 0,
+        'iids-added': 0,
+        'iids-duplicates': 0,
+      },
       groups: 1,
       'first-entry': null,
       'last-entry': null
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version: ["18", "20", "22"]
-        redis-version: ["rs-7.4.0-v1", "8.0.2", "8.2", "8.4.0"]
+        redis-version: ["rs-7.4.0-v1", "8.0.2", "8.2", "custom-21183968220-debian-amd64"]
```
PavelPashov: Should we replace 8.0.2 instead of 8.4.0?

package-lock.json
```diff
@@ -713,6 +713,278 @@
         "tslib": "^2.4.0"
       }
     },
+    "node_modules/@esbuild/aix-ppc64": {
```
PavelPashov: Seems like this file is committed by mistake?

packages/client/lib/commands/XINFO_STREAM.ts
```diff
@@ -4,14 +4,20 @@ import { isNullReply, transformTuplesReply } from './generic-transformers';
...
  * @property last-generated-id - Last generated message ID
  * @property max-deleted-entry-id - Highest message ID deleted (Redis 7.2+)
  * @property entries-added - Total number of entries added (Redis 7.2+)
  * @property recorded-first-entry-id - ID of the first recorded entry (Redis 7.2+)
+ * @property idmp-duration - The duration value configured for the stream's IDMP map
+ * @property idmp-maxsize - The maxsize value configured for the stream's IDMP map
+ * @property pids-tracked - The number of idempotent pids currently tracked in the stream
+ * @property iids-tracked - The number of idempotent ids currently tracked in the stream
+ * @property iids-added - The count of all entries with an idempotent iid added to the stream
+ * @property iids-duplicates - The count of all duplicate iids detected during the stream's lifetime
```
PavelPashov: Should we mention these are added in 8.6?

packages/client/lib/commands/XINFO_STREAM.spec.ts
```diff
@@ -29,6 +29,14 @@ describe('XINFO STREAM', () => {
         'entries-added': 0,
         'recorded-first-entry-id': '0-0',
       },
+      ...testUtils.isVersionGreaterThan([8, 4]) && {
```
PavelPashov: Should this version be `[8, 6]`?

packages/client/lib/commands/XINFO_STREAM.ts
nkaradzhov: yes

package-lock.json
nkaradzhov: yes, will remove

.github/workflows/tests.yml
nkaradzhov: yes, totally

PR #3163, opened 2026-01-22
[8.6] Xcfgset - new command

nkaradzhov:
### Description

NOTE: tests will fail for XINFO_STREAM command and that is expected and unrelated to this PR

Configures the idempotency parameters for a stream's IDMP map. Sets how long Redis remembers each iid and the maximum number of iids to track. This command clears the existing IDMP map (Redis forgets all previously stored iids), but only if the configuration value actually changes.

https://docs.google.com/document/d/1Bf9ElvtVmhTj9Q6AHj8Lf9-1dsK_I9FEh5v5vi56TBU

---

### Checklist

### DIFF
--- a/packages/client/lib/commands/XCFGSET.spec.ts
+++ b/packages/client/lib/commands/XCFGSET.spec.ts
@@ -1,0 +1,87 @@
+import { strict as assert } from 'node:assert';
+import testUtils, { GLOBAL } from '../test-utils';
+import XCFGSET from './XCFGSET';
+import { parseArgs } from './generic-transformers';
+
+describe('XCFGSET', () => {
+  describe('transformArguments', () => {
+    it('simple', () => {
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version: ["18", "20", "22"]
-        redis-version: ["rs-7.4.0-v1", "8.0.2", "8.2", "8.4.0"]
+        redis-version: ["rs-7.4.0-v1", "8.0.2", "8.2", "custom-21183968220-debian-amd64"]
```
PavelPashov: Should we replace 8.0.2 instead of 8.4.0?

nkaradzhov: yes, addressed in another PR

PR #3164, opened 2026-01-22
[8.6] Hotkeys - new command

nkaradzhov:
### Description

NOTE: tests will fail for XINFO_STREAM command and that is expected and unrelated to this PR

Starts/stops tracking of hotkeys in a redis-server instance and reports top K such keys. Hotkeys are measured by two metrics - cpu execution time and total bytes read/written from the network. Reporting returns meaningful results only during or after stopping tracking.

https://docs.google.com/document/d/14BKzU7u0b-B54b18AbW3WZEZWvIDwGzMPjHFg-Tksl0
---

### Checklist

### DIFF
--- a/packages/client/lib/commands/HOTKEYS_GET.spec.ts
+++ b/packages/client/lib/commands/HOTKEYS_GET.spec.ts
@@ -4,8 +4,6 @@ import HOTKEYS_GET from './HOTKEYS_GET';
 import { parseArgs } from './generic-transformers';
 
 describe('HOTKEYS GET', () => {
-  testUtils.isVersionGreaterThanHook([8, 4]);
-
   it('transformArguments', () => {
     assert.deepEqual(
       parseArgs(HOTKEYS_GET),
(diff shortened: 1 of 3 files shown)


packages/client/lib/commands/HOTKEYS_GET.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    await client.get('testKey');
+
+    // Stop tracking
+    await client.hotkeysStop();
+
+    // GET should still return data in STOPPED state
+    const reply = await client.hotkeysGet();
+    assert.notEqual(reply, null);
+
+    if (reply !== null) {
```
PavelPashov: This if seems redundant given the assert above it

packages/client/lib/commands/HOTKEYS_GET.spec.ts
```diff
@@ -0,0 +1,100 @@
+import { strict as assert } from 'node:assert';
+import testUtils, { GLOBAL } from '../test-utils';
+import HOTKEYS_GET from './HOTKEYS_GET';
+import { parseArgs } from './generic-transformers';
+
+describe('HOTKEYS GET', () => {
+  testUtils.isVersionGreaterThanHook([8, 4]);
```
PavelPashov: This seems redundant if we're enforcing it in the `minimumDockerVersion` as well

packages/client/lib/commands/HOTKEYS_RESET.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { strict as assert } from 'node:assert';
+import testUtils, { GLOBAL } from '../test-utils';
+import HOTKEYS_RESET from './HOTKEYS_RESET';
+import { parseArgs } from './generic-transformers';
+
+describe('HOTKEYS RESET', () => {
+  testUtils.isVersionGreaterThanHook([8, 4]);
```
PavelPashov: This seems redundant if we're enforcing it in the `minimumDockerVersion` as well

packages/client/lib/commands/HOTKEYS_STOP.spec.ts
```diff
@@ -0,0 +1,35 @@
+import { strict as assert } from 'node:assert';
+import testUtils, { GLOBAL } from '../test-utils';
+import HOTKEYS_STOP from './HOTKEYS_STOP';
+import { parseArgs } from './generic-transformers';
+
+describe('HOTKEYS STOP', () => {
+  testUtils.isVersionGreaterThanHook([8, 4]);
```
PavelPashov: This seems redundant if we're enforcing it in the `minimumDockerVersion` as well

PavelPashov (approved):

packages/client/lib/commands/HOTKEYS_GET.spec.ts
nkaradzhov: thanks, adjusted

packages/client/lib/commands/HOTKEYS_RESET.spec.ts
nkaradzhov: thanks, adjusted

packages/client/lib/commands/HOTKEYS_STOP.spec.ts
nkaradzhov: thanks, adjusted

PR #3165, opened 2026-01-23
Add socket.servname property to createClient configuration.

itrich:
### Description

Our Redis server needs SNI enabled during the TLS handshake. It took me a while to find this property in the respective [Node.js documentation](https://nodejs.org/api/tls.html#tlsconnectoptions-callback). To hopefully make it easier for future users, I added this property to the client config documentation.


nkaradzhov:
Thanks @itrich, i will take a look

PR #3173, opened 2026-02-05
Hotkeys Adjustmens

nkaradzhov:
- Change time units from ms to us for CPU time fields
- Add SlotRange interface for slot range representation
- Update field names (byCpuTime → byCpuTimeUs, etc.)
- Fix HOTKEYS_STOP to return null in empty state
- Update response parsing for new array-wrapped format
- Add test for HOTKEYS_STOP empty state behavior

---

### Checklist

### DIFF
--- a/packages/client/lib/commands/HOTKEYS_GET.ts
+++ b/packages/client/lib/commands/HOTKEYS_GET.ts
@@ -25,7 +25,7 @@ export interface HotkeysGetReply {
   sampleRatio: number;
   selectedSlots: Array<SlotRange>;
   /** Only present when sample-ratio > 1 AND selected-slots is not empty */
-  sampledCommandSelectedSlotsUs?: number;
+  sampledCommandsSelectedSlotsUs?: number;
   /** Only present when selected-slots is not empty */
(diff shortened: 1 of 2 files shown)


packages/client/lib/commands/HOTKEYS_GET.ts
```diff
@@ -105,7 +105,7 @@ function transformHotkeysGetReply(reply: UnwrapReply<HotkeysGetRawReply>): Hotke
       case 'selected-slots':
         result.selectedSlots = parseSlotRanges(value as unknown as Array<ArrayReply<NumberReply>>);
         break;
-      case 'sampled-command-selected-slots-us':
+      case 'sampled-commands-selected-slots-us':
         result.sampledCommandSelectedSlotsUs = Number(value);
```
PavelPashov: Should we update the field in the response to `sampledCommandsSelectedSlotsUs` as well?

nkaradzhov: totally, i thought i updated it...

Issue #3175, opened 2026-02-08
Unable to connect to literal IPv6 addresses via the `url` option


TimWolla:
### Description

When using the `url` option one cannot connect to IPv6 addresses:

```
let redis = require('redis');
let c = redis.createClient({ url: 'redis://[::1]:6379', socket: { reconnectStrategy: false } });

c.connect();
```

results in:

```
Error: getaddrinfo ENOTFOUND [::1]
    at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:122:26)
```

It works fine when passing the IPv6 address within the `hostname`:

```
let redis = require('redis');
let c = redis.createClient({ socket: { reconnectStrategy: false, hostname: '::1', port: 6379 } });

c.connect();
```

------------------------

It worked fine with node-redis 3.x, but no longer works with node-redis 4.x. The issue is that the `parse` function from the `url` module stripped the square brackets, but the global `URL` constructor does not:

```
> require('url').parse('redis://[::1]:6379');
Url {
  protocol: 'redis:',
  slashes: true,
  auth: null,
  host: '[::1]:6379',
  port: '6379',
  hostname: '::1',
  hash: null,
  search: null,
  query: null,
  pathname: '/',
  path: '/',
  href: 'redis://[::1]:6379/'
}
> new URL('redis://[::1]:6379')
URL {
  href: 'redis://[::1]:6379',
  origin: 'null',
  protocol: 'redis:',
  username: '',
  password: '',
  host: '[::1]:6379',
  hostname: '[::1]',
  port: '6379',
  pathname: '',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

------

This is a new version of #2734, which I am unable to reopen.

### Node.js Version

v24.13.0

### Node Redis Version

5.10.0

### Platform

Linux

### Logs

```bash
node:internal/process/promises:394
    triggerUncaughtException(err, true /* fromPromise */);
    ^

Error: getaddrinfo ENOTFOUND [::1]
    at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:122:26)
Emitted 'error' event on Class instance at:
    at RedisSocket.<anonymous> (/pwd/node_modules/@redis/client/dist/lib/client/index.js:445:18)
    at RedisSocket.emit (node:events:508:28)
    at #shouldReconnect (/pwd/node_modules/@redis/client/dist/lib/client/socket.js:132:18)
[...]

PR #3176, opened 2026-02-08
fix(client): Support IPv6 literals in URL

TimWolla:
### Description

Fixes redis/node-redis#3175.

With node.js 22.13.0 / 23.4.0 the manual regex replacement can be avoided in favor of [`SocketAddress.parse()`](https://nodejs.org/api/net.html#socketaddressparseinput).

### Checklist

### DIFF
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -178,7 +178,114 @@ describe('Client', () => {
       });
 
       assert.equal(client?.options?.database, 5);
-    })
+    });
+
+    describe('should correctly parse IPv6 literals', () => {
+      it('redis://[::1]:6379', () => {
+        assert.deepEqual(
+          RedisClient.parseURL('redis://[::1]:6379'),
+          {
+            socket: {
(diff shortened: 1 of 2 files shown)


nkaradzhov:
@TimWolla thanks for the contribution, I will take a look!

nkaradzhov (commented):
@TimWolla 

I've tested it locally and confirmed it works correctly.

The current test only covers `redis://[::1]:6379`. Consider adding tests for these URL patterns as well:

- Full IPv6: `redis://[2001:db8::1]:6379`
- With TLS: `rediss://[::1]:6379`
- Without port: `redis://[::1]`
- With database: `redis://[::1]:6379/5`
- With auth: `redis://user:pass@[::1]:6379`

LGTM otherwise! 👍

TimWolla:
@nkaradzhov Thank you. Added the extra tests.

Please apologize the extra noise with the force-pushes, I don't have a local set-up that is compatible with the test suite, so needed to test by pushing into my fork :see_no_evil:

PR #3181, opened 2026-02-20
feat: expose sendCommand on multi for all clients

### LINKED ISSUE #3172: Implement sendCommand inside multi context
### Motivation

Can we have the `sendCommand` which accepts array of strings inside `multi` return context?, I want to abstract different redis libraries via single interface so that application code doesn't has to change according to different favourites out there a simple function like `sendCommand` does the trick but it is not present as soon as i start a multi context..
you can check 2 implementations below:
1. ioredis: https://github.com/LRagji/redis-abstraction/blob/main/source/ioredis-client-pool.ts
2. redis: https://github.com/LRagji/redis-abstraction/blob/main/source/redis-client-pool.ts

### Basic Code Example

```JavaScript
await x.multi()
        .sendCommands([['set', 'a', '1'], ['get', 'a']])
        .exec();
```

nkaradzhov:
This change makes sendCommand a public method on multi/pipeline for standalone client, cluster, sentinel, and pool.

Note: Using sendCommand breaks the type inference for execTyped and execAsPipelineTyped. It is possible to fix this by modifying InternalRedisClientMultiCommandType, but this significantly hinders TypeScript performance and is unfeasible for now.

fixes #3172

### Description

---

### Checklist

### DIFF
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -510,6 +510,16 @@ describe('Client', () => {
       );
     }, GLOBAL.SERVERS.OPEN);
 
+    testUtils.testWithClient('multi sendCommand', async client => {
+      assert.deepEqual(
+        await client.multi()
+          .sendCommand(['SET', 'key', 'value'])
+          .sendCommand(['GET', 'key'])
+          .exec(),
+        ['OK', 'value']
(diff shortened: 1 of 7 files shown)


PavelPashov (approved):
LGTM

PR #3195, opened 2026-03-11
feat: implement diagnostic channels for observability

logaretm:
---

Actual Diff is here: https://github.com/logaretm/node-redis/pull/1/changes
This is currently waiting on #3110 to get merged and then it will show the true size of changes

---

Re-implements otel metrics tracing using diagnostic and tracing channels, while setting up more channels for tracing.

The key shift: **core code now describes what is happening, not what to compute.**

Before, every instrumented code path had to know about OTel, creating closures, capturing timestamps, building attribute maps, recording histograms. The code was telling the system *how* to measure itself. And every metric class needed a noop counterpart for when its metric group was disabled.

Now core code says "a command started," "a connection closed," "an error occurred." Subscribers like OTel, Sentry, a debug logger, anything can independently decide what to compute from those events. 

The domain knowledge in #3110 (which metrics to emit, which semantic conventions to follow, how to classify errors) is preserved. It just moved from being inline in the hot path to being a subscriber that opts in when needed.

**Noops are gone.** When a metric group is disabled, no subscription is created. No subscribers means `hasSubscribers` is false, which means `trace()` and `publish()` skip the channel entirely, zero cost by design, not by maintaining parallel noop implementations. Both `noop-metrics.ts` and `noop-meter.ts` are deleted — when OTel is not initialized, no `OTelMetrics` instance is created, no instruments are registered, no subscriptions exist.

Two functions for all observability:

- **`trace()`** — wraps async operations via `TracingChannel` (commands, batches, connections, pool waits)
- **`publish()`** — emits point events via `dc.channel()` (errors, cache hits/misses, pubsub, maintenance, connection lifecycle)

Both use factory callbacks + `hasSubscribers` checks for zero overhead when no APM subscribes. This is true for metrics, if the user never calls the opentelemetry plugin then no channel subscriptions occur, and no object allocation happens compared to what we had before.

## Channels

### TracingChannel (async lifecycle: start/end/asyncStart/asyncEnd/error)

This sets up tracing with spans in the future, but still can be used by metrics or logs.

| Channel | Wraps |
|---|---|
| `node-redis:command` | Individual command execution |
| `node-redis:batch` | MULTI/PIPELINE as a whole |
| `node-redis:connect` | Client connection |
| `node-redis:connection:wait` | Pool wait for available client |

### Single instance events

This is what metrics usually use, not suitable for traces but can also be used by logs.

| Channel | Emits |
|---|---|
| `node-redis:connection:ready` | Connection established (includes create time) |
| `node-redis:connection:closed` | Connection closed (includes reason) |
| `node-redis:connection:relaxed-timeout` | Maintenance timeout relaxation |
| `node-redis:connection:handoff` | Connection handed off to new node |
| `node-redis:error` | Client/cluster errors |
| `node-redis:maintenance` | Enterprise maintenance notifications |
| `node-redis:pubsub` | PubSub messages in/out |
| `node-redis:cache:request` | Client-side cache hit/miss |
| `node-redis:cache:eviction` | Cache eviction |
| `node-redis:command:reply` | Command reply (for stream lag, pubsub out) |

## What this enables

Any APM can subscribe without monkey-patching or depending on OTel:

```js
import dc from 'node:diagnostics_channel';
import { CHANNELS } from 'redis';

// Create spans
dc.tracingChannel(CHANNELS.TRACE_COMMAND).subscribe({
  start(ctx) { /* start span */ },
  asyncEnd(ctx) { /* end span */ },
  error(ctx) { /* record error, end span */ },
});

// Track errors
dc.channel(CHANNELS.ERROR).subscribe((ctx) => { /* record error metric */ });
```

## Key Changes

- Zero `OTelMetrics` imports in core code
- Zero noop files — both `noop-metrics.ts` and `noop-meter.ts` deleted
- Existing production code: +603 / -1,152 = **-549 lines**
- New channel infrastructure (`tracing.ts`): **+280 lines**
- Net production code: **-269 lines**
- All channel names exported via `CHANNELS` const map
- All event types exported for consumer typing

## Setting Up OTEL Tracing

Almost Everything Is Already Done                                                                                                                                                                                                           
                                                                                                                                                                                                                                                           
- **Async context propagation:** tracePromise (used by trace()) already manages AsyncLocalStorage context. The OTel context manager piggybacks on this. No context.with(trace.setSpan(...)) needed, spans created in start are automatically the active span for downstream operations.                               
- **Sanitized args:** `ctx.args` is already sanitized by `sanitizeArgs()`, so `db.query.text` is just a simple join (e.g., SET mykey ?).                                                                                                                        
- **Error classification:** `getErrorInfo()` already extracts errorType, category, and statusCode for span attributes.                                                                                                                                        

Here is an example of tracing for command calls:

```ts
import { trace, context, SpanKind, SpanStatusCode } from "@opentelemetry/api";
import dc from "node:diagnostics_channel";
import { CHANNELS } from "redis";

const tracer = trace.getTracer("node-redis");

const commandTC = dc.tracingChannel(CHANNELS.TRACE_COMMAND);
commandTC.subscribe({
  start(ctx) {
    const span = tracer.startSpan(ctx.command, {
      kind: SpanKind.CLIENT,
      attributes: {
        "db.system.name": "redis",
        "db.operation.name": ctx.command,
        "db.query.text": ctx.args.join(" "),
        "db.namespace": String(ctx.database),
        "server.address": ctx.serverAddress,
        "server.port": ctx.serverPort,
      },
    });
    // Set span on OTel context — AsyncLocalStorage propagates it
    // to asyncEnd/error automatically via TracingChannel
    trace.setSpan(context.active(), span);
  },
  asyncEnd() {
    trace.getSpan(context.active())?.end();
  },
  error(ctx) {
    const span = trace.getSpan(context.active());
    if (span) {
      span.recordException(ctx.error);
      span.setStatus({
        code: SpanStatusCode.ERROR,
        message: ctx.error?.message,
      });
      span.end();
    }
  },
});                                      
```

Note that the exact implementation can differ since we may need to use `bindStore` on the `AsyncLocalStorage` instance inside OTEL itself, but should be straight forward.

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Touches core command execution, connection, pooling, and cache paths to emit diagnostics events, so regressions could impact performance or error handling despite the largely additive nature. Also raises the minimum supported Node.js version to `>=18.19.0`, which can be a breaking runtime requirement for some consumers.
> 
> **Overview**
> Adds first-class observability via Node.js `diagnostics_channel`/`TracingChannel`: core client paths now emit **tracing** events for commands/batches/connect and **point events** for connections, errors, maintenance, pubsub, client-side cache, command replies, and pool waits (with argument sanitization).
> 
> Refactors OpenTelemetry metrics to be implemented as `diagnostics_channel` subscribers (removing inline metric recording/noop implementations), updates tests accordingly, and exports `CHANNELS` plus payload types from `@redis/client` with new documentation. Updates package `engines.node` across packages to require `>= 18.19.0`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dc7da510733d278acdf3a1959297be3c8b24e6be. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### DIFF
--- a/docs/diagnostics-channel.md
+++ b/docs/diagnostics-channel.md
@@ -12,13 +12,19 @@ import { CHANNELS, type CommandTraceContext } from "@redis/client";
 
 ### TracingChannels (async lifecycle)
 
-Requires Node.js >= 18.19.0. These channels wrap an async operation and emit `start`, `asyncEnd`, and `error` sub-events. Subscribe via `tracing:<name>:start`, `tracing:<name>:asyncEnd`, and `tracing:<name>:error`.
+Requires Node.js >= 18.19.0. These channels use Node.js `TracingChannel#tracePromise()` and emit `start`, `end`, `asyncStart`, `asyncEnd`, and `error` sub-events. `start`/`end` wrap the synchronous portion of the traced callback, while `asyncStart`/`asyncEnd` wrap the returned promise. Subscribe via `tracing:<name>:<event>`, for example `tracing:node-redis:command:start` or `tracing:node-redis:command:asyncEnd`.
 
 ```typescript
 import dc from "node:diagnostics_channel";
 
+// Fired when the command starts.
 dc.subscribe("tracing:node-redis:command:start", ({ command, args }) => {
   console.log(`> ${command}`, args);
+});
+
+// Fired when the async Redis operation settles (success or failure).
+dc.subscribe("tracing:node-redis:command:asyncEnd", ({ command }) => {
+  console.log(`${command} settled`);
 });
 
 dc.subscribe("tracing:node-redis:command:error", ({ command, error }) => {


tlhunter:
FYI @Qard

packages/client/lib/client/index.ts
```diff
@@ -1092,9 +1100,25 @@ export default class RedisClient<
...
+      },
+      this._self.#commandTraceContext(args)
+    );
+  }
+
+  #commandTraceContext(args: ReadonlyArray<RedisArgument>): CommandTraceContext {
+    const socketOptions = this.#options.socket;
+    return {
+      command: String(args[0]).toUpperCase(),
+      args,
```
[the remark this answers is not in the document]
Qard: Might not be the case here, but it's actually intentional that diagnostics_channel be able to modify the given data in some cases. The main use case is adding distributed tracing headers to requests, but may be desired for other things too. Unless you have a compelling reason to do so, you can generally pass mutable data directly into it.

packages/client/lib/client/index.ts
```diff
@@ -1092,9 +1100,25 @@ export default class RedisClient<
...
-    const promise = this._self.#queue.addCommand<T>(args, opts);
-    this._self.#scheduleWrite();
-    return promise;
+    return traceCommand(
+      () => {
+        const promise = this._self.#queue.addCommand<T>(args, opts);
+        this._self.#scheduleWrite();
+        return promise;
+      },
+      this._self.#commandTraceContext(args)
```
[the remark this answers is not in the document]
Qard: Yes, you may want to pass a context builder in so you can use hasSubscribers to only build the context when it would be published.

packages/client/lib/client/index.ts
```diff
@@ -1286,12 +1310,20 @@ export default class RedisClient<
...
+        commands.map(({ args }) => traceCommand(
+          () => this._self.#queue.addCommand(args, {
+            chainId,
+            typeMapping: this._commandOptions?.typeMapping
+          }),
+          {
+            ...this._self.#commandTraceContext(args),
+            batchMode: 'PIPELINE',
+            batchSize
+          }
```
Qard: Might want a context builder function for this too.

packages/client/lib/client/index.ts
```diff
@@ -1328,17 +1360,25 @@ export default class RedisClient<
...
+        traceCommand(
+          () => this._self.#queue.addCommand(args, {
+            chainId,
+            typeMapping
+          }),
+          {
+            ...this._self.#commandTraceContext(args),
+            batchMode: 'MULTI',
+            batchSize
+          }
```
Qard: Context builder here also.

packages/client/lib/client/index.ts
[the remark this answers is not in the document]
logaretm: Don't have a strong opinion here, most of the tracing channels instrumentations I worked on didn't require mutable contexts but could be good to have. 


EDIT: Misunderstood the review 🤦‍♂️

packages/client/lib/client/index.ts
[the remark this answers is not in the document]
logaretm: Yep, addressed it. Now the trace helpers accept a factory for context building.

packages/client/lib/client/index.ts
logaretm: Thanks, I addressed that in latest commits.

logaretm:
@tlhunter @Qard Thank you folks for taking a look, do you think this should be ported over to [ioredis](https://github.com/redis/ioredis) as well?

packages/client/lib/client/index.ts
Qard: Yep, looks like you were addressing it while I was reviewing the prior code at the same time. 😅 

I've reviewed again. All looks good to me now. 🙂

Qard:
Feel free to port it there too. It's valuable to have diagnostics_channel built in to all the libraries. Thanks for working on this! ❤️

Qard (approved):

packages/client/lib/client/tracing.ts
```diff
@@ -0,0 +1,62 @@
...
+interface TracingChannel<ContextType extends object> extends NodeTracingChannel<unknown, ContextType> {
+  readonly hasSubscribers: boolean;
+  tracePromise<T>(fn: () => Promise<T>, context?: ContextType): Promise<T>;
+}
+
+const dc = ('getBuiltinModule' in process)
+  ? (process as any).getBuiltinModule('node:diagnostics_channel')
+  : require('node:diagnostics_channel');
+
+const hasTracingChannel = typeof dc.tracingChannel === 'function';
```
tlhunter: You could use the `dc-polyfill` package from npm. We built it to backport modern diagnostics channel functionality to older versions of Node.js. That may alleviate the checks you have for the presence of tracing channel. But of course that adds a new dependency that node-redis might not want.

https://www.npmjs.com/package/dc-polyfill

logaretm: @tlhunter zero dependency is what makes this attractive for 3rd party authors I think.

Not sure if it works but we could potentially override `getBuiltinModule` and intercept it, kinda like IITM except it wouldn't need module hooks and can live on the APM's SDK side:

something like this:

```ts
process.getBuiltinModule = function(id) {
  // Check for both the prefixed and unprefixed module ID
  if (id === 'node:diagnostics_channel' || id === 'diagnostics_channel') {
    return dcPolyfill;
  }

  // Fall back to original behavior for all other modules
    return originalGetBuiltin.call(process, id);
}
```

Would that work? it kinda imposes them to use `getBuiltInModule` to load in tracing channels but I found that to be common across several libraries like this one due to compatibility concerns with both ESM and older node releases, `pg` is one example.

tlhunter: I think what you currently have is better

PavelPashov:
Hi @logaretm & all, thank you for the PR.

Have you had a chance to look at the OTel metrics [PR](https://github.com/redis/node-redis/pull/3110)? It should be merged into master very soon. We are currently considering adding native OpenTelemetry tracing support as well, ideally building on top of the same OTel integration.

From a maintenance perspective, maintaining separate implementations for metrics and tracing would increase the complexity of the driver. For that reason, our current preference would be to extend the OTEL integration so that tracing builds on top of the upcoming metrics work, rather than introducing a separate diagnostics_channel–based approach.

If native OTel tracing support were available, would that address your use case? As far as we know, most APM platforms already support OTLP.

Qard:
Hi there, creator of diagnostics_channel (and subsequently TracingChannel) here. 👋🏻 

It was actually created _specifically_ for this purpose. Many libraries, and Node.js core itself, did not want to pull in a large dependency on OpenTelemetry or to be tied to its API model. I created diagnostics_channel specifically to enable decoupling the sharing of Observability data from specific consumer APIs so libraries could easily add tracing without additional dependencies.

While much of the market has adopted OpenTelemetry just on hype and the resulting market pressure produced by that hype, diagnostics_channel has _also_ been adopted by most Node.js vendors, and without any hype pushing that along artificially. It's built to be extremely low-cost, being _zero_ cost unless there are actually active subscribers, while the OpenTelemetry API often has actually quite a bit of overhead. 🙈 

A lot of performance-focused libraries like fastify and undici have gone for diagnostics_channel-based instrumentations instead and building OpenTelemetry plugins on top of that as it performs a lot better and is a lot less invasive of a change.

logaretm:
Hey @PavelPashov, thanks for the response. I totally understand the concern about maintaining two separate systems. Building on what @Qard said above, I wanted to show how TracingChannel doesn't compete with OTel, it actually reduces the OTel maintenance burden on your team.

The command duration measurement in #3110 already follows a start/end pattern that TracingChannel handles natively:

```js
// Current approach in #3110 — manual start/end wired into every callsite:
const recordOperation = OTelMetrics.instance.commandMetrics
    .createRecordOperationDuration(args, clientId);
// ...execute...
recordOperation(err);

// TracingChannel — single instrumentation point:
tracePromise(commandChannel, { command, args, serverAddress, ... }, promise);
// start/asyncEnd/error lifecycle managed by Node.js
```

If node-redis emits lifecycle events via TracingChannel:

1. **OTel metrics** can subscribe to those events instead of being wired into every callsite, fewer touchpoints in core, less coupling.
2. **OTel tracing (spans)** comes almost for free, just subscribe to the same channels and create spans. No additional surgery to core code.
3. **Non-OTel consumers** get access too. Sentry recently shipped a [light mode](https://github.com/getsentry/sentry-javascript/issues/19157) that drops OTel entirely and instruments via `diagnostics_channel` directly. These vendors currently have to monkey-patch node-redis, TracingChannel gives them a clean subscription target without any work from your team.

This is the same architecture `node:http` uses today, OTel's own `@opentelemetry/instrumentation-http` subscribes to `diagnostics_channel` events rather than monkey-patching. As @Qard mentioned, fastify and undici have taken the same approach.

Worth noting: OTel JS SDK v3 is [planning to raise the minimum Node.js version to 22](https://github.com/open-telemetry/opentelemetry-js/issues/6420). Users who want to actually use node-redis's OTel metrics or tracing need SDK packages (`sdk-metrics`, `sdk-trace-base`), not just the API, so they'd inherit that Node version constraint. This is already [forcing us at Sentry to drop Node 18 and 20 support](https://github.com/getsentry/sentry-javascript/issues/19775) in our next major, against our usual support philosophy. TracingChannel, as a Node.js built-in, has no such constraint.

I'm happy to rework this PR to sit underneath the metrics layer and power it, and show how it can be used to set up the tracing integration for the future. What do you think?

PavelPashov:
@logaretm & @Qard,

Thanks for the detailed explanation. We’d be open to accepting this, but we need a few things first:
	1.	Sensitive command arguments need to be sanitized. Since third-party libraries may be able to subscribe to these channels, we cannot expose raw arguments by default. We can help identify which commands and argument positions should be treated as sensitive.
	2.	We’d like to see a concrete proposal for how OTel tracing would be built on top of these channels, including what would be emitted, which context would be included, and how subscribers would create and complete spans.
	3.	This also needs to account for our upcoming OTel metrics work in #3110, which we expect to merge into master soon. Please keep that in mind when reworking the PR so the design remains compatible with it.

If anything is unclear or you’d like to discuss the design in more detail, we’d be happy to schedule a meeting.

logaretm:
@PavelPashov Sounds good, I'm aligned with all 3 points and will target the metrics PR as a base to make merging smoother.

Regarding sanitization, I initially thought that it wouldn't matter much since if malcious code gets in then they can do that with patching anyways today. But, for sloppy instrumentations? then it should not leak accidentlly.

I'll update the PR so that the args field in the channel context is sanitized before emission, we can start by following a similar approach to OTel's own redis instrumentation uses with `db.query.text` attribute which is [sanitized by default](https://arc.net/l/quote/nyqmlxja)). 

[This is what they have today](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/redis-common/src/index.ts) for both `node-redis` and `ioredis`, we can adopt that or a similar logic. It should cover all commands like `AUTH` and drop values from `SET` which is acceptable for low cardinality.

packages/client/lib/client/tracing.ts
```diff
@@ -0,0 +1,280 @@
...
+/**
+ * Argument sanitization rules adapted from @opentelemetry/redis-common (Apache 2.0).
+ * Controls how many arguments after the command name are included in trace context.
+ * https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/redis-common/src/index.ts
+ */
+const SERIALIZATION_SUBSETS: Array<{ regex: RegExp; args: number }> = [
+  { regex: /^ECHO/i, args: 0 },
+  { regex: /^(LPUSH|MSET|PFA|PUBLISH|RPUSH|SADD|SET|SPUBLISH|XADD|ZADD)/i, args: 1 },
+  { regex: /^(HSET|HMSET|LSET|LINSERT)/i, args: 2 },
+  { regex: /^(ACL|BIT|B[LRZ]|CLIENT|CLUSTER|CONFIG|COMMAND|DECR|DEL|EVAL|EX|FUNCTION|GEO|GET|HINCR|HMGET|HSCAN|INCR|L[TRLM]|MEMORY|P[EFISTU]|RPOP|S[CDIMORSU]|XACK|X[CDGILPRT]|Z[CDILMPRS])/i, args: -1 },
```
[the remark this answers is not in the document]
logaretm: I think this is a false positive, `SUBSCRIBE` args are channel names, not sensitive data.

packages/client/lib/client/index.ts
```diff
@@ -1080,21 +1170,60 @@ export default class RedisClient<
...
+      (err) => {
+        publish(CHANNELS.ERROR, () => ({
+          error: err,
+          origin: 'client',
+          internal: false,
+          clientId: this._self._clientId,
+        }));
+        throw err;
+      }
+    );
```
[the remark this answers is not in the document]
logaretm: Node docs says:

> Any errors thrown in the message handler will trigger an ['uncaughtException'](https://nodejs.org/api/process.html#event-uncaughtexception).

So it is better to leave it, otherwise we shadow the subscriber error itself.

packages/client/lib/client/index.ts
```diff
@@ -21,6 +21,11 @@ import { BasicCommandParser, CommandParser } from './parser';
 import SingleEntryCache from '../single-entry-cache';
 import { version } from '../../package.json'
 import EnterpriseMaintenanceManager, { MaintenanceUpdate, MovingEndpointType, SMIGRATED_EVENT, SMigratedEvent } from './enterprise-maintenance-manager';
+import { ClientMetricsHandle, ClientRegistry } from '../opentelemetry';
+import { ClientIdentity, ClientRole, generateClientId } from './identity';
+import { trace, sanitizeArgs, publish, CHANNELS, type CommandTraceContext } from './tracing';
+
+const noop = () => {};
```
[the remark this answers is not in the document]
nkaradzhov: ignore this one

packages/client/lib/opentelemetry/metrics.ts
```diff
@@ -0,0 +1,883 @@
...
+    };
+
+    this.#unsubscribers.push(subscribeTC(tc, handlers));
+  }
+
+  #recordError(error: Error, clientId?: string, extra?: Record<string, any>) {
+    const clientAttributes = resolveClientAttributes(clientId);
+    const errorInfo = getErrorInfo(error);
+
+    if (isRedirectionError(errorInfo.statusCode)) return;
```
PavelPashov: This test `should deduplicate client-origin redirection errors but keep cluster-origin` is failing  because we no longer check the origin of the error.

We need to reintroduce the previous logic here.

```ts
if (
  context.origin === METRIC_ERROR_ORIGIN.CLIENT &&
  isRedirectionError(errorInfo.statusCode)
) {
  return;
}
```

packages/client/lib/opentelemetry/metrics.ts
```diff
@@ -175,412 +186,367 @@ class OTelCommandMetrics implements IOTelCommandMetrics {
...
-  }) {
+  private constructor(
+    api: typeof import("@opentelemetry/api"),
+    config?: ObservabilityConfig,
+  ) {
     this.#options = this.parseOptions(config);
-    this.#meter = this.getMeter(api, this.#options);
-    this.#instruments = this.registerInstruments(this.#meter, this.#options);
+    const meter = this.#getMeter(api, this.#options);
+    this.#instruments = this.registerInstruments(meter, this.#options);
```
PavelPashov, line 516: We should add a guard if the metrics are not enabled, for example someone might want to enable the upcoming tracing configuration only.

```ts
this.#options = this.parseOptions(config);

if (!this.#options.enabled) {
  // No-op: don't register any instruments or subscribers
  this.commandMetrics = { destroy() {} };
  this.#channelSubscribers = new OTelChannelSubscribers(options, instruments, []);
  return;
}
```

packages/client/lib/opentelemetry/metrics.spec.ts
PavelPashov, line 550: Related to [this comment](https://github.com/redis/node-redis/pull/3195/changes#r3001706777): the test is broken and producing a false positive (this is my fault), so it needs to be updated with the following configuration.

```ts
OTelMetrics.init({
  api,
  config: {
    metrics: {
      enabled: false,
      meterProvider,
      enabledMetricGroups: ["command"],
    },
  },
});
```

packages/client/lib/client/index.ts
```diff
@@ -1438,75 +1457,76 @@ export default class RedisClient<
...
+            }),
+            () => ({
+              ...this._self.#commandTraceContext(args),
+              batchMode: 'MULTI' as const,
+              batchSize
+            })
+          );
+          traced.catch(noop);
+          promises.push(traced);
+        }
```
PavelPashov, line 1494: Inside MULTI, Redis replies [QUEUED](https://redis.io/docs/latest/develop/using-commands/transactions/#usage) for each command, so TracingChannel marks them as successful and emits asyncEnd before EXEC actually runs. If EXEC later fails, those command-level success traces remain in place even though the commands never executed. That leaves us reporting work as successful when it never actually happened. Only the batch-level trace reflects the real failure.

Maybe it makes sense to treat MULTI as an atomic operation and not trace its individual commands?

packages/client/lib/opentelemetry/metrics.ts
```diff
@@ -175,412 +186,367 @@ class OTelCommandMetrics implements IOTelCommandMetrics {
...
-      [OTEL_ATTRIBUTES.redisClientConnectionNotification]: notification,
+  // -- Resiliency --
+
+  #subscribeResiliency() {
+    // Cluster/internal errors via point-event channel
+    // Skip client-origin redirections (MOVED/ASK) — these are retried
+    // transparently by the cluster client and are not real errors.
+    // Cluster-origin redirections are recorded as they indicate slot migration.
+    this.#sub(CHANNELS.ERROR, (ctx: any) => {
+      if (!ctx.internal && isRedirectionError(getErrorInfo(ctx.error).statusCode)) return;
```
PavelPashov: Here it still makes sense to use the old check, because it's possible for an error to have `internal: false` and still originate from the cluster.

```ts
if (ctx.origin === 'client' && isRedirectionError(getErrorInfo(ctx.error).statusCode)) return;
```

packages/client/lib/client/index.ts
logaretm: I think that makes sense, it would end up showing multiple spans being successful with very short times and their batch parent failing which would be misleading.

docs/diagnostics-channel.md
```diff
@@ -0,0 +1,57 @@
...
+
+```typescript
+import { CHANNELS, type CommandTraceContext } from "@redis/client";
+```
+
+## Channel Types
+
+### TracingChannels (async lifecycle)
+
+Requires Node.js >= 18.19.0. These channels wrap an async operation and emit `start`, `asyncEnd`, and `error` sub-events. Subscribe via `tracing:<name>:start`, `tracing:<name>:asyncEnd`, and `tracing:<name>:error`.
```
nkaradzhov: and also `end` and `asyncEnd`?

PR #3215, opened 2026-03-31
Default Resp3

nkaradzhov:
### Description

---

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Switching the library default from RESP2 to RESP3 and normalizing multiple reply shapes (including object prototypes and stream/search/module replies) can cause subtle runtime/type breakages for consumers that depended on previous defaults or null-prototype objects.
> 
> **Overview**
> **RESP3 becomes the default protocol** across `createClient`/cluster/pool/legacy mode via a new `DEFAULT_RESP = 3`, and docs add a v5→v6 migration guide covering breaking reply-shape/type changes and the new `maintNotifications` default behavior.
> 
> **Unstable RESP3 gating is removed** (`unstableResp3`/`unstableResp3Modules` and per-command `unstableResp3` flags), and several RESP3 transforms are stabilized/normalized (notably `XREAD`/`XREADGROUP` now return a v4/v5-compatible stream list shape on RESP3; `HOTKEYS GET` and `MODULE LIST` normalize map/array/object variants into stable DTOs).
> 
> **Reply/object normalization changes**: RESP3 map decoding and many command transforms now return plain objects instead of `Object.create(null)`, and geo “WITH” commands now parse doubles as `number` (distance/coordinates). Bloom module updates include `CF.INSERTNX` reply typing changing to numeric codes, plus broader test coverage and defaulting Bloom tests to `RESP: 3`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5110f79dcefa1d3514dcf4674a85e44c1ab811ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/client/lib/RESP/decoder.ts
PavelPashov, line 1115: Should we use `Object.defineProperty` here?

packages/client/lib/RESP/decoder.ts
```diff
@@ -924,7 +924,7 @@ export class Decoder {
 
       default:
         return this.#decodeMapAsObject(
-          Object.create(null),
+          {},
```
PavelPashov, line 930: Could this cause issues if the server returns a `__proto__` key in a map? Related to this [comment](https://github.com/redis/node-redis/pull/3215/changes#r3166343155)

packages/test-utils/lib/index.ts
```diff
@@ -822,6 +827,12 @@ export default class TestUtils {
     it(title, async function () {
       if (options.skipTest) return this.skip();
       if (!dockersPromise) return this.skip();
+      const RESP = (options.clusterConfiguration?.RESP ?? 3) as RESP;
+      const {
+        RESP: _RESP,
+        minimizeConnections = false,
```
PavelPashov: Is `minimizeConnections: false` intentional?

packages/bloom/lib/commands/cuckoo/INSERTNX.ts
```diff
@@ -16,5 +16,5 @@ export default {
     args[0].push('CF.INSERTNX');
     parseCfInsertArguments(...args);
   },
-  transformReply: INSERT.transformReply
+  transformReply: undefined as unknown as () => ArrayReply<NumberReply<-1 | 0 | 1>>
```
PavelPashov, line 19: This might break the old RESP 2 responses as now it will return numbers instead of booleans for both protocols.

packages/search/lib/commands/AGGREGATE.ts
```diff
@@ -138,6 +140,67 @@ export interface AggregateReply {
...
+}
+
+function transformAggregateReplyResp3(
+  rawReply: ReplyUnion,
+  preserve?: any,
+  typeMapping?: TypeMapping
+): AggregateReply {
+  if (Array.isArray(rawReply)) {
+    return transformAggregateReplyResp2(rawReply as unknown as AggregateRawReply, preserve, typeMapping);
+  }
```
PavelPashov: Could this break RESP3 clients using `RESP_TYPES.MAP: Array`, since map replies decode as flat arrays and would be parsed as RESP2 here?

packages/search/lib/commands/HYBRID.ts
```diff
@@ -435,34 +441,53 @@ export interface HybridSearchResult {
...
-        }
-      } else {
+      if (
+        key === "id" ||
+        key === "values" ||
+        key.toLowerCase() === "extra_attributes" ||
+        key === "extraAttributes"
+      ) {
+        continue;
+      }
```
PavelPashov, line 483: Will this stop working for RESP2, because we'll still need to `JSON.parse` there?

PavelPashov:
Suggestion: Should we add a `DEFAULT_RESP` const so we don't hardcode 3 everywhere?

packages/client/lib/RESP/decoder.ts
nkaradzhov: Yes, and we added this type of change to the v5-to-v6 migration doc

packages/client/lib/RESP/decoder.ts
nkaradzhov: why?

packages/test-utils/lib/index.ts
nkaradzhov: no, fixing

packages/bloom/lib/commands/cuckoo/INSERTNX.ts
nkaradzhov: that is intentional, documented in v5-to-v6

nkaradzhov: Sorry, i see what you mean. This actually breaks the RESP 2, but we know that the server is in fact returning number in both resp2 and resp3. so its more of a fix for resp2. So i would keep it like this

packages/search/lib/commands/AGGREGATE.ts
nkaradzhov: true, has to be fixed

packages/search/lib/commands/HYBRID.ts
nkaradzhov: JSON.parse still runs — it just moved into parseDocumentValue → assignDocumentField. 
Worth adding a JSON test though.

nkaradzhov:
> Suggestion: Should we add a `DEFAULT_RESP` const so we don't hardcode 3 everywhere?

the value depends on which 3s you mean:

  - True defaults (this.#options.RESP ?? 3 at client/index.ts:689,701,751, plus the RESP = 3 fallback in sentinel/test-util.ts:196 and test-utils/lib/index.ts:655): ~5 sites total — a const would help name the intent.
  - TS generic defaults (RESP extends RespVersions = 3 — ~6 sites in client/index.ts, sentinel/index.ts): cant reference a runtime const, so we would end up with a parallel type DefaultResp = 3 — duplication.
  - Per-test/per-package RESP: 3 as const (~10+ sites across */lib/test-utils.ts and the e2e specs): these are statements of intent ("I am pinning RESP3 here"), not deferrals to a default. Replacing with DEFAULT_RESP would obscure that — and silently change behavior if the const
  ever moved.

packages/search/lib/commands/SEARCH.ts
```diff
@@ -158,6 +159,61 @@ export function parseSearchOptions(parser: CommandParser, options?: FtSearchOpti
...
+  };
+}
+
+function transformSearchReplyResp3(
+  rawReply: ReplyUnion,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- matches TransformReply contract
+  preserve?: any,
+  typeMapping?: TypeMapping
+): SearchReply {
+  if (Array.isArray(rawReply)) {
```
PavelPashov, line 192: With `typeMapping[RESP_TYPES.MAP] = Array`, RESP3 maps also arrive as flat arrays like `['total_results', 1, 'results', ...]`, and this branch would misparse them as RESP2.

packages/redis/index.ts
PavelPashov, line 52: Should we update the `RedisClientPoolType` as well?

PavelPashov (approved):

PR #3281, opened 2026-05-17
fix(client): parse xInfoGroups replies by field

### LINKED ISSUE #2388: Add support for `lag` and `entries-read` results from `XINFO GROUPS` in Redis 7
Redis 7 introduced new `lag` and `entries-read` data items to the output of the `XINFO GROUPS` command.

node-redis currently doesn't support these:

https://github.com/redis/node-redis/blob/c5b6f77c338c84a43f6dd4924277010ea029552a/packages/client/lib/commands/XINFO_GROUPS.ts#L11

Add support for these, using [this PR](https://github.com/redis/node-redis/pull/2368) as an example of how to do it.

raashish1601:
## Summary

Updates the RESP2 `XINFO GROUPS` transformer to build each group from the returned field names instead of fixed array positions.

This keeps existing fields working, preserves Redis 7's `entries-read` and `lag` fields, and makes the transformer tolerate nullable lag values as documented by Redis. It also corrects the reply type for `last-delivered-id` to a blob string and allows `lag` to be null.

Closes #2388.

Redis command reference: https://redis.io/docs/latest/commands/xinfo-groups/

## Validation

- direct `node -r ts-node/register/transpile-only` transformer check for Redis 7 `entries-read`/`lag` fields
- `npm run lint:changed`
- `npm run build`
- `git diff --check`

I also attempted `npx mocha --require ts-node/register/transpile-only packages/client/lib/commands/XINFO_GROUPS.spec.ts --grep transformReply`, but this repo starts its Docker-backed Redis test environment in global hooks before grep-selected tests run. It failed locally at `spawn docker ENOENT` because Docker is not installed in this environment.

---

> [!NOTE]
> **Medium Risk**
> Changes the RESP2 reply transformer and types for `XINFO GROUPS`, which can affect downstream consumers that rely on the previous shape/typing or on positional parsing; behavior should improve but touches command decoding logic.
> 
> **Overview**
> Updates `XINFO GROUPS` RESP2 reply handling to build each group object by *field name* using `transformTuplesReply`, instead of reading fixed tuple positions.
> 
> Adjusts the reply typing to match Redis behavior (`last-delivered-id` as a blob string; `lag` nullable) and adds unit tests covering both pre-Redis-7 and Redis-7+ fields (`entries-read`, `lag`) including null values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 311dbc6a3569308cb828f0a6cd20143477afd7d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/commands/XINFO_GROUPS.spec.ts
+++ b/packages/client/lib/commands/XINFO_GROUPS.spec.ts
@@ -11,6 +11,44 @@ describe('XINFO GROUPS', () => {
     );
   });
 
+  it('transformReply', () => {
+    assert.deepEqual(
+      XINFO_GROUPS.transformReply[2]([[
+        'name', 'group',
+        'consumers', 0,
+        'pending', 0,
+        'last-delivered-id', '0-0'
+      ]]),
+      [Object.assign(Object.create(null), {
+        name: 'group',
+        consumers: 0,
+        pending: 0,
+        'last-delivered-id': '0-0'
+      })]
+    );
+  });
+
+  it('transformReply - Redis 7 fields', () => {
+    assert.deepEqual(
+      XINFO_GROUPS.transformReply[2]([[
+        'name', 'group',
+        'consumers', 0,
+        'pending', 0,
+        'last-delivered-id', '0-0',
+        'entries-read', null,
+        'lag', null
+      ]]),
+      [Object.assign(Object.create(null), {
+        name: 'group',
+        consumers: 0,
+        pending: 0,
+        'last-delivered-id': '0-0',
+        'entries-read': null,
+        lag: null
+      })]
+    );
+  });
+
   testUtils.testAll('xInfoGroups', async client => {
     const [, reply] = await Promise.all([
       client.xGroupCreate('key', 'group', '$', {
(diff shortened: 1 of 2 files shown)


nkaradzhov:
Hi @raashish1601, thanks for the PR, i will look into it, hopefully soon!

nkaradzhov (changes requested):
- Keep the input typed as `UnwrapReply<Resp2Reply<XInfoGroupsReply>>` instead of the looser `Array<Array<BlobStringReply | NumberReply | NullReply>>` — then the trailing `as
  unknown as XInfoGroupsReply['DEFAULT']` cast can go away.
  - Forward `preserve` / `typeMapping` to `transformTuplesReply` like `HGETALL.ts:14` does, so users who configure `typeMapping[RESP_TYPES.MAP] = Map` get a `Map` back. Not a
  regression, but worth fixing while you're here.
  - Add a unit test with the fields in a non-canonical order (e.g. `lag` / `entries-read` before `last-delivered-id`) — that's the property the refactor is actually buying, and
   neither new test exercises it.

nkaradzhov (changes requested):
Thanks @raashish1601. The entries-read and lag fields from https://github.com/redis/node-redis/issues/2388 are already returned by the current XINFO GROUPS transformer on master (positional handling + existing test), so that part is already covered.

The useful bit here is the typing: last-delivered-id is a stream ID (0-0), not NumberReply, and lag can be null. Both worth fixing.

Could you scope this down to just the two type fixes — last-delivered-id: BlobStringReply and lag: NumberReply | NullReply — and leave transformReply as-is? Happy to re-review then.

nkaradzhov:
@raashish1601 closing this due to inactivity. Feel free to reopen with the requested changes.

PR #3299, opened 2026-05-29
test(client): fix three flaky tests around socket/connect lifecycle

nkaradzhov:
## Summary

Three flaky tests in `packages/client` consolidated into one branch, one commit each:

- **`Socket > socketTimeout > should timeout with positive socketTimeout values`** — listener attachment race let the socket-timeout error surface as uncaught on slow CI.
- **`Client > ConnectionTimeoutError`** — `connectTimeout: 1` raced a 1ms event-loop block; on fast hardware the `connect` event landed before the timeout callback and the connection succeeded.
- **`Client > PubSub > should resubscribe`** — `once(subscriber, 'error')` removed itself after the first emit, leaving the second (transient) error during the forced reconnect uncaught.

All three are wired up via the same root-cause shape: relying on a one-shot `once` listener for an event source that can fire more than once. The fix in each case is a permanent listener plus a one-shot promise gate to keep the test deterministic.

The `ConnectionTimeoutError` case is purely a timing margin bump.

## Test plan

- [x] `should timeout with positive socketTimeout values` × 10 runs locally — all pass
- [x] `ConnectionTimeoutError` × 5 runs locally — all pass
- [x] `should resubscribe` × 10 runs locally — all pass
- [x] Full Socket spec — 11/11 pass
- [x] Full PubSub spec — 5/5 pass
- [ ] CI on Node 18, 20, 22, 24

---

> [!NOTE]
> **Low Risk**
> Only test harness changes in spec files; no production client or socket behavior is modified.
> 
> **Overview**
> Stabilizes three flaky **client** integration tests that were racing socket/connect lifecycle timing and mishandling **multiple `error` emissions**.
> 
> For **Pub/Sub resubscribe** and **socketTimeout**, tests no longer rely on `once(..., 'error')` alone: they add a **permanent `error` listener** plus a **one-shot promise** to await the first error while swallowing follow-up errors (e.g. after `CLIENT KILL` or with `reconnectStrategy: false`).
> 
> The **socketTimeout** case also **connects explicitly** (`disableClientSetup`), bumps the timeout to **200ms**, sets **`reconnectStrategy: false`**, and asserts **`isOpen` synchronously** instead of deferring via `nextTick`.
> 
> The **ConnectionTimeoutError** test **blocks the event loop for ~50ms** (was ~1ms) so a **1ms `connectTimeout`** reliably fires before a successful connect on fast hosts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5fdc1925d7f3f7e0379691affa17677cbb4f8cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/client/socket.spec.ts
+++ b/packages/client/lib/client/socket.spec.ts
@@ -240,39 +240,50 @@ describe('Socket', () => {
   });
 
   describe('socketTimeout', () => {
-    const timeout = 50;
+    const timeout = 200;
     testUtils.testWithClient(
       'should timeout with positive socketTimeout values',
       async client => {
-        let timedOut = false;
-
-        assert.equal(client.isReady, true, 'client.isReady');
-        assert.equal(client.isOpen, true, 'client.isOpen');
-
+        // Attach a permanent error listener before connecting so the timeout
+        // event can never surface as an uncaught error — neither during the
+        // handshake nor in the idle window afterwards. `reconnectStrategy:
+        // false` makes the run deterministic and causes the client to emit
+        // `error` twice, so a permanent listener (not `once`) is required.
+        const errors: Error[] = [];
+        let resolveFirstError!: (err: Error) => void;
+        const firstError = new Promise<Error>(resolve => {
+          resolveFirstError = resolve;
+        });
         client.on('error', err => {
+          errors.push(err);
+          resolveFirstError(err);
+        });
+
+        try {
+          await client.connect();
+          assert.equal(client.isReady, true, 'client.isReady');
+          assert.equal(client.isOpen, true, 'client.isOpen');
+
+          const err = await firstError;
           assert.equal(
             err.message,
             `Socket timeout timeout. Expecting data, but didn't receive any in ${timeout}ms.`
           );
-
           assert.equal(client.isReady, false, 'client.isReady');
-
-          // This is actually a bug with the onSocketError implementation,
-          // the client should be closed before the error is emitted
-          process.nextTick(() => {
-            assert.equal(client.isOpen, false, 'client.isOpen');
-          });
-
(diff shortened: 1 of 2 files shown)


packages/client/lib/client/socket.spec.ts
```diff
@@ -240,39 +240,50 @@ describe('Socket', () => {
...
     testUtils.testWithClient(
       'should timeout with positive socketTimeout values',
       async client => {
-        let timedOut = false;
+        // Attach a permanent error listener before connecting so the timeout
+        // event can never surface as an uncaught error — neither during the
+        // handshake nor in the idle window afterwards. `reconnectStrategy:
+        // false` makes the run deterministic and causes the client to emit
+        // `error` twice, so a permanent listener (not `once`) is required.
+        const errors: Error[] = [];
```
PavelPashov: Do we need this array of errors?

packages/client/lib/client/socket.spec.ts
```diff
@@ -240,39 +240,50 @@ describe('Socket', () => {
...
+        let resolveFirstError!: (err: Error) => void;
+        const firstError = new Promise<Error>(resolve => {
+          resolveFirstError = resolve;
+        });
+        client.on('error', err => {
+          errors.push(err);
+          resolveFirstError(err);
+        });
```
PavelPashov: This might be a bit simpler

```ts
client.on('error', () => {
  // keep a permanent listener for any later error emission
});
const firstError = once(client, 'error') as Promise<[Error]>;
// ...
const [err] = await firstError;
```

PavelPashov (approved):

PR #3307, opened 2026-06-04
fix(client): accept boolean TLS socket options

### LINKED ISSUE #3113: Typescript error configuring client socket at version v5.9 (Heroku example)
### Description

In continuation to https://github.com/redis/node-redis/issues/3023
The issue still happening

Configured according to: https://devcenter.heroku.com/articles/connecting-heroku-redis#connecting-in-node-js
```
import { createClient, type RedisClientType } from 'redis';

const redis_url = process.env.REDIS_URL || 'redis://127.0.0.1:6379';
const client: RedisClientType = createClient({
  url: redis_url,
  socket: {
    tls: redis_url.match(/rediss:/) != null,
    rejectUnauthorized: false,
  },
});
```

Image

### Node.js Version

v22.16.0

### Redis Server Version

_No response_

### Node Redis Version

5.9.0

### Platform

MacOS

### Logs

```bash

```

puneetdixit200:
### Description

Fixes #3113.

`RedisTlsOptions` required `socket.tls` to be the literal type `true`, which rejected the common Heroku-style configuration where TLS is derived from the URL:

```ts
socket: {
  tls: redisUrl.match(/rediss:/) != null,
  rejectUnauthorized: false
}
```

This widens the TLS socket option type to accept a runtime boolean while preserving the existing runtime behavior: `RedisSocket` still creates a TLS socket only when `options?.tls === true`.

I also added a compile-time regression in the client type tests for the Heroku-style configuration.

---

### Checklist

Verification:

- `npm run test:types -w @redis/client`
- `npm run build`
- `npm run lint:changed`
- `git diff --check`
- `npm exec -w @redis/client -- mocha -r tsx --exit './lib/client/socket.spec.ts' './lib/client/index.spec.ts'` ran until Redis-backed hooks tried to start Docker; the non-Docker socket cases passed, then the run failed because Docker is not running locally (`~/.docker/run/docker.sock` missing).

---

> [!NOTE]
> **Low Risk**
> Type-level widening plus internal factory narrowing; TLS is still only used when `tls === true`, with no change to connection security behavior.
> 
> **Overview**
> **Fixes compile-time rejection** of `socket.tls` when it is a runtime `boolean` (e.g. Heroku-style `tls: redisUrl.match(/rediss:/) != null`), by widening `RedisTlsOptions.tls` from the literal `true` to `boolean`.
> 
> **Runtime behavior is unchanged:** `RedisSocket` still opens a TLS connection only when `options?.tls === true`; `false` continues to use plain TCP/IPC. IPC and TCP branches in `#createSocketFactory` now narrow options with explicit casts so spreading defaults stays type-safe.
> 
> A **compile-time regression** in `create-client.types-test.ts` locks in the Heroku-style `createClient` configuration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6bdb2544783d7a7e810fcaac48ebf6efbed966ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/client/socket.ts
+++ b/packages/client/lib/client/socket.ts
@@ -40,7 +40,7 @@ type RedisTcpOptions = RedisSocketOptionsCommon & NetOptions & Omit<
 };
 
 type RedisTlsOptions = RedisSocketOptionsCommon & tls.ConnectionOptions & {
-  tls: true;
+  tls: boolean;
 }
 
 type RedisIpcOptions = RedisSocketOptionsCommon & Omit<
@@ -170,8 +170,9 @@ export default class RedisSocket extends EventEmitter {
 
     // IPC
     if (options && 'path' in options) {
+      const ipcOptions = options as RedisIpcOptions;
       const withDefaults: net.IpcNetConnectOpts = {
-        ...options,
+        ...ipcOptions,
         timeout: undefined,
         onread: undefined,
         readable: true,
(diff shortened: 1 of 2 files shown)


nkaradzhov:
Thanks @puneetdixit200, i will have a look.

nkaradzhov:
Thanks for the PR. The compile error on the Heroku pattern is real, but widening `RedisTlsOptions.tls` to boolean collapses the discriminated union — that's why the new as casts were needed in `#createSocketFactory` — and it makes `{ tls: false, ca: ... }` typeable while silently running plain TCP. That's a permanent loss of type safety to fix a type-ergonomics issue that has workarounds, so we'd prefer not to take this shape.

If you switch scheme dynamically, pass the `rediss:///redis://` url and omit `socket.tls` the url scheme sets it. If you also need TLS sub-options like rejectUnauthorized, use `tls: cond ? (true as const) : undefined`, which keeps the discriminant. We'll track first-class dynamic-boolean support in #3113 but it needs a design that preserves narrowing rather than a blanket widen. Closing this PR in favor of that — appreciate the investigation.

PR #3313, opened 2026-06-18
Fix/issue 3311

### LINKED ISSUE #3311: RedisCluster: sharded Pub/Sub silently dropped after in-place slot migration — event-name mismatch (client emits `sharded-channel-moved`, cluster listens for `server-sunsubscribe`)
### Description

Hi there, I'm having issues with channel resubscription on sharded channels. The below issue description is LLM generated, but reviewed and verified by me.

## Bug

When a sharded Pub/Sub channel's hash slot is migrated **in place between two live nodes** (ordinary `CLUSTER SETSLOT` rebalancing, no node going down), `RedisCluster` permanently stops delivering messages for that channel. The client stays subscribed on the *old* owner — which no longer owns the slot — and is never resubscribed on the new owner. No error reaches the application.

The pub/sub docs say this is "handled automatically for you" with `RedisCluster`; it is not, for the in-place migration case.

## Environment

- `redis` / `@redis/client`: **6.0.0** (latest). Same code on current `master`.
- Server: Valkey 7 (open-source Redis Cluster), 3 masters + 3 replicas.

## Root cause — event-name mismatch

The emitter and the listener use different event names, so they never connect:

- **Emitter** — on a server-initiated sharded unsubscribe push (sent by the migrating node when the slot changes hands), the commands queue invokes its `onShardedChannelMoved` callback, which in `client/lib/client/index.ts` (`#initiateQueue`) emits:
  ```ts
  (channel, listeners) => this.emit('sharded-channel-moved', channel, listeners)
  ```
- **Listener** — the cluster's sharded Pub/Sub client in `client/lib/cluster/cluster-slots.ts` (`#initiateShardedPubSubClient`) listens for a name that is **never emitted anywhere** in the package:
  ```ts
  .on('server-sunsubscribe', async (channel, listeners) => {
    await this.rediscover(client);
    const redirectTo = await this.getShardedPubSubClient(channel);
    await redirectTo.extendPubSubChannelListeners(PUBSUB_TYPE.SHARDED, channel, listeners);
  })
  ```
  (The catch block in this method also has a typo: it emits `'sharded-shannel-moved-error'` — "shannel".)

Because `'server-sunsubscribe'` is never emitted and nothing handles `'sharded-channel-moved'` internally, the rediscover-and-resubscribe logic never runs on an in-place migration.

`RESUBSCRIBE_LISTENERS_EVENT → resubscribeAllPubSubListeners` *does* handle `SHARDED` listeners, but it only fires on node removal / socket reconnect (`'__MOVED'`) and on the Redis Enterprise "smigrated" push (`#handleSmigrated`) — not on an in-place slot move between two live nodes.

## Reproduction (verified)

3-master OSS cluster. Subscribe with a `RedisCluster` client, then migrate the channel's slot to another live master:

```js
import { createCluster } from 'redis'
const cluster = createCluster({ rootNodes: [{ url: 'redis://127.0.0.1:7000' }] })
await cluster.connect()

const channel = 'reprochan'            // slot 7061, owned by :7002
let received = 0
await cluster.sSubscribe(channel, () => received++)
```

Migrate slot 7061 from `:7002` to `:7000` (both stay up; a channel is not a key, so no `MIGRATE` step):

```
valkey-cli -p 7000 cluster setslot 7061 importing <node-id-7002>
valkey-cli -p 7002 cluster setslot 7061 migrating <node-id-7000>
valkey-cli -p 7000 cluster setslot 7061 node      <node-id-7000>
valkey-cli -p 7002 cluster setslot 7061 node      <node-id-7000>
# propagate the final SETSLOT NODE to every master
```

Then `SPUBLISH reprochan "hi"` on the **new** owner `:7000`.

**Observed:**

```
[BEFORE] SPUBLISH on :7002 -> 1 receiver; client received it          ✅
'sharded-channel-moved' EMITTED for channel="reprochan"
SHARDCHANNELS @old-owner :7002 -> ""        (subscription dropped)
SHARDCHANNELS @new-owner :7000 -> ""        (client NOT resubscribed here)
[AFTER]  SPUBLISH on :7000 -> 0 receivers; client received nothing    ❌
```

**Expected:** after migration the subscriber keeps receiving messages. **Actual:** delivery stops permanently, silently.

## Suggested fix

Make emitter and listener agree — either have `#initiateShardedPubSubClient` listen for `'sharded-channel-moved'` (the event actually emitted) or rename the emitted event to `'server-sunsubscribe'`. Also fix the `'sharded-shannel-moved-error'` typo. Note the related correctness problem when several sharded channels hash to different slots (a single `SSUBSCRIBE` with all of them returns `CROSSSLOT`) — resubscription should be grouped per slot (see #2902).

## Workaround (verified)

Listen for `'sharded-channel-moved'` on the underlying node pub/sub client(s) and resubscribe. Confirmed to recover delivery (subscription reappears on the new owner; `SPUBLISH` reaches 1 receiver again). Requires reaching into cluster internals, so it's fragile across versions:

```js
for (const master of cluster.masters) {
  master.pubSub?.client?.on('sharded-channel-moved', ch => cluster.sSubscribe(ch, handler))
}
```

Related: #2902.

### Node.js Version

_No response_

### Redis Server Version

7.x (3+ clients)

### Node Redis Version

6.0.0

### Platform

_No response_

### Logs

```bash

```

aartisonigra:
Fixes #3311

This change resolves a sharded Pub/Sub resubscription issue during in-place slot migrations in Redis Cluster.

Changes
Align sharded Pub/Sub migration event handling.
Ensure resubscribe and rediscover logic is triggered when a sharded channel is moved.
Fix typo in the emitted error event name:
sharded-shannel-moved-error → sharded-channel-moved-error

---

> [!NOTE]
> **Medium Risk**
> Touches cluster sharded Pub/Sub and topology rediscovery during slot moves; behavior change is targeted but migration paths are easy to get wrong.
> 
> **Overview**
> Fixes **#3311**: sharded Pub/Sub listeners were not reattached when a slot moved while the client was already subscribed.
> 
> **`#initiateShardedPubSubClient`** now listens for **`sharded-channel-moved`** (matching what the node client emits) instead of **`server-sunsubscribe`**. On that event it **rediscovers** cluster topology, opens the sharded Pub/Sub client on the new slot owner, and **re-extends** the channel listeners. Cluster error emission is corrected to **`sharded-channel-moved-error`**.
> 
> A cluster integration test covers **subscribe first, then migrate the slot in place** and asserts messages are received again after resubscribe.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 065363985f18a102d893f7f8a007bc45bf893e65. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -748,9 +748,7 @@ export default class RedisClient<
     return new RedisCommandsQueue(
       this.#options.RESP ?? DEFAULT_RESP,
       this.#options.commandsQueueMaxLength,
-      (channel, listeners) => {
-        this.emit('sharded-channel-moved', channel, listeners);
-      },
+      (channel, listeners) => this.emit('sharded-channel-moved', channel, listeners),
       clientId
     );
   }
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -748,6 +748,10 @@ export default class RedisClient<
     return new RedisCommandsQueue(
       this.#options.RESP ?? DEFAULT_RESP,
       this.#options.commandsQueueMaxLength,
+      (channel, listeners) => {
+        this.emit('sharded-channel-moved', channel, listeners);
+        this.emit('server-sunsubscribe', channel, listeners);
+      },
       (channel, listeners) => this.emit('sharded-channel-moved', channel, listeners),
       clientId
     );
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -750,9 +750,7 @@ export default class RedisClient<
       this.#options.commandsQueueMaxLength,
       (channel, listeners) => {
         this.emit('sharded-channel-moved', channel, listeners);
-        this.emit('server-sunsubscribe', channel, listeners);
       },
-      (channel, listeners) => this.emit('sharded-channel-moved', channel, listeners),
       clientId
     );
   }
--- a/packages/search/lib/commands/SEARCH.ts
+++ b/packages/search/lib/commands/SEARCH.ts
@@ -8,36 +8,23 @@ import { getMapValue, mapLikeToObject, mapLikeValues, parseDocumentValue, parseS
 export type FtSearchParams = Record<string, RedisArgument | number>;
 
 export function parseParamsArgument(parser: CommandParser, params?: FtSearchParams) {
-  if (!params) return;
+  if (params) {
+    parser.push('PARAMS');
 
-  parser.push('PARAMS');
+    const args: Array<RedisArgument> = [];
+    for (const key in params) {
+      if (!Object.hasOwn(params, key)) continue;
 
-  // FT.SEARCH expects: PARAMS <num-args> <k1> <v1> <k2> <v2> ...
-  // Where <num-args> is the *number of pairs*.
-  //
-  // The previous implementation incorrectly used `pushVariadicWithLength(args)`
-  // which sets the length to the number of *arguments*, not the required
-  // number of pairs. This causes exact-match queries like `@field:"$x"`
-  // to bind parameters incorrectly on some Redis Stack/FT versions.
-  const pairArgs: Array<RedisArgument> = [];
-  let pairs = 0;
+      const value = params[key];
+      args.push(
+        key,
+        typeof value === 'number' ? value.toString() : value
+      );
+    }
 
-  for (const key in params) {
-    if (!Object.hasOwn(params, key)) continue;
-
-    const value = params[key];
-    pairArgs.push(
-      key,
-      typeof value === 'number' ? value.toString() : value
-    );
-    pairs++;
+    parser.pushVariadicWithLength(args);
   }
-
-  // <num-args> is the number of pairs, so it must be `pairs`.
-  parser.push(pairs.toString());
-  parser.pushVariadic(pairArgs);
 }
-
 
 export interface FtSearchOptions {
   VERBATIM?: boolean;
--- a/packages/search/lib/commands/SEARCH.ts
+++ b/packages/search/lib/commands/SEARCH.ts
@@ -8,23 +8,36 @@ import { getMapValue, mapLikeToObject, mapLikeValues, parseDocumentValue, parseS
 export type FtSearchParams = Record<string, RedisArgument | number>;
 
 export function parseParamsArgument(parser: CommandParser, params?: FtSearchParams) {
-  if (params) {
-    parser.push('PARAMS');
+  if (!params) return;
 
-    const args: Array<RedisArgument> = [];
-    for (const key in params) {
-      if (!Object.hasOwn(params, key)) continue;
+  parser.push('PARAMS');
 
-      const value = params[key];
-      args.push(
-        key,
-        typeof value === 'number' ? value.toString() : value
-      );
-    }
+  // FT.SEARCH expects: PARAMS <num-args> <k1> <v1> <k2> <v2> ...
+  // Where <num-args> is the *number of pairs*.
+  //
+  // The previous implementation incorrectly used `pushVariadicWithLength(args)`
+  // which sets the length to the number of *arguments*, not the required
+  // number of pairs. This causes exact-match queries like `@field:"$x"`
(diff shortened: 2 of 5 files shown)


nkaradzhov:
@aartisonigra thanks, i will have a look!

packages/search/lib/commands/SEARCH.ts
nkaradzhov, line 1: why is the search part of this PR?

packages/client/lib/client/index.ts
```diff
@@ -748,6 +748,10 @@ export default class RedisClient<
     return new RedisCommandsQueue(
       this.#options.RESP ?? DEFAULT_RESP,
       this.#options.commandsQueueMaxLength,
+      (channel, listeners) => {
+        this.emit('sharded-channel-moved', channel, listeners);
+        this.emit('server-sunsubscribe', channel, listeners);
+      },
```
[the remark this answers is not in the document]
nkaradzhov: this has not been resolved really

nkaradzhov (changes requested):
Thanks for tracking down #3311 — the diagnosis is right (the cluster resubscribe never fires on in-place slot migration), but I'd like to land it differently. 

1. server-sunsubscribe is a stale event name, the right one is sharded-channel-moved

So instead of emitting a second event from the queue, we should just point the cluster listener at the event we already emit. That keeps a single, documented event and avoids touching the queue at all.

2. The index.ts change breaks constructor arity. The queue needs no change at all, so please revert the index.ts edit entirely.

3. Please drop the FT.SEARCH changes. 

4. Regression test. The existing tests never exercise the resubscribe path. Here's one that subscribes first, then migrates in place — it fails on the dead listener and passes with the fix:

```ts
    // Regression for #3311: subscribe FIRST, then migrate the slot in place.
    // The server pushes SUNSUBSCRIBE to the already-subscribed client, which
    // must drive the cluster to rediscover and reattach the listener on the
    // new owner. The dead `server-sunsubscribe` listener meant this never
    // happened and the subscription was silently lost after migration.
    testUtils.testWithCluster('should resubscribe a sharded channel after in-place slot migration (#3311)', async cluster => {
      const SLOT = 10328, // slot of `channel`
        migrating = cluster.slots[SLOT].master,
        importing = cluster.masters.find(master => master !== migrating)!,
        [migratingClient, importingClient] = await Promise.all([
          cluster.nodeClient(migrating),
          cluster.nodeClient(importing)
        ]);

      const listener = spy();

      // subscribe BEFORE migration -> the sharded PubSub client attaches to `migrating`
      await cluster.sSubscribe('channel', listener);

      // move the slot in-place to `importing`; `migrating` loses the slot and
      // pushes SUNSUBSCRIBE to the subscribed client
      await Promise.all([
        migratingClient.clusterDelSlots(SLOT),
        importingClient.clusterDelSlots(SLOT),
        importingClient.clusterAddSlots(SLOT),
        migratingClient.clusterSetSlot(SLOT, 'NODE', importing.id),
        importingClient.clusterSetSlot(SLOT, 'NODE', importing.id)
      ]);

      // the reattach is async and sharded PubSub does not buffer, so
      // publish until the resubscribed listener receives the message.
      // With the bug this never reattaches and the loop exhausts -> assertion fails.
      for (let i = 0; i < 50 && !listener.called; i++) {
        await cluster.sPublish('channel', 'message');
        await setTimeout(100);
      }

      assert.ok(listener.calledWithExactly('message', 'channel'));
    }, {
      serverArguments: [],
      minimumDockerVersion: [7]
    });

```
Requires import { setTimeout } from 'node:timers/promises'; at the top of the spec.

Net change: ~2 lines in cluster-slots.ts + the test. No queue/index.ts churn, no RediSearch change.

packages/search/lib/commands/SEARCH.ts
aartisonigra: Sorry, that was my mistake. The SEARCH changes were unrelated to #3311. I've removed them and kept the PR focused on the sharded Pub/Sub migration fix and the regression test.

aartisonigra:
I've addressed the review feedback and updated the PR. Please take another look when you have a chance.

nkaradzhov (approved):

PR #3322, opened 2026-07-08
feat(time-series): add TS.READ cursor read command

nkaradzhov:
This pull request adds client support for the RedisTimeSeries `TS.READ` command (`since 8.10.0`), a read-only, single-key cursor reader that returns a batch of `(timestamp, value)` samples with timestamp at or after a cursor, in ascending order. With the optional `BLOCK` group it waits until at least `minCount` qualifying samples exist or a timeout elapses, enabling continuous batch consumption of historical and newly appended samples without polling `TS.RANGE`.

Wire shape: `TS.READ key timestamp [BLOCK milliseconds min_count] [MAX_COUNT max_count]`. The `timestamp` cursor accepts a non-negative integer or one of the sentinels `-`/`+`/`$` (exposed as `TS_READ_TIMESTAMP`). The `BLOCK` group is all-or-nothing; `minCount` may be omitted in the typed API (defaults to `1`) but is always emitted on the wire. Replies reuse the existing `TS.RANGE` sample parser (RESP2/RESP3).

Behavior considerations:
- Command is marked read-only and is not cacheable (server declares `dont_cache`); cluster clients route by the key's hash slot.
- Blocking connection handling is the caller's responsibility, consistent with existing blocking commands (e.g. `BLPOP`); the command object carries no blocking flag. Callers opting into `BLOCK` should run it on an isolated connection.
- An empty reply is a successful result (nothing qualifies, timeout, or key removal) — no automatic retry.

---

> [!NOTE]
> **Low Risk**
> Additive read-only API with no changes to existing commands; blocking behavior matches other blocking Redis commands and is caller-managed.
> 
> **Overview**
> Adds **client support for RedisTimeSeries `TS.READ`** (Redis 8.10+): a read-only, single-key cursor read that returns ascending `(timestamp, value)` batches from an inclusive cursor, with optional **`BLOCK`** (wait for `minCount`, default 1) and **`MAX_COUNT`**.
> 
> The new `READ` command module exposes **`TS_READ_TIMESTAMP`** sentinels (`-`, `+`, `$`), typed **`TsReadTimestamp`** / **`TsReadOptions`**, and wires **`client.ts.read`** / **`READ`** into the time-series command map. Replies reuse the existing **`TS.RANGE`** sample transformers for RESP2/RESP3.
> 
> **Tests** cover argument encoding (cursors, precision-safe string timestamps, `BLOCK`/`MAX_COUNT`) and integration cases (full read, caps, missing key, `LATEST`, blocking when `minCount` is already met, empty flush on `NEW` + `BLOCK` timeout).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 79ae841b5c9c4e0ddf89a0941934a7af148b8425. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/time-series/lib/commands/READ.ts
+++ b/packages/time-series/lib/commands/READ.ts
@@ -1,6 +1,6 @@
 import { CommandParser } from '@redis/client/dist/lib/client/parser';
 import { RedisArgument, Command, Resp2Reply } from '@redis/client/dist/lib/RESP/types';
-import { Timestamp, transformTimestampArgument, SamplesRawReply, transformSamplesReply } from './helpers';
+import { transformTimestampArgument, SamplesRawReply, transformSamplesReply } from './helpers';
 
 /**
  * Sentinel cursor values for {@link READ}'s `timestamp` argument.
@@ -18,7 +18,7 @@ export const TS_READ_TIMESTAMP = {
   NEW: '$'
 } as const;
 
-export type TsReadTimestamp = Timestamp | typeof TS_READ_TIMESTAMP[keyof typeof TS_READ_TIMESTAMP];
+export type TsReadTimestamp = number | Date | typeof TS_READ_TIMESTAMP[keyof typeof TS_READ_TIMESTAMP];
 
 export interface TsReadOptions {
   /**
--- a/packages/time-series/lib/commands/READ.ts
+++ b/packages/time-series/lib/commands/READ.ts
@@ -1,0 +1,70 @@
+import { CommandParser } from '@redis/client/dist/lib/client/parser';
+import { RedisArgument, Command, Resp2Reply } from '@redis/client/dist/lib/RESP/types';
+import { Timestamp, transformTimestampArgument, SamplesRawReply, transformSamplesReply } from './helpers';
+
+/**
+ * Sentinel cursor values for {@link READ}'s `timestamp` argument.
+ *
+ * The wire values must remain `-`, `+`, and `$`; only the first call of an
+ * iteration should use a sentinel. Every subsequent call should pass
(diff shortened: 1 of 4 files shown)


packages/time-series/lib/commands/READ.ts
```diff
@@ -0,0 +1,70 @@
...
+export const TS_READ_TIMESTAMP = {
+  /** `-` — no lower bound, reads from the earliest sample (equivalent to `0`). */
+  EARLIEST: '-',
+  /** `+` — the latest existing sample's timestamp, inclusive. */
+  LATEST: '+',
+  /** `$` — the latest sample's timestamp + 1: only samples added after the call qualify. Meaningful only with `BLOCK`. */
+  NEW: '$'
+} as const;
+
+export type TsReadTimestamp = number | Date | typeof TS_READ_TIMESTAMP[keyof typeof TS_READ_TIMESTAMP];
```
PavelPashov: Could we preserve precision-safe string cursors by reusing the existing `Timestamp` type here? Redis timestamps can exceed `Number.MAX_SAFE_INTEGER`, and `Timestamp` already supports strings for that case.

```suggestion
export type TsReadTimestamp = Timestamp;
```

PavelPashov (approved):

packages/time-series/lib/commands/READ.ts
nkaradzhov: thanks, added timestamp, but kept the +/-/$ that come from TS_READ_TIMESTAMP

PavelPashov (approved):

PR #3329, opened 2026-07-10
fix(sentinel): emit error when the resolved master is unreachable

### LINKED ISSUE #3066: createSentinel promis doesn't reslove
### Description

I encountered this issue when trying to connect to my redis cluster using SSl/TLS.
If the connection to sentinel succeeds but then the connection to the master redis node doesn't succeed the promise returned from the master client connection never resolves.
No timeout no nothing even though if you try to connect to the redis node with createClient you do recieve a proper error back.

Redis 7.2
Docker

example code
`import { createClient, createSentinel } from "redis";
import fs from 'fs';

const sentinel = await createSentinel({
    name: 'myprimary',
    sentinelRootNodes: [
        {
            host: '127.0.0.1',
            port: 26379,
        }
    ],
    sentinelClientOptions: {
        socket: {
            tls: true,
            rejectUnauthorized: true,
            cert: fs.readFileSync("./certs/client.crt"),
            ca: fs.readFileSync("./certs/ca.crt"),
            key: fs.readFileSync("./certs/client.key"),
        }
    },
    nodeClientOptions: {
        socket: {
            tls: true,
            rejectUnauthorized: true,
            cert: fs.readFileSync("./certs/client1.crt"),
            ca: fs.readFileSync("./certs/ca.crt"),
            key: fs.readFileSync("./certs/client1.key"),
        }
    }
});

sentinel.on("error", (err) => console.log("Redis subscriber Error", err));
sentinel.on("connect", () => console.log("Redis sentinel connected"));

try {
    console.log("connecting...");
    await sentinel.connect();
} catch (e) {
    console.log(e)
}
console.log("connected");
`
If there is an issue with the TLS keys in nodeClientOptions it will log connecting... and nothing else
After some debugging i found the issue to be in
redis/client/dist/lib/sentinel/index.js at line 896
the promises created in that if clause are never resolved.

### Node.js Version

_No response_

### Redis Server Version

_No response_

### Node Redis Version

_No response_

### Platform

_No response_

### Logs

```bash

```

GiHoon1123:
### Description

Fixes #3066.

`RedisSentinelInternal#connect()`'s retry loop discovers the current master via the Sentinel nodes (`observe()`/`analyze()`), then connects a client to it (`transform()`). If sentinel discovery succeeds but the resolved master address is actually unreachable (e.g. a TLS misconfiguration, or a firewalled/NATed address — the exact scenario in #3066), the failure was only ever `console.log`'d on every retry attempt, with nothing an application could observe, catch, or react to. Since the retry loop waits 1s between attempts and only gives up (rejecting `connect()`) after `maxCommandRediscovers` (default 16) attempts, this can look like a ~16 second silent hang with zero visibility into why.

The fix is a single line: emit `'error'` instead of `console.log`-ing it. This matches the existing convention already used elsewhere in the same class — `observe()` already emits `'error'` for a sentinel node it can't reach, and the `PubSubProxy` wiring does the same — so this isn't a new pattern, just applying the existing one to a path that had been missed. The `'no valid master node'` case remains excluded from emission, unchanged from before: that message is a normal, expected transient during a real failover (the old master is briefly flagged down before a replica is promoted), not something worth surfacing as an error.

`RedisSentinel`'s constructor already wires `this.#internal.on('error', err => this.emit('error', err))` before `connect()` can ever run, so this emission is guaranteed to have a listener and won't crash on an unhandled `'error'` event.

A regression test was added (`emits error while retrying an unreachable master`) using `nodeAddressMap` to remap only the resolved master to an unreachable address, so sentinel discovery itself stays healthy the whole time — matching the reported scenario without needing a real TLS setup. It asserts an `'error'` event fires during the retry loop instead of the connect attempt going silent.

**Local verification note**: I was not able to get the full docker-based sentinel test suite running end-to-end in my local environment to watch this specific new test go green (an environment issue on my end — an *existing, unmodified* sentinel test hits the same `beforeEach` timeout, so it's unrelated to this change). I did verify: `tsc --build` is clean, lint is clean, and the existing docker-free sentinel unit tests (`commandOptions` suite) still pass. The new test's control-flow was traced by hand against `destroy()`'s teardown behavior to confirm it won't hang or leave dangling state, but I'd appreciate CI (or a maintainer) confirming it actually passes.

---

### Checklist

- [ ] Does `npm test` pass with this change (including linting)? — lint and `tsc --build` are clean; the docker-free sentinel unit tests pass; the full docker-based suite (including the new test) could not be run end-to-end locally (see note above) and needs CI to confirm.
- [ ] Is the new or changed code fully tested? — a regression test was added, but not verified passing locally (see note above).
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? — N/A, no public API change; `'error'` is an existing, documented event on `RedisSentinel`.

---

> [!NOTE]
> **Low Risk**
> Observability-only change on an existing `error` event path; connect still fails after the same retry limit, with slightly noisier error events during retries.
> 
> **Overview**
> When Sentinel finds a master but **connecting to that address fails**, `RedisSentinelInternal#connect()` no longer only **`console.log`s** each retry—it **`emit('error', err)`** on every attempt (still skipping the expected transient **`no valid master node`** message during failover).
> 
> That surfaces failures such as TLS or NAT/firewall mismatches (#3066) while `connect()` keeps retrying up to **`maxCommandRediscovers`**, instead of a long silent wait. Errors propagate through the existing **`RedisSentinel`** internal → public **`error`** wiring.
> 
> A docker-based regression test uses **`nodeAddressMap`** to point the resolved master at an unreachable port while Sentinels stay healthy, and asserts an **`error`** event during **`connect()`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e6df6a298937f229783df305cdc804aaab211ccf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/sentinel/index.spec.ts
+++ b/packages/client/lib/sentinel/index.spec.ts
@@ -903,6 +903,44 @@ describe('legacy tests', () => {
       assert.notEqual(masterNode!.port, newMaster.port);
     });
 
+    // Regression: when sentinel discovery succeeds but the resolved master is
+    // unreachable (e.g. #3066 - TLS misconfiguration, or a firewalled/NATed
+    // address), #connect()'s retry loop used to silently `console.log` the
+    // failure on every attempt with nothing an application could observe or
+    // catch, leaving `connect()` looking hung until it eventually gave up
+    // after `maxCommandRediscovers` attempts. It should emit `'error'` on
+    // every retry attempt instead.
+    it('emits error while retrying an unreachable master', async function () {
+      this.timeout(60000);
+
+      const masterPort = await frame.getMasterPort();
+
+      sentinel = frame.getSentinelClient({
+        nodeAddressMap: (address: string) => {
+          const [host, portStr] = address.split(':');
+          // Remap only the master to an address nothing listens on, so the
+          // connection attempt fails immediately (ECONNREFUSED) without
+          // ever needing a failover - sentinel itself stays healthy the
+          // whole time, matching the reported scenario.
+          if (Number(portStr) === masterPort) {
+            return { host: '127.0.0.1', port: 1 };
+          }
+          return { host, port: Number(portStr) };
+        }
+      }, false);
+
+      const errors: Array<Error> = [];
+      sentinel.on('error', (err: Error) => errors.push(err));
+
+      const connectPromise = sentinel.connect();
+      connectPromise.catch(() => {});
+
+      // Wait for at least one retry attempt to surface an 'error' event
+      // instead of hanging silently.
+      await assert.doesNotReject(once(sentinel, 'error'));
+      assert.ok(errors.length > 0);
+    });
+
     // if master changes, client should make sure user knows watches are invalid
     it('watch across master change', async function () {
       this.timeout(60000);
--- a/packages/client/lib/sentinel/index.ts
+++ b/packages/client/lib/sentinel/index.ts
@@ -1000,7 +1000,7 @@ export class RedisSentinelInternal<
         }
 
         if (err.message !== 'no valid master node') {
-          console.log(e);
+          this.emit('error', err);
         }
         await setTimeout(1000);
       } finally {


nkaradzhov:
Thanks for digging into #3066 and adding a regression test.

Unfortunately, this change doesn't reach the reported scenario. In #3066 the master clients are created with the default `reconnectStrategy`, so a failed master connection retries internally forever and `client.connect()` never settles. `transform()` then blocks on `Promise.all(...)` indefinitely, and the `catch` in `#connect()` — where this PR replaces `console.log` with `emit('error')` — never executes. That's why the issue reports a permanent hang rather than a ~16s retry loop; the retry loop belongs to sentinel-discovery failures, not an unreachable master.

The test also wouldn't pass: with `passthroughClientErrorEvents` defaulting to `false`, master client connection errors are emitted as `'client-error'`, not `'error'`, so `once(sentinel, 'error')` never resolves.

Also note: an unguarded `this.emit('error', err)` in the retry loop throws `ERR_UNHANDLED_ERROR` for apps without an `'error'` listener, aborting the loop (and causing an unhandled rejection when `#connect` runs from the background scan timer).

The real fix for #3066 is making initial master connection failure observable to `#connect()` — e.g. awaiting the master connect promises with failure/timeout awareness so `connect()` rejects instead of hanging. Closing as-is, but happy to review that approach. Meanwhile, `passthroughClientErrorEvents: true` (or `'client-error'`) already surfaces these errors today.

PR #3340, opened 2026-07-14
feat(client): add LMOVEM and BLMOVEM commands

nkaradzhov:
This pull request adds the `LMOVEM` and `BLMOVEM` commands to `@redis/client`, closing the gap where only a single element could be moved between lists (`LMOVE`/`BLMOVE`) while the pop family already moved many.

- `LMOVEM source destination <LEFT|RIGHT> <LEFT|RIGHT> [<COUNT|EXACTLY> count <OBO|BULK>]` — moves up to (`COUNT`) or exactly (`EXACTLY`) N elements from one end of the source list to one end of the destination; reply is always an array of moved elements (destination order), or `null` when nothing moved.
- `BLMOVEM` — blocking variant, same options plus a `timeout` (seconds, `0` = forever).

## API

Options are modeled as a discriminated union — `{ COUNT: number, ORDER?: 'OBO' | 'BULK' }` XOR `{ EXACTLY: number, ORDER?: 'OBO' | 'BULK' }` — so `COUNT`/`EXACTLY` are mutually exclusive at the type level, and `ORDER` (OBO = stack/reversed, BULK = queue/preserved) only applies when a count is given. Omitting options moves a single element, still as an array reply. Both raw and camelCase aliases are exposed (`LMOVEM`/`lMoveM`, `BLMOVEM`/`blMoveM`).

## Considerations

- New commands (not options on `LMOVE`/`BLMOVE`) so the reply type never changes with arguments.
- Single-slot multi-key: source and destination must be hash-tag collocated in cluster mode; specs use `{tag}` prefixes.
- Commands are unreleased upstream, so behavior tests were authored but not executed locally (no server build exposing `LMOVEM`); argument serialization is type-checked and asserted via `parseArgs`.

---

> [!NOTE]
> **Low Risk**
> Additive client command definitions and tests only; no changes to existing LMOVE/BLMOVE behavior or shared runtime paths beyond new exports.
> 
> **Overview**
> Adds **Redis 8.10** list commands **`LMOVEM`** and **`BLMOVEM`** to `@redis/client`, so callers can move **multiple** elements between lists (alongside existing single-element **`LMOVE`** / **`BLMOVE`**).
> 
> **`LMOVEM`** parses source/destination keys, list sides, and optional **`COUNT`** or **`EXACTLY`** plus required **`OBO`** / **`BULK`** ordering via shared **`parseLMoveMOptions`**. Replies are typed as an array of moved elements or **`null`**. **`BLMOVEM`** mirrors that with a blocking **timeout** and reuses **`LMOVEM`**’s reply handling.
> 
> The public API exposes **`LMOVEM`** / **`lMoveM`** and **`BLMOVEM`** / **`blMoveM`** on the command index with JSDoc. **`LMoveMOptions`** is a discriminated union so **`COUNT`** and **`EXACTLY`** are mutually exclusive at compile time.
> 
> Specs cover argument serialization and integration behavior (empty source, **`COUNT`** clamping, **`EXACTLY`** no-op, **`OBO`** vs **`BULK`**, blocking timeout), gated at Redis **8.10** and using hash-tagged keys for cluster.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d83f9ec7e96a28c1022beb60e2c943cc3893473. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/commands/LMOVEM.ts
+++ b/packages/client/lib/commands/LMOVEM.ts
@@ -21,30 +21,6 @@ export type LMoveMOptions = {
   ORDER?: LMoveMOrder;
 };
 
-export function parseLMoveMArguments(
-  parser: CommandParser,
-  source: RedisArgument,
-  destination: RedisArgument,
-  sourceSide: ListSide,
-  destinationSide: ListSide,
-  options?: LMoveMOptions
-) {
-  parser.pushKeys([source, destination]);
-  parser.push(sourceSide, destinationSide);
-
-  if (options) {
-    if ('EXACTLY' in options) {
-      parser.push('EXACTLY', options.EXACTLY.toString());
-    } else {
-      parser.push('COUNT', options.COUNT.toString());
-    }
-
-    if (options.ORDER !== undefined) {
-      parser.push(options.ORDER);
-    }
-  }
-}
-
 export default {
   IS_READ_ONLY: false,
   parseCommand(
@@ -56,7 +32,20 @@ export default {
     options?: LMoveMOptions
   ) {
     parser.push('LMOVEM');
-    parseLMoveMArguments(parser, source, destination, sourceSide, destinationSide, options);
+    parser.pushKeys([source, destination]);
+    parser.push(sourceSide, destinationSide);
+
+    if (options) {
+      if ('EXACTLY' in options) {
+        parser.push('EXACTLY', options.EXACTLY.toString());
+      } else {
+        parser.push('COUNT', options.COUNT.toString());
+      }
+
+      if (options.ORDER !== undefined) {
+        parser.push(options.ORDER);
+      }
+    }
   },
   transformReply: undefined as unknown as () => ArrayReply<BlobStringReply> | NullReply
 } as const satisfies Command;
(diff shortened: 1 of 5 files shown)


packages/client/lib/commands/LMOVEM.ts
```diff
@@ -0,0 +1,59 @@
...
+ * - `OBO` (one-by-one): push each element as it is popped, reversing the
+ *   moved block's relative order (stack semantics).
+ * - `BULK`: preserve the moved block's original relative order (queue semantics).
+ */
+export type LMoveMOrder = 'OBO' | 'BULK';
+
+export type LMoveMOptions = {
+  /** Move up to `COUNT` elements (fewer if the source has fewer). */
+  COUNT: number;
+  ORDER?: LMoveMOrder;
```
PavelPashov: Could we make `ORDER` required in both option variants and always serialize it? Redis requires the complete `<COUNT|EXACTLY> count <OBO|BULK>` trailer, so omitting `ORDER` currently produces a syntax error.

PavelPashov (approved):

packages/client/lib/commands/LMOVEM.ts
nkaradzhov: thanks, fixed

PR #3270, opened 2026-05-14
feat: add COMMAND DOCS command support

ruguoba:
## Summary

Implement the `COMMAND DOCS` command which returns detailed documentation about Redis commands. This command was introduced in Redis 7.0.

## Changes

- Add `COMMAND_DOCS.ts` with full implementation
- Add `COMMAND_DOCS.spec.ts` with unit tests
- Support optional command name filtering
- Transform reply into structured documentation object

## Usage

```typescript
// Get docs for all commands
const allDocs = await client.commandDocs();

// Get docs for specific commands
const docs = await client.commandDocs('GET', 'SET');
```

## Testing

- Verified that `commandDocs()` returns documentation for all commands
- Tested with specific command names
- Confirmed proper error handling

Fixes #1934

---

> [!NOTE]
> **Low Risk**
> Low risk and additive: introduces a new read-only command and its reply transformation without changing existing command behavior. Main risk is limited to potential edge cases in reply parsing/argument flag normalization across Redis versions.
> 
> **Overview**
> Adds a new `COMMAND DOCS` command implementation (`COMMAND_DOCS.ts`) and wires it into the client command registry as `client.commandDocs()`, with optional filtering by command names.
> 
> Implements RESP2 reply transformation into a typed object, including normalization of the `arguments` field (handling `flags` such as `optional`/`multiple`). Adds a spec (`COMMAND_DOCS.spec.ts`) that validates basic result structure for both unfiltered and filtered calls.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 926d731dd3dc579293937416f9b6e9cb6688d593. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2043, opened 2022-03-16
Support COMMAND DOCS

Avital-Fine:
### Description

[COMMAND DOCS](https://redis.io/commands/command-docs)

Closes #1934 

---

### Checklist

- [ ] Does `npm test` pass with this change (including linting)?
- [ ] Is the new or changed code fully tested?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

PR #3362, opened 2026-07-26
feat: add COMMAND DOCS support

### LINKED ISSUE #1934: Add support for redis command: COMMAND DOCS (R7)

Hashim1999164:
## Summary
- Adds `COMMAND DOCS` (`client.commandDocs`) for Redis 7+, including optional command name filters.
- Transforms RESP2 flat map replies into the RESP3 object shape, including nested `arguments` and recursive `subcommands`.
- Registers the command in the client command index with JSDoc (`@since 7.0.0`).

Fixes #1934

## Why previous attempts were not enough
- #2043 used the old `transformArguments` API and did not match the RESP3-first map reply shape maintainers asked for.
- #3270 hit reply-transform bugs (including treating already-parsed arrays as JSON) and was closed by the author.

## Test plan
- [x] `parseArgs` covers no args, string, and array filters
- [x] Unit test for RESP2 `transformReply[2]` nested map/arguments/subcommands
- [x] Integration: `client.commandDocs()` and `client.commandDocs(['GET', 'SET'])` against Redis 7+
- [x] `npm run check:command-jsdoc` and lint on changed files
---

> [!NOTE]
> **Low Risk**
> Read-only server introspection with localized reply parsing; no auth, data, or connection behavior changes.
> 
> **Overview**
> Adds **`COMMAND DOCS`** as **`client.commandDocs`** / **`COMMAND_DOCS`** (Redis 7+), with optional variadic command name filters.
> 
> The implementation builds `COMMAND DOCS` via `parseCommand`, marks the command read-only, and registers it in the command index with JSDoc (`@since 7.0.0`). For **RESP2**, a dedicated **`transformReply[2]`** turns flat map arrays into the same nested object shape as RESP3—recursive **`arguments`**, **`subcommands`**, and **`withTypeMapping`** support for plain objects vs `Map`/`Array`. RESP3 relies on the default map decoder (`transformReply[3]` is unset).
> 
> Tests cover argument parsing, RESP2 reply transformation, integration against Redis 7+, filtered queries, and Map type mapping on RESP2.
>

### DIFF
--- a/packages/client/lib/commands/COMMAND_DOCS.spec.ts
+++ b/packages/client/lib/commands/COMMAND_DOCS.spec.ts
@@ -1,0 +1,132 @@
+import { strict as assert } from 'node:assert';
+import testUtils, { GLOBAL } from '../test-utils';
+import COMMAND_DOCS from './COMMAND_DOCS';
+import { parseArgs } from './generic-transformers';
+
+describe('COMMAND DOCS', () => {
+  testUtils.isVersionGreaterThanHook([7]);
+
+  describe('transformArguments', () => {
+    it('simple', () => {
+      assert.deepEqual(
+        parseArgs(COMMAND_DOCS),
+        ['COMMAND', 'DOCS']
+      );
+    });
+
+    it('string', () => {
+      assert.deepEqual(
+        parseArgs(COMMAND_DOCS, 'GET'),
+        ['COMMAND', 'DOCS', 'GET']
+      );
+    });
+
+    it('array', () => {
+      assert.deepEqual(
+        parseArgs(COMMAND_DOCS, ['GET', 'SET']),
+        ['COMMAND', 'DOCS', 'GET', 'SET']
+      );
+    });
+  });
+
+  describe('transformReply', () => {
+    it('RESP2 map with nested arguments', () => {
+      assert.deepEqual(
+        COMMAND_DOCS.transformReply[2]([
+          'get',
+          [
+            'summary', 'Get the value of a key',
+            'since', '1.0.0',
+            'group', 'string',
+            'complexity', 'O(1)',
+            'history', [['2.2.3', 'Added MSET']],
+            'arguments', [
+              [
+                'name', 'key',
+                'type', 'key',
+                'display_text', 'key',
+                'key_spec_index', 0
+              ]
+            ]
+          ],
+          'command',
+          [
+            'summary', 'Get or set information about commands',
+            'since', '2.8.13',
+            'group', 'server',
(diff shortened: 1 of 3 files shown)


nkaradzhov (changes requested):
@Hashim1999164 thanks for the PR!

Three points before merge. 

First, the RESP2 transformReply should honor the typeMapping parameter (see transformTuplesReply and CONFIG_GET) so users mapping MAP to Map or Array get consistent results across protocols. 

Second, please add an integration test running commandDocs against a real Redis 7 server with RESP 2 client options — the nested transform is currently covered only by a synthetic fixture, which is exactly where the previous PR broke.

Third, drop the NOT_KEYED_COMMAND flag, it has no runtime consumer and the other COMMAND family commands do not set it.

Hashim1999164:
Addressed the Codex review notes:

1. RESP2 `transformReply` now accepts `typeMapping` and returns `Map` / flat `Array` / plain object according to `RESP_TYPES.MAP`.
2. Nested `arguments` are typed as `CommandDocsArgument` (including recursive `arguments`) so `docs.get.arguments?.[0].name` type-checks.

nkaradzhov (changes requested):
@Hashim1999164 Thanks for the update! The typeMapping handling at the top level looks right now, but two of the three requested changes are still outstanding, and one part of the first remains:

1. The nested transforms (transformDoc, transformArgument) still produce plain objects unconditionally, so with RESP_TYPES.MAP: Map a RESP2 client gets Map<string, object> while RESP3 returns Map<string, Map>. Please either apply the mapping consistently to the nested maps or normalize both protocols to the same shape.
2. Both integration tests use the default RESP3 test setup. Please add a testWithClient run with RESP: 2 client options so the nested RESP2 transform is exercised against a real Redis 7 server — that path is currently only covered by the synthetic fixture.
3. NOT_KEYED_COMMAND is still set — please drop it; it has no runtime consumer and the rest of the COMMAND family does not use it.

One smaller thing while you're in there: reply_schema is recursed through transformDoc, which special-cases keys like arguments inside JSON Schema content and only converts one nesting level. Passing the raw value through on RESP2 would be safer.

Happy to take another look once these are in.

Hashim1999164:
Addressed the remaining review points: nested RESP2 maps now honor the MAP type mapping, `reply_schema` is left untouched, `NOT_KEYED_COMMAND` is removed, and an integration test exercises `commandDocs` against Redis 7 with a RESP2 client and `Map` mapping.

nkaradzhov (approved):

Issue #3256, opened 2026-04-24
MOVED errors with Azure Managed Redis cluster


mderriey:
## Short summary

We're running a Node.js API on Azure App Service that connects to an Azure Managed Redis cluster using [the OSS clustering policy](https://learn.microsoft.com/azure/redis/architecture#clustering) over a private endpoint.

The app is occasionally getting into a state where all issued commands will return `MOVED <slot> <ip>` errors, and the only fix we've found is to restart the API.

## General setup

We were using Azure Cache for Redis, and moved to Azure Managed Redis since Azure announced they're sunsetting the former.

We switched from `createClient` to `createCluster` since AMR runs a "proper" cluster:

```ts
const hostname = '<instance>.<region>.redis.azure.net'
const port = 10000
const client = createCluster({
  rootNodes: [{ url: `rediss://${hostname}:${port}` }],
  defaults: {
    password,
    pingInterval: 60_000,
    socket: {
      tls: true,
      // The Redis server can instruct the client to connect to specific nodes in the cluster, it does so using IP addresses.
      // When connecting to the raw IPs, TLS will fail hostname validation unless we set it explicitly.
      servername: hostname,
    },
  },
})
```

We changed which events we listen to following the advice from <https://github.com/redis/node-redis/blob/master/docs/clustering.md#events>:

```ts
type RedisNode = { host: string; port: number }
client.on('connect', () => redisLogger.info('The Redis cluster has successfully connected and is ready to use'))
client.on('disconnect', () => redisLogger.info('The Redis cluster has disconnected'))
client.on('error', async (error: Error) => {
  redisLogger.info('The Redis cluster has errored', { error })
  // More on this later
  await handleError(error, client)
})
client.on('node-ready', (node: RedisNode) => redisLogger.info('A Redis cluster node is ready to establish a connection', { node }))
client.on('node-connect', (node: RedisNode) => redisLogger.info('A Redis cluster node has connected', { node }))
client.on('node-reconnecting', (node: RedisNode) =>
  redisLogger.info('A Redis cluster node is attempting to reconnect after an error', { node }),
)
client.on('node-disconnect', (node: RedisNode) => redisLogger.info('A Redis cluster node has disconnected', { node }))
client.on('node-error', async (error: Error, node: RedisNode) => {
  redisLogger.warn('A Redis cluster node has errored', { error, node })
  // More on this later
  await handleError(error, client)
})
```

[...]

Issue #2857, opened 2024-10-19
ErrorReply: WRONGPASS invalid username-password pair or user is disabled.


arunsingh28:
### Description

I'm encountering a persistent `WRONGPASS invalid username-password pair or user is disabled` error when trying to connect to a Redis server using the Node.js Redis client. Despite setting the correct password for the default user and verifying it works through the Redis CLI, the connection from the Node.js application continues to fail with the same error. I've followed standard procedures for setting and verifying the password, but the issue remains unresolved.

Note: Authentication works successfully through the Redis CLI, but when attempting to connect via the Node.js application, the WRONGPASS error is thrown.

Node code

```
import * as redis from "redis";
import { env } from "../../config/env";

const redisClient = redis.createClient({
  url: `${env.MODE === "dev" ? "redis://" : "rediss://"}${env.REDIS_URI}`,
  socket: {
    tls: env.MODE !== "dev",
  },
});

redisClient.on("connect", () => {
  console.log(
    env.MODE === "dev" ? "connected to local redis" : "connected to AWS Valkey caches"
  );
});

redisClient.on("error", (err) => {
  console.log("error connecting to redis", err);
});

export default redisClient;
```

here I'm firing the server

```
import App from "./src/app";
import MongoDB from "./src/services/database";
import RedisClient from "./src/services/redis";

import { env } from "./config/env";

const { port } = env;

App.server.listen(port, async () => {
  MongoDB.connectDB();
  await RedisClient.connect();
  console.log(`server listening on port ${port}`);
});
```

this is my `REDIS_URI=default:password@localhost:6379`

### Node.js Version

18.0.0

### Redis Server Version

4.7.0

### Node Redis Version

4.7.0

### Platform

macOS

### Logs

```bash
error connecting to redis [ErrorReply: WRONGPASS invalid username-password pair or user is disabled.]
```

FahmiChbil:
for me  the problem is in prod  with Azure Cache for Redis and the container app just worked when I have enabled keys in Azure Cache for Redis in the authentication part in the Azure portal   
azure cache for redis : 
settings -----> authentication -----> access keys

PR #3368, opened 2026-07-27
Fixes #3366

### LINKED ISSUE #3366: MULTI/pipeline commands have no slotNumber, so a full-node slot migration can route the whole batch to the wrong node
Found while reviewing #3364.

`_executeMulti`/`_executePipeline` queue their commands without a `slotNumber`:

```ts
this._self.#queue.addCommand(args, { chainId, typeMapping })
```

`extractCommandsForSlots(slots)` (used for partial slot migrations) filters by
`slotNumber`, so these batch commands are never matched there — they just stay queued on
the original node, which is fine as long as that node is still alive.

But when a node loses *all* its slots, `extractAllCommands()` drains everything
regardless of slot, and `cluster-slots.ts` forwards the whole result to `lastDestNode` —
whichever destination happened to be processed last while walking the migration, not
necessarily the node that owns the batch's actual key slot. A MULTI/pipeline batch queued
at that point can end up sent to a node that doesn't own its slot, resulting in `MOVED`
errors or an aborted transaction instead of surviving the handoff.

Fixing this needs slot info preserved on batch-queued commands (or the full-node drain
routed per-command instead of to a single `lastDestNode`).

aartisonigra:
## Problem
MULTI/pipeline commands have no slotNumber, so during full-node slot migration (extractAllCommands), the whole batch can be routed to the wrong node (lastDestNode), causing MOVED errors or aborted transactions.

## Changes
- packages/client/lib/client/index.ts: Added optional slotNumber param to _executePipeline() and _executeMulti(), passed to each addCommand() call
- packages/client/lib/cluster/index.ts: MULTI() executors now destructure slotNumber from getClientAndSlotNumber() and pass it to client._executeMulti() / client._executePipeline()

---

> [!NOTE]
> **High Risk**
> Removes the entire Sentinel client surface while exports may remain, and changes cluster MOVED/MULTI routing behavior in production-critical paths.
> 
> **Overview**
> **Cluster MULTI/pipeline (#3366):** `MULTI()` now passes the routing key’s **slot number** into `_executeMulti` / `_executePipeline`, and those methods forward it on every queued `addCommand`. During slot migration, commands without a slot were ignored or mis-routed; tagged batches stay on the correct shard.
> 
> **MOVED recovery (#3256):** On `MOVED`, the client **disconnects** the failing connection, rediscover runs, then retry prefers the master from the error (with slot fallback via the key).
> 
> **`RedisClient.parseURL` (#2857):** Wraps `URL` parsing, rejects empty host with a **percent-encode** hint for passwords containing `@`, plus tests.
> 
> **Breaking:** The full **`lib/sentinel`** implementation (client, commands, tests) is **deleted** in this diff while `createSentinel` may still be exported from the package entry—verify build/exports.
> 
> **Other:** New optional `topologyRefreshInterval` on cluster options; placeholder `MOVED-recovery.spec.ts`; minor `ASK` handler `const` tweak.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba329622ddd96b469625b6184404acb40ea1ac7f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nkaradzhov (changes requested):
@aartisonigra, thanks, the underlying issue is real and passing the slot number from getClientAndSlotNumber into _executeMulti and _executePipeline is the right direction. 

However, this PR bundles several unrelated changes and some are even incorrect.
Please leave only the slotNumber plumbing plus a regression test, and ensure a MULTI chain is relocated only when the entire chain is still in the write queue.

nkaradzhov:
@aartisonigra thans for the effort!  Unfortunately the PR as a whole can't be merged in its current shape, and a competing PR (#3377) already covers the same fix more completely. Closing this one

PR #3376, opened 2026-07-28
test(client): add message assertion test for XCLAIM JUSTID

Piyush0049:
This pull request updates `packages/client/lib/commands/XCLAIM_JUSTID.spec.ts` to resolve the inline maintainer TODO (`// TODO: test with messages`).

Previously, the unit test only asserted behavior against an empty stream (`[]`). This change adds an end-to-end test case that populates stream entries, moves them to pending status via `xReadGroup`, claims them using `xClaimJustId`, and verifies that the returned array of message IDs matches (`['1-0', '2-0']`). The test is also updated to run across both standalone client and cluster topologies via `testUtils.testAll`.

### Checklist

### DIFF
--- a/packages/client/lib/commands/XCLAIM_JUSTID.spec.ts
+++ b/packages/client/lib/commands/XCLAIM_JUSTID.spec.ts
@@ -11,15 +11,23 @@ describe('XCLAIM JUSTID', () => {
     );
   });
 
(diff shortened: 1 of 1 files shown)


nkaradzhov (approved):
Thanks @Piyush0049, looks good

PR #3378, opened 2026-07-29
fix(json): unwrap RESP3 array reply for JSON.TYPE

Piyush0049:
### Description

This pull request resolves an inline maintainer TODO in `packages/json/lib/commands/TYPE.ts` (`// TODO: RESP3 wraps the response in another array, but only returns 1`).

Under the RESP3 protocol, RedisJSON wraps JSONPath command replies in a top-level array. Previously, the transformer rigidly accessed `reply[0]`, which lacked runtime safety and could crash (`TypeError: Cannot read properties of null`) if the response was unexpectedly `null` or scalar. 

This change:
1. Implements safe unwrapping (`Array.isArray(reply) ? reply[0] : reply`).
2. Introduces a dedicated `transformReply` unit test suite in `TYPE.spec.ts` to verify the exact RESP3 behavior for single elements, mapped arrays, and non-existent keys.

---

### Checklist

---

> [!NOTE]
> **Low Risk**
> Narrow change to JSON.TYPE RESP3 parsing with added unit tests; no auth, persistence, or broad API surface impact.
> 
> **Overview**
> **JSON.TYPE** RESP3 reply handling no longer always indexes `reply[0]`. The transformer now unwraps RedisJSON’s top-level array only when the value is an array, and passes through scalars unchanged so `null` or unexpected shapes do not throw at runtime.
> 
> Unit tests were added for the RESP3 `transformReply` path: single-type unwrap, JSONPath multi-match (`[['string']]`), and missing keys (`[null]`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee7da8f2ce6d2c53a1ccb93b2a403ed8eadda450. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/json/lib/commands/TYPE.spec.ts
+++ b/packages/json/lib/commands/TYPE.spec.ts
@@ -22,6 +22,24 @@ describe('JSON.TYPE', () => {
     });
   });
 
+  describe('transformReply', () => {
+    describe('RESP3', () => {
+      type Resp3ReplyParam = Parameters<typeof TYPE.transformReply[3]>[0];
+
+      it('unwraps single element array response', () => {
+        assert.equal(TYPE.transformReply[3](['object'] as unknown as Resp3ReplyParam), 'object');
+      });
+
+      it('unwraps array of types when path is provided', () => {
+        assert.deepEqual(TYPE.transformReply[3]([['string']] as unknown as Resp3ReplyParam), ['string']);
+      });
+
+      it('returns null for non-existent key', () => {
+        assert.equal(TYPE.transformReply[3]([null] as unknown as Resp3ReplyParam), null);
+      });
+    });
+  });
+
   testUtils.testWithClient('client.json.type', async client => {
     assert.equal(
       await client.json.type('key'),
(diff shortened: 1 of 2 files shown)


nkaradzhov:
Hi @Piyush0049 ,thanks for the PR!

I looked into whether the reply[0] access can actually fail, and I don't think the reported crash is reachable on supported servers. Under RESP3, JSON.TYPE wraps its result in an array in every case we exercise — including the missing-key case. Our existing integration test (TYPE.spec.ts, client.json.type('key') on a non-existent key) already goes through the RESP3 transformer and asserts null; for the current return reply[0] to produce null there, the server must be returning [null] rather than a bare null. That test passes on master, which confirms the reply is always array-wrapped.

As a result, the new Array.isArray(reply) ? reply[0] : reply branch is unreachable given the transformer's input type (ArrayReply<...>), and none of the three added unit tests exercise the non-array path — their fixtures (['object'], [['string']], [null]) all pass against the current implementation unchanged. The array cases are also already covered by the existing real-client test.

PR #3381, opened 2026-07-29
feat(client): add HIMPORT command family with managed fieldset lifecycle

nkaradzhov:
This pull request adds first-class support for the Redis 8.10 `HIMPORT` command family (hinted hash templates): `HIMPORT PREPARE`, `HIMPORT SET`, `HIMPORT DISCARD`, and `HIMPORT DISCARDALL`, exposed as `hImportPrepare`/`hImportSet`/`hImportDiscard`/`hImportDiscardAll` and marked `@experimental`.

Fieldsets are server-side session state scoped to a single physical connection, so a naive client binding breaks under pools, reconnects, and clusters. Per the HIMPORT client HLD, the client owns the fieldset lifecycle: a shared `FieldsetRegistry` records registrations on the logical client (standalone, pool, cluster, sentinel), and a per-connection hook transparently pipelines `PREPARE` in front of `SET` on sessions that lack the fieldset or hold a stale field list, replays pending discards before dependent commands, and retries a `SET` once on `no such fieldset` for registered fieldsets. `DISCARD`/`DISCARDALL` replies are registry-based (registered-and-removed), giving deterministic semantics across topologies, and a rejected discard restores the session bookkeeping so the discard replays on that connection.

On a cluster, every serving node client runs the same hook, so sessions prepare and reconcile lazily per node with no HIMPORT-specific routing code. The eager `all_shards` fan-out for the session commands (server command tips) and its response-policy handling arrive with the request/response policies work (#2996), which upgrades cluster support from lazy-per-node to fully tip-compliant.

Behavior notes: raw `sendCommand(['HIMPORT', ...])` bypasses the managed layer; `duplicate()` shares the parent's registrations while `createPool()` does not; discarded fieldsets may linger on idle pooled connections (memory only, reconciled before their next HIMPORT command).

---

> [!NOTE]
> **High Risk**
> Large new command path with optimistic registry/session bookkeeping, injected wire commands, and rollback/reconcile logic across standalone, pool, cluster, and sentinel—errors could cause wrong hash field mapping or inconsistent session state.
> 
> **Overview**
> Adds **experimental** Redis 8.10 **HIMPORT** support (`hImportPrepare`, `hImportSet`, `hImportDiscard`, `hImportDiscardAll`) so hinted hash templates work across reconnects, pools, cluster nodes, and sentinel failovers.
> 
> A shared **`FieldsetRegistry`** on the logical client records prepared field lists; each connection keeps **`PreparedFieldsets`** for what that socket’s session actually holds. Typed HIMPORT calls go through **`#executeHimport`**, which **pipelines** lazy `PREPARE` and reconcile `DISCARD`/`DISCARDALL` ahead of `SET`, returns **registry-based** discard replies, **gates** `SET` on failed injected `PREPARE`, **retries `SET` once** on `no such fieldset` when still registered, and **re-issues `ASKING`** on cluster ASK chains (`askRedirect`). Session state is cleared on socket errors, `reset()`, and socket swap; **`duplicate()`** shares the registry; **pools/cluster/sentinel** inject one registry per topology (pools always get a **new** registry). **`MULTI`/pipeline** reject `PREPARE`/`DISCARD`/`DISCARDALL` client-side; raw `sendCommand(['HIMPORT', ...])` skips the managed layer.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 994ac8d7acc6e65c829f28771991566ef06a828c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -25,8 +25,34 @@ import { ClientMetricsHandle, ClientRegistry } from '../opentelemetry';
 import { ClientIdentity, ClientRole, generateClientId } from './identity';
 import { trace, sanitizeArgs, publish, CHANNELS, type CommandTraceContext } from './tracing';
 import { DEFAULT_COMMAND_TIMEOUT } from '../defaults';
+import { FieldsetRegistry, PreparedFieldsets } from '../himport/registry';
+import HIMPORT_DISCARD from '../commands/HIMPORT_DISCARD';
+import HIMPORT_DISCARDALL from '../commands/HIMPORT_DISCARDALL';
+import HIMPORT_PREPARE from '../commands/HIMPORT_PREPARE';
+import HIMPORT_SET from '../commands/HIMPORT_SET';
 
 const noop = () => {};
+
+const HIMPORT_SESSION_SUBCOMMANDS = new Set(['PREPARE', 'DISCARD', 'DISCARDALL']);
+
+/**
+ * MULTI/pipeline stores raw args only, so the HIMPORT transparency hook never sees these
+ * commands — a PREPARE/DISCARD executed that way would silently diverge the client registry
+ * from server state (a fieldset the registry doesn't know about, or a discarded one it
+ * would resurrect via lazy prepare). Rejected client-side at the exec funnel, which covers
+ * both the typed multi methods and raw `multi.addCommand(...)`. HIMPORT SET stays allowed:
+ * it mutates no registry state (the fieldset must already exist on the carrying connection).
+ */
+function assertNoHimportSessionCommands(commands: Array<RedisMultiQueuedCommand>) {
+  for (const { args } of commands) {
+    if (String(args[0]).toUpperCase() !== 'HIMPORT') continue;
+    if (HIMPORT_SESSION_SUBCOMMANDS.has(String(args[1]).toUpperCase())) {
+      throw new Error(
+        'HIMPORT PREPARE/DISCARD/DISCARDALL are not supported inside MULTI/pipeline; call them on the client before the transaction'
+      );
+    }
+  }
+}
 
 export interface RedisClientOptions<
   M extends RedisModules = RedisModules,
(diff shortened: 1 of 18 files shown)


packages/client/lib/client/index.ts
```diff
@@ -1244,6 +1300,209 @@ export default class RedisClient<
...
+        client.sendCommand(injection.args, injectOpts).catch(injection.rollback);
+      }
+      mainPromise = send();
+    }
+
+    let reply: unknown;
+    try {
+      reply = await mainPromise;
+    } catch (err) {
+      if (command === HIMPORT_PREPARE && userPrepare !== undefined) {
```
PavelPashov, line 1505: Could we restore the previous registry entry when a replacement `PREPARE` fails, removing it only when no prior definition existed? Redis retains the existing fieldset after validation failure, so deleting it here breaks transparent replay for duplicates and reconnects.

packages/client/lib/client/index.ts
```diff
@@ -1244,6 +1300,209 @@ export default class RedisClient<
...
+          prelude.push({
+            args: ['HIMPORT', 'PREPARE', name, ...entry.fields],
+            // Injected-PREPARE failures are often transient (LOADING, failover) — roll back
+            // only the session claim, never the registration, or a valid fieldset would
+            // permanently lose lazy re-prepare.
+            rollback: () => {
+              if (prepared.get(name) === entry.version) prepared.delete(name);
+            }
+          });
+        }
```
PavelPashov, line 1437: Could we tie the SET's outcome to its injected PREPARE? If the session holds an older version and the injected PREPARE fails while the connection stays alive (e.g. ACL denies the PREPARE subcommand), the SET still runs against the stale field list and succeeds — values land under the wrong field names with no error. One option: keep the PREPARE promise and, after the SET resolves, reject (or re-verify the version) if the PREPARE had failed.

packages/client/lib/client/index.ts
```diff
@@ -1244,6 +1300,209 @@ export default class RedisClient<
...
+      mainPromise = send();
+      for (let i = prelude.length - 1; i >= 0; i--) {
+        const injection = prelude[i];
+        client.sendCommand(injection.args, injectOpts).catch(injection.rollback);
+      }
+    } else {
+      for (const injection of prelude) {
+        client.sendCommand(injection.args, injectOpts).catch(injection.rollback);
+      }
+      mainPromise = send();
```
PavelPashov, line 1498: When this runs under an ASK redirect, the wire order becomes ASKING → injected PREPARE → SET, and the keyless PREPARE consumes the one-shot ASKING flag — the SET then gets MOVED again and loops until `maxCommandRedirections`. Would it make sense to re-send ASKING between the prelude and the main command when on an ASK chain (or prepare the target node eagerly in `_handleAsk`)?

PavelPashov (approved):

PR #3388, opened 2026-08-02
fix(sentinel): cap post-connect rediscovery retries

### LINKED ISSUE #3385: Sentinel client gets stuck connecting to an unreachable instance
### Description

.
## Description

`RedisSentinelClient#connect()`'s retry cap becomes dead code after the very first successful connection, so once a Sentinel client has connected once, it can retry Sentinel topology rediscovery *forever* on any later failure — silently, with no error ever emitted for the app's own `client.on('error', ...)` handler to react to.

Root cause, in `sentinel/index.js`:

if (!this.#isReady && count > this.#maxCommandRediscovers) {
  throw e;
}
await setTimeout(1000); // otherwise: sleep 1s and loop forever

`this.#isReady` is the Sentinel client's own top-level flag. It's set `true` once, after the first successful connect, and is only ever reset back to `false` in `.close()`. So on any *subsequent* topology-discovery failure (e.g. after the app has been running fine for hours and then the Sentinel topology becomes unreachable), `!this.#isReady` is always `false` — the `count > maxCommandRediscovers` cap can never trip, and `connect()` just sleeps 1s and loops indefinitely instead of ever throwing.

Net effect: a long-lived Sentinel client that hits a topology-discovery outage after its initial connect does not recover on its own, does not emit any lifecycle/error event, and does not honor the configured retry cap at all. From the outside the process looks "stuck" — no crash, no error log, just a client silently spinning on the topology-discovery loop.

Related but distinct from #3346 (fixed by #3374) — that one is about the *standalone* client hanging mid-async-suspension during a socket-death race. This is Sentinel-specific, purely in the post-initial-connect retry-cap logic, and reproduces deterministically once the client has connected at least once.

### Node.js Version

20.x (also seen on 22.x)

### Redis Server Version

7.x, Sentinel topology (3 sentinel + primary/replica set)

### Node Redis Version

@redis/client@6.0.0 — confirmed still present in 6.1.0 and 6.2.0-beta.

### Platform

Linux (Kubernetes pods), reproduced locally on macOS via Docker Compose Sentinel topology

### Logs

```bash
RedisCommandTimeoutError repeating indefinitely with no error/reconnecting/end event ever emitted, once the client has connected successfully at least once and a later topology-discovery attempt starts failing. count > maxCommandRediscovers never trips because #isReady is already true and is never reset outside of .close().
```

DebadityaHait:
Fixes #3385.

Apply maxCommandRediscovers to rediscovery after the Sentinel has already connected. This causes the operation waiting on failed rediscovery to reject once the cap is exhausted instead of waiting indefinitely.

Tests:
- npm run build
- npm run test:types -w @redis/client
- npm run lint

---

> [!NOTE]
> **Medium Risk**
> Changes Sentinel failover/rediscovery behavior for already-connected clients; mis-tuned caps could cause earlier command failures or background errors, but the fix aligns with documented limits and prevents indefinite hangs.
> 
> **Overview**
> Fixes **#3385** by enforcing **`maxCommandRediscovers`** on every topology rediscovery in `#connect()`, not only before the first successful connect. After a later full Sentinel outage, commands and background resets **reject or emit `error`** once the cap is hit instead of looping forever.
> 
> Periodic scan, pub/sub control messages, and sentinel client failures now trigger **`#resetInBackground()`**, which runs `#reset()` without leaving unhandled rejections and surfaces failures via **`error`**. **`close()`** / **`destroy()`** no longer block on a failing in-flight `#connectPromise` (they swallow rejection while waiting).
> 
> Regression tests cover **in-flight commands** after total outage with a low cap and **background** rediscovery with `maxCommandRediscovers: 0`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd90ebdb0381d4375b54d40a9ca50ded5e45699b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/sentinel/index.spec.ts
+++ b/packages/client/lib/sentinel/index.spec.ts
@@ -1363,6 +1363,40 @@ describe('legacy tests', () => {
       assert.equal(sawRecoverySuccess, true, 'expected periodic GET to recover after restart');
     });
 
+    // Regression test for https://github.com/redis/node-redis/issues/3385
+    //
+    // The rediscovery cap previously applied only before the first successful
+    // connect. A later full Sentinel outage left an operation waiting on the
+    // rediscovery loop forever.
+    it('rejects a command after rediscovery reaches its cap following a full outage', async function () {
+      this.timeout(30000);
+
+      sentinel = frame.getSentinelClient({
+        maxCommandRediscovers: 1,
+        commandOptions: { timeout: undefined }
+      });
+      sentinel.on('error', () => { });
+      await sentinel.connect();
+
+      await Promise.all([
+        ...frame.getAllNodesPort().map(port => frame.stopNode(port.toString())),
+        ...frame.getAllSentinelsPort().map(port => frame.stopSentinel(port.toString()))
+      ]);
+
+      const timeoutError = new Error('rediscovery cap was not enforced');
+      const result = await Promise.race([
+        sentinel.get('some-key').then(
+          () => new Error('command unexpectedly resolved'),
+          err => err
+        ),
+        setTimeout(5000).then(() => timeoutError)
+      ]);
+
+      assert.ok(result instanceof Error);
+      assert.notStrictEqual(result, timeoutError);
+      assert.notStrictEqual(result.message, 'command unexpectedly resolved');
+    });
+
     it('timer works, and updates sentinel list', async function () {
       this.timeout(60000);
 
--- a/packages/client/lib/sentinel/index.spec.ts
+++ b/packages/client/lib/sentinel/index.spec.ts
@@ -1397,6 +1397,27 @@ describe('legacy tests', () => {
       assert.notStrictEqual(result.message, 'command unexpectedly resolved');
     });
 
+    it('emits an error when background rediscovery reaches its cap', async function () {
+      this.timeout(30000);
+
+      sentinel = frame.getSentinelClient({
+        maxCommandRediscovers: 0,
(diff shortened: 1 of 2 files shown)


nkaradzhov (changes requested):
Thanks @DebadityaHait, before we can merge, the change needs to handle the side effects of #connect() now being able to reject after the client is ready:

1. Unhandled rejections. #reset() is called fire-and-forget in three places — #handlePubSubControlChannel, #handleSentinelFailure, and the scan-interval timer set up in connect(). Previously #connect() never rejected post-ready, so these were safe. With this change, during a full outage these background resets will reject with no handler, which on supported Node versions can terminate the process. Please attach a handler (e.g. surface it via the client error event) to each of these call sites.
3. Teardown during rediscovery. close() and destroy() do a bare await this.#connectPromise. If that promise now rejects mid-rediscovery, the await throws and the rest of teardown (clearing #scanTimer, closing the sentinel/master/replica clients) is skipped, leaking timers and sockets. Please wrap those awaits so teardown always runs to completion.

DebadityaHait:
Thanks @nkaradzhov, addressed in `dd90ebd`: background resets now catch and surface capped rediscovery failures; `close()` and `destroy()` complete cleanup if rediscovery rejects; retry counting now advances only on failed attempts. Added coverage for the background-error path as well.

PR #3391, opened 2026-08-05
fix(client): dispose credentials subscription on every close() path

### LINKED ISSUE #3390: close() skips disposing the credentials subscription when the command queue is empty, leaking the entraid token-refresh timer
### Environment

- **Node.js Version**: 24.x
- **Redis Server Version**: Azure Managed Redis (Enterprise cluster policy)
- **Node Redis Version**: 5.12.1 (bug also present on current master, permalinks below)
- **Platform**: Linux / macOS

### Description

When a client is constructed with a `credentialsProvider` from `@redis/entraid`, `client.close()` only disposes the credentials subscription on one of its two exit paths. The `#credentialsSubscription?.dispose()` call sits **after** the empty-queue early return:

https://github.com/redis/node-redis/blob/2215b5d48d1d932d0cdf6e32010f643c684501db/packages/client/lib/client/index.ts#L2185-L2209

- **Queue not empty**: the code falls through to `#credentialsSubscription?.dispose()` — subscription disposed correctly.
- **Queue empty** (the normal case for a graceful shutdown — all commands already settled): `close()` returns at the early `return resolve()` and the credentials subscription is never disposed.

The undisposed subscription leaves the entraid `TokenManager` refresh timer armed, which keeps the event loop alive — the process hangs after `close()` resolves instead of exiting.

`destroy()` disposes the subscription unconditionally, which is the behaviour `close()` presumably intends on both paths:

https://github.com/redis/node-redis/blob/2215b5d48d1d932d0cdf6e32010f643c684501db/packages/client/lib/client/index.ts#L2214-L2222

### Reproduction

```ts
import { createClient } from 'redis';
import { EntraIdCredentialsProviderFactory, REDIS_SCOPE_DEFAULT } from '@redis/entraid';
import { DefaultAzureCredential } from '@azure/identity';

const client = createClient({
  url: 'rediss://<amr-instance>.redis.azure.net:10000',
  RESP: 3,
  credentialsProvider: EntraIdCredentialsProviderFactory.createForDefaultAzureCredential({
    credential: new DefaultAzureCredential(),
    scopes: REDIS_SCOPE_DEFAULT,
    tokenManagerConfig: { expirationRefreshRatio: 0.8 },
  }),
});

await client.connect();
await client.ping();      // all commands settle — queue is now empty
await client.close();     // resolves, but the token-refresh timer stays armed
// process does not exit
```

Swapping `close()` for `destroy()` (or calling `destroy()` after `close()`) lets the process exit, confirming the credentials subscription is what holds the loop.

### Expected behaviour

`close()` disposes the credentials subscription on every exit path, as `destroy()` does — a graceful close of an idle client should release everything the client owns.

### Workaround

Quiesce then destroy: `await client.close()` (drains pending replies) followed by `client.destroy()` (releases the credentials subscription). We run this in production against Azure Managed Redis with Entra ID auth; it behaves correctly.

lazerg:
### Description

Fixes #3390.

`close()` only disposed the streaming credentials subscription on the path where the command queue still had pending commands. On a graceful shutdown of an idle client the queue is already empty, so `close()` hit the early `return resolve()` and left the subscription alive. With `@redis/entraid` that keeps the token-refresh timer armed and the process never exits.

The dispose call now runs before the empty-queue early return, so both paths release the subscription, matching what `destroy()` already does.

The new test connects a client with a streaming credentials provider to a mock TCP server, closes it with an empty queue, and asserts the subscription was disposed.

---

### Checklist

---

> [!NOTE]
> **Low Risk**
> Small lifecycle fix aligned with `destroy()`/`quit()`; behavior change only affects graceful close with streaming auth when the queue is empty.
> 
> **Overview**
> **`close()`** now disposes the streaming **`credentialsSubscription`** at the start of shutdown, before the command-queue empty check. Previously that cleanup only ran on the branch that waited for pending replies, so an idle client could exit **`close()`** without calling **`dispose()`**—leaving token-refresh timers alive (e.g. with **`@redis/entraid`**) and preventing process exit.
> 
> A regression test connects with a streaming credentials provider, **`close()`**s with an empty queue, and asserts **`dispose`** ran. **`countRespCommands`** was hoisted to module scope for shared mock-server use.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1af6ea8dccf86b22252adf9e6a80f3b1ee74524e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -1542,18 +1542,6 @@ describe('Client', () => {
       return log;
     }
 
-    function countRespCommands(chunk: Buffer): number {
-      let commands = 0;
-
-      for (let i = 0; i < chunk.length; i++) {
-        if (chunk[i] === 42 && (i === 0 || chunk[i - 1] === 10)) {
-          commands++;
-        }
-      }
-
-      return commands;
-    }
-
     // Create a TCP server that accepts connections but immediately drops them <dropImmediately> times.
     // For accepted connections, reply with one `+OK` per incoming RESP command.
     function setupMockServer(dropImmediately: number) {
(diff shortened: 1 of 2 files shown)


nkaradzhov (approved):
@lazerg thanks, this looks good to go!

Issue #3397, opened 2026-08-14
Pool: failed scale-up connect() is an unhandled rejection, hides real error behind TimeoutError


kyungseopk1m:
### Description

`RedisClientPool.execute()` scales the pool up by calling `#create()` without awaiting or catching it:

```ts
// packages/client/lib/client/pool.ts:483-485
if (this.totalClients < this._self.#options.maximum) {
  this._self.#create();
}
```

`#create()` re-throws when `client.connect()` fails (pool.ts:443-448), so this becomes an unhandled rejection. Compare with `connect()`, in the same file, which calls the same `#create()` but properly awaits it inside try/catch (pool.ts:419-429):

```ts
const promises = [];
while (promises.length < this._self.#options.minimum) {
  promises.push(this._self.#create());
}
try {
  await Promise.all(promises);
} catch (err) {
  this.destroy();
  throw err;
}
```

Effect on the caller: with no idle client available, `execute()` pushes the task onto `#tasksQueue` and returns without ever finding out the scale-up connect failed. The task just sits there until `acquireTimeout` elapses, so the caller gets a `TimeoutError` instead of the real connection error, and with `acquireTimeout: 0` the task never resolves at all.

Repro (published `redis@6.2.1`, no server needed, port 1 is never listening):

```js
const { createClientPool } = require('redis');
process.on('unhandledRejection', e => console.log('[unhandledRejection]', e.constructor.name, e.message));

const pool = createClientPool(
  { socket: { host: '127.0.0.1', port: 1, reconnectStrategy: false } },
  { minimum: 0, maximum: 4, acquireTimeout: 500 }
);
await pool.connect(); // resolves immediately, minimum is 0
await pool.ping();
```

Output:

```
[unhandledRejection] Error connect ECONNREFUSED 127.0.0.1:1
execute() rejected after 502 ms with TimeoutError Timeout waiting for a client after 500ms
```

With `acquireTimeout: 0` the `ping()` call above just hangs forever instead of rejecting, and `tasksQueueLength` stays at 1.

The obvious fix is to have `execute()`'s `#create()` call reject the queued task it triggered instead of letting it dangle. But that means a second call site starts removing tasks from `#tasksQueue`, and the removal API isn't safe for that:

```ts
// packages/client/lib/client/pool.ts:459-461, 467
const client = this._self.#idleClients.shift(),
  { tail } = this._self.#tasksQueue;
...
this._self.#tasksQueue.remove(task, tail);
```

[...]

PR #3400, opened 2026-08-14
fix(client): settle the close() drain when the last connect fails

### LINKED ISSUE #3398: Pool: close() hangs forever if a scale-up connect fails while draining
### Description

`RedisClientPool.close()` waits for `#clientsInUse` to drain before closing idle clients:

```ts
// packages/client/lib/client/pool.ts:584-590
if (this._self.#clientsInUse.length > 0) {
  await new Promise<void>(resolve => {
    this._self.#drainResolve = resolve;
  });
}
```

The only place that resolves `#drainResolve` is `#returnClient()`, when `#clientsInUse` reaches 0 (pool.ts:528-531). But `#create()` pushes a node onto `#clientsInUse` before it awaits `client.connect()` (so it counts toward capacity while connecting), and on a connect failure it only removes the node, it never goes through `#returnClient()`:

```ts
// packages/client/lib/client/pool.ts:441-448
const node = this._self.#clientsInUse.push(client);
try {
  await client.connect();
} catch (err) {
  this._self.#clientsInUse.remove(node);
  throw err;
}
```

So if `close()` starts while a scale-up connect is in flight (the only client currently in `#clientsInUse`), and that connect then fails, `#clientsInUse` drops to 0 without the drain check ever running, and `close()` hangs forever.

Repro (published `redis@6.2.1`, port 1 is never listening):

```js
const { createClientPool } = require('redis');
const pool = createClientPool(
  { socket: { host: '127.0.0.1', port: 1, reconnectStrategy: false } },
  { minimum: 1, maximum: 1, acquireTimeout: 500 }
);

const connectPromise = pool.connect();
connectPromise.catch(() => {}); // expected to reject, just don't want it unhandled
console.log('clientsInUse right after connect() is called:', pool.clientsInUse);

pool.close().then(
  () => console.log('close() resolved'),
  err => console.log('close() rejected', err.message)
);

setTimeout(() => console.log('still hung after 2s'), 2000);
```

Output:

```
clientsInUse right after connect() is called: 1
still hung after 2s
```

`close()` never settles.

### Node.js Version

v22.16.0

### Redis Server Version

n/a, reproduces without a reachable Redis server

### Node Redis Version

6.2.1

### Platform

macOS (darwin), also inspected against `packages/client/lib/client/pool.ts` in the repo directly

### Logs

```bash
clientsInUse right after connect() is called: 1
still hung after 2s
```

kyungseopk1m:
### Description

Fixes #3398.

`#create()` pushes the client onto `#clientsInUse` before it awaits `client.connect()`, so a connecting client counts toward capacity. On a connect failure it removes the node directly instead of going through `#returnClient()`, which is the only place that resolves `#drainResolve`. If `close()` was already waiting for the pool to drain, `#clientsInUse` reached 0 without waking it and `close()` never settled.

Signalling the drain from that path is not enough on its own. A task queued behind the pending connect is exactly what `close()` is still waiting for, and once the connect fails there is no client left to run it on: `execute()` rejects new work while closing, and nothing will call `#returnClient()` again. So the failure path now settles those tasks with the connection error before it resolves the drain. The caller gets the real reason instead of a later `TimeoutError`, or nothing at all when `acquireTimeout` is 0.

This is scoped to a closing pool. While the pool is open a failed scale-up leaves the queued task on its existing `acquireTimeout` path, which is #3397 and a separate discussion. Draining the queue here only uses `shift()`, so it does not touch `SinglyLinkedList.remove()`.

Two tests cover it, neither needs a server:

- `close()` settles when the only pending connect fails
- a task queued behind that connect is rejected with the connection error, `close()` settles, and the task's acquire timer is cleared with it (otherwise it later fires on a node that is already out of the queue and `remove()` throws)

---

### Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches pool shutdown and task-queue settlement on connect failure; behavior is scoped to closing pools but affects promise lifecycle and timers.
> 
> **Overview**
> Fixes a **hang in `RedisClientPool.close()`** when shutdown overlaps a failing `connect()`: `#create()` removed the in-flight client from `#clientsInUse` on error without invoking `#returnClient()`, so `#drainResolve` never ran even though the pool was empty.
> 
> On that path, when the pool is **closing** and **no clients remain in use**, the failure handler now **drains `#tasksQueue`**—`clearTimeout` on each waiter, **reject with the connect error**, then **call `#drainResolve`**—so queued `execute()` work fails with the real connection error instead of waiting on `acquireTimeout`, and `close()` can finish.
> 
> Adds two **no-server** regression tests: `close()` resolves after the only pending connect fails, and a task queued behind that connect is rejected, timers are cleared, and `close()` still completes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 274ffd39e7ae62b383979c2482a714e5dba1334b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/client/pool.spec.ts
+++ b/packages/client/lib/client/pool.spec.ts
@@ -160,6 +160,59 @@ describe('RedisClientPool', () => {
     poolOptions: { minimum: 1, maximum: 1, acquireTimeout: 2000, cleanupDelay: 400  }
   });
 
+  it('close resolves when the last pending connect fails', async () => {
+    // Regression: on a failed connect `#create()` removed the client from `#clientsInUse`
+    // directly instead of going through `#returnClient()`, the only place that resolves the
+    // drain promise. `#clientsInUse` reached 0 without waking `close()`, which hung forever.
+    // Nothing listens on port 1 here, so the connect always fails.
+    const pool = RedisClientPool.create(
+      { socket: { host: '127.0.0.1', port: 1, reconnectStrategy: false } },
+      { minimum: 1, maximum: 1, acquireTimeout: 500 }
+    );
+
+    const connectPromise = pool.connect();
+    assert.equal(pool.clientsInUse, 1, 'a connecting client should count as in use');
+
+    const closePromise = pool.close().then(() => 'closed');
+    const hangGuard = new Promise<string>(resolve => {
+      setTimeout(resolve, 1000, 'hung').unref();
+    });
+
+    await assert.rejects(connectPromise);
+    assert.equal(await Promise.race([closePromise, hangGuard]), 'closed');
+  });
+
+  it('close rejects the tasks it can no longer serve when the connect fails', async () => {
+    // A task queued behind the pending connect is the one thing `close()` is still waiting
+    // for. Once that connect fails there is no client left to run it on, so it has to be
+    // settled with the connection error rather than left sitting until `acquireTimeout`.
+    const acquireTimeout = 200;
+    const pool = RedisClientPool.create(
+      { socket: { host: '127.0.0.1', port: 1, reconnectStrategy: false } },
+      { minimum: 1, maximum: 1, acquireTimeout }
+    );
+
+    const connectPromise = pool.connect();
+    const connectRejects = assert.rejects(connectPromise);
+    const taskRejects = assert.rejects(pool.execute(() => 'never runs'), /ECONNREFUSED/);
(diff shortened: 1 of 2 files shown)


nkaradzhov (approved):
Thanks @kyungseopk1m, this looks good to go

PR #3403, opened 2026-08-14
fix(client): recreate cluster node client after a terminal connect failure

### LINKED ISSUE #3396: Cluster node client stays cached after terminal reconnectStrategy failure, and clearing it on failure leaks clients
## Summary

`RedisClusterSlots#createNodeClient` assigns `node.client` before `connect()` resolves. If `connect()` rejects because the socket's `reconnectStrategy` gives up (returns `false` or an `Error`), the client stays cached with `isOpen === false`. `nodeClient()` only creates a new client when `node.client` is falsy, so every later call for that node returns the same dead client instead of retrying.

```js
// packages/client/lib/cluster/cluster-slots.ts
#createNodeClient(node, readonly) {
  const client = node.client = this.#createClient(node, readonly);
  return node.connectPromise = client.connect()
    .finally(() => node.connectPromise = undefined);
}
```

The only place that clears `node.client` is the topology cleanup loop in `#discover`, and it only runs for addresses that have left the cluster topology entirely. A node that is still part of the topology but whose client has terminally failed is never cleared.

## Reproduction

With `redis@6.2.1`, reaching into `cluster-slots.js` to call `nodeClient()` directly against a closed port (no server needed to reproduce):

```js
const RedisClusterSlots = require('@redis/client/dist/lib/cluster/cluster-slots.js').default;

const slots = new RedisClusterSlots({
  rootNodes: [{ socket: { host: '127.0.0.1', port: 39999 } }],
  defaults: {
    socket: {
      port: 39999,
      reconnectStrategy: (retries, cause) => new Error('giving up after first failure'),
    },
  },
}, () => {});

const node = { id: 'node-a', host: '127.0.0.1', port: 39999, address: '127.0.0.1:39999', readonly: false };

await slots.nodeClient(node).catch(err => console.log('call #1 rejected:', err.message));
console.log('node.client cached:', !!node.client, 'isOpen:', node.client.isOpen);

const before = node.client;
const result = await slots.nodeClient(node).catch(() => node.client);
console.log('call #2 returned the same dead client, no reconnect attempted:', result === before);
```

Output:

```
call #1 rejected: giving up after first failure
node.client cached: true isOpen: false
call #2 returned the same dead client, no reconnect attempted: true
```

`reconnectStrategy: retries => retries > N ? new Error(...) : delay` is a common production setting, this is not an edge case.

## Why clearing node.client on failure makes it worse

The obvious patch is to clear `node.client` in a `.catch()` on the connect promise. That fixes the symptom above, but a terminally-failed client can't be disposed of cleanly. `RedisSocket#destroy()` throws `ClientClosedError` when `isOpen` is already `false`, and `RedisClient#destroy()` calls `this.#socket.destroy()` before unregistering from the metrics registry and clearing credentials (`packages/client/lib/client/index.ts`, `destroy()`). So on a terminal-failed client, `destroy()` throws before any of that runs.

With the stock code this is a one-client leak: the dead client stays cached forever, so it only leaks once. If `node.client` is cleared on failure instead, every retry creates a brand new client and abandons the previous one, still registered, still holding its socket-level `error`/`ready`/`end`/`__MOVED` listeners, without ever reaching cleanup.

Measured with OpenTelemetry enabled (`ClientRegistry.init()`), same node, 6 calls to `nodeClient()`:

Baseline, stock `cluster-slots.ts`:
```
call #1: registry size = 1
call #2: registry size = 1
call #3: registry size = 1
call #4: registry size = 1
call #5: registry size = 1
call #6: registry size = 1
unique clientIds registered: 1
destroy() on the cached dead client threw: ClientClosedError - The client is closed
```

With `node.client = undefined` added in the `.catch()`:
```
call #1: registry size = 1
call #2: registry size = 2
call #3: registry size = 3
call #4: registry size = 4
call #5: registry size = 5
call #6: registry size = 6
unique clientIds registered: 6
```

Each abandoned client still has its `error`, `ready`, `end` and `__MOVED` listeners attached, so it also keeps a strong reference to the cluster instance.

## What's needed

The stuck-cache problem needs a disposal path for a terminal-failed client that does not throw, so metrics unregistration, credential cleanup and listener teardown can still run regardless of `isOpen`. `destroy()` currently assumes the client was open at some point. Curious what direction you'd want here before anyone builds a fix that trades one leak for a worse one.

### Node.js Version

22.16.0

### Redis Server Version

n/a, reproduces without a server (connect fails against a closed local port)

### Node Redis Version

6.2.1

### Platform

macOS (Darwin 24.6.0, arm64)

### Logs

See reproduction output above.

nkaradzhov:
Closes #3396.

### Background

In cluster mode `RedisClusterSlots#createNodeClient` assigns `node.client` before the connect promise settles. When `reconnectStrategy` gives up (returns `false`/`Error`), the connect rejects but the dead client stayed in `node.client`. Since `nodeClient()` only creates a new client when `node.client` is falsy, it then kept returning the same non-functional client forever — a node could never recover until a topology refresh happened to drop its address.

Clearing `node.client` naively would leak: the abandoned client stays in the global `ClientRegistry` with its `error`/`ready`/`end`/`__MOVED` listeners attached (holding a strong ref back to the cluster), and `RedisClient.destroy()` throws `ClientClosedError` on an already-closed socket, so the dead client could not be disposed.

### Change

- `RedisSocket.destroy()` is now idempotent — it no longer throws when the socket is already closed, so a terminally-failed client can be disposed and its metrics registration and credentials subscription are released instead of leaking. `RedisSocket` is internal (users call `client.destroy()`), and this only relaxes behavior — `destroy()` was previously the sole path that threw `ClientClosedError` on a closed client — so it is not a breaking change.
- `#createNodeClient` now catches a connect rejection, clears `node.client` (guarded on identity so an in-flight `#discover` swap is never clobbered), destroys the dead client, and rethrows so callers still see the real error. The next `nodeClient()` call retries with a fresh client.

### Notes

- **Config-conditional.** This only triggers when the connect can terminally reject: a `reconnectStrategy` returning `false`/`Error`, or a configured `socketTimeout` (whose idle timeout the default strategy declines to retry). The default `reconnectStrategy` retries with exponential backoff forever and never reaches this path, so default-configured clients are unaffected.
- **Scope.** This fixes the cluster node-client caching/disposal path. A standalone client also leaves its `ClientRegistry` entry registered after a terminal connect failure (nothing calls `destroy()` on that path); that is a separate latent issue and is intentionally not addressed here.
- Regression test (`cluster-slots.spec.ts`, no server required) drives a node at a dead address with `reconnectStrategy: false` and asserts the second `nodeClient()` retries rather than returning the cached dead client, and that the failure surfaces the real connect error rather than a `ClientClosedError` from a non-idempotent `destroy()`. It fails if either half of the fix is reverted.

---

> [!NOTE]
> **Medium Risk**
> Touches cluster connection lifecycle, socket destroy semantics, and RESP3 pooled client-side cache on teardown—important paths but narrowly scoped with regression tests.
> 
> **Overview**
> Fixes **#3396**: cluster shard clients that fail to connect terminally were left cached on `node.client`, so later `nodeClient()` calls kept returning a dead connection until topology refresh.
> 
> **Cluster:** On connect rejection, `#createNodeClient` now clears `node.client` (only if it still refers to that client), **`destroy()`s** the failed client, and rethrows the real error so the next `nodeClient()` builds a fresh client.
> 
> **Socket / client teardown:** `RedisSocket.destroy()` is **idempotent** (no `ClientClosedError` when already closed), so cleanup after a failed connect can finish. `close()` / `destroy()` only invoke **`clientSideCache.onClose()`** when `socketEpoch > 0`, so disposing a never-connected node does not flush a **shared pooled** client-side cache used by healthy connections.
> 
> Regression tests cover cluster retry behavior and pooled-cache preservation on `destroy()` without connect.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb129662ddcbb2296722f8779b5d171f26237678. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/client/socket.ts
+++ b/packages/client/lib/client/socket.ts
@@ -466,10 +466,13 @@ const retryIn = strategy(retries, cause);
   }
 
   destroy() {
-    if (!this.#isOpen) {
-      throw new ClientClosedError();
-    }
-
+    // Idempotent: return instead of throwing when already closed. A terminal
+    // connect failure (reconnectStrategy gave up) leaves #isOpen === false, and
+    // the owning client still needs to dispose itself (unregister metrics,
+    // dispose credentials) — throwing here would abort that cleanup. Returning
+    // also means a repeated destroy() won't re-run destroySocket() and
+    // republish CONNECTION_CLOSED / re-emit 'end'.
+    if (!this.#isOpen) return;
     this.#isOpen = false;
     this.destroySocket();
   }
--- a/packages/client/lib/client/cache.spec.ts
+++ b/packages/client/lib/client/cache.spec.ts
@@ -3,8 +3,28 @@ import testUtils, { GLOBAL } from "../test-utils"
 import { BasicClientSideCache, BasicPooledClientSideCache, CacheStats } from "./cache"
 import { REDIS_FLUSH_MODES } from "../commands/FLUSHALL";
 import { once } from 'events';
+import RedisClient from "./index";
 
 describe("Client Side Cache", () => {
+  it("destroy() on a never-connected client does not flush a shared pooled cache (#3396)", () => {
+    // Cluster/pool/sentinel share one pooled cache across node clients. Disposing
+    // a client whose connect failed terminally must not wipe entries the healthy
+    // connections are still using (the dead client cached nothing / missed no
+    // invalidations). It must also not throw — the owning client still needs to
+    // finish its own cleanup (unregister metrics, dispose credentials).
+    const cache = new BasicPooledClientSideCache({ maxEntries: 100 });
+    const client = RedisClient.create({ RESP: 3, clientSideCache: cache });
+
+    // Seed the shared cache as if a healthy pooled connection had cached a key.
+    cache.set("k", cache.createValueEntry(client as never, "v"), ["k"]);
+    assert.ok(cache.get("k"));
+
+    // Never connected (socketEpoch 0): idempotent, no throw, cache untouched.
+    assert.doesNotThrow(() => client.destroy());
+    assert.doesNotThrow(() => client.destroy());
+    assert.ok(cache.get("k"), "a never-connected client must not flush the shared cache");
+  });
+
   describe('Basic Cache', () => {
     const csc = new BasicClientSideCache({ maxEntries: 10 });
 
@@ -306,13 +326,13 @@ describe("Client Side Cache", () => {
 
       let p: Promise<Array<string>> = once(csc, 'invalidate');
       await client.set("x", 2);
-      let [i] = await p;
+      await p;
 
       assert.equal(await client.get("x"), '2', 'second get');
 
       p = once(csc, 'invalidate');
       await client.set("x", 3);
-      [i] = await p;
+      await p;
 
       assert.equal(await client.get("x"), '3');
 
--- a/packages/client/lib/client/index.ts
+++ b/packages/client/lib/client/index.ts
@@ -2190,7 +2190,10 @@ export default class RedisClient<
     return new Promise<void>(resolve => {
       clearTimeout(this._self.#pingTimer);
       this._self.#socket.close();
-      this._self.#clientSideCache?.onClose();
+      // See destroy(): only flush the cache if this client ever connected.
+      if (this._self.#socket.socketEpoch > 0) {
+        this._self.#clientSideCache?.onClose();
+      }
       this._self.#credentialsSubscription?.dispose();
       this._self.#credentialsSubscription = null;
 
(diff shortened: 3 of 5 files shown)


packages/client/lib/client/socket.ts
```diff
@@ -466,10 +466,9 @@ const retryIn = strategy(retries, cause);
   }
 
   destroy() {
```
PavelPashov, line 468: Could we ensure `destroySocket()` emits `CONNECTION_CLOSED` and `end` only once per connection lifecycle, resetting that guard on `connect()`? Repeated `destroy()` calls currently trigger duplicate cleanup handlers and telemetry despite being described as idempotent.

PavelPashov (approved):

PR #3407, opened 2026-08-21
fix(client): settle argument-less unsubscribe while patterns remain

kyungseopk1m:
### Description

An argument-less `unsubscribe()` or `pUnsubscribe()` hangs when a subscription of the other type is still active, and it misaligns later replies.

The completion check for an argument-less unsubscribe waited for the reply count to reach 0. Redis reports the remaining channel plus pattern count on each UNSUBSCRIBE or PUNSUBSCRIBE reply, and sharded subscriptions are tracked separately, so with a pattern still subscribed the count never reaches 0. The command hangs, and when the server confirmation finally arrives it is matched to the next queued command, so a following PING receives the unsubscribe reply.

This settles the command when the reported count reaches the number of remaining subscriptions of the opposite type instead of 0. Argument-specified unsubscribe and sharded paths keep their previous behavior.

Reproduction, against a real server on both RESP2 and RESP3:

```
await client.subscribe('ch', () => {});
await client.pSubscribe('pat*', () => {});
await client.unsubscribe();      // never resolves before this change
await client.ping('a');          // returns the unsubscribe reply before this change
```

Added tests for both directions (channels left after `pUnsubscribe()`, patterns left after `unsubscribe()`) on RESP2 and RESP3. They use an asymmetric channel and pattern count so that using the wrong subscription type's count is caught as well.

### Checklist

- [ ] Does `npm test` pass with this change (including linting)? Type check and lint pass. The full mocha suite spins its own Redis test container per case, which did not come up within the harness timeout on my machine, so I verified the change against a local server with the reproduction above rather than the full suite.
- [ ] Is a documentation update included? N/A, this is a bug fix with no API change.

---

> [!NOTE]
> **Medium Risk**
> Touches Pub/Sub reply matching in the command queue; a wrong residual would hang unsubscribes or steal later command replies. Scope is small and covered by new tests.
> 
> **Overview**
> Fixes argument-less `unsubscribe()` / `pUnsubscribe()` hanging (and misaligning later replies) when the other subscription type is still active.
> 
> Redis reports remaining **channel + pattern** count on those replies, so the old `=== 0` completion check never fired. The queue now treats the command as complete when the reported count equals remaining subscriptions of the *other* type (`residualAfterUnsubscribeAll`). Sharded unsubscribes still complete at 0.
> 
> Tests cover both directions on RESP2 and RESP3, with asymmetric channel vs pattern counts so the wrong residual would fail.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a9f3c3140575c695f748f8c8426a263508b056d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/client/commands-queue.ts
+++ b/packages/client/lib/client/commands-queue.ts
@@ -201,7 +201,8 @@ export default class RedisCommandsQueue {
     } else if (isShardedUnsubscribe || PubSub.isStatusReply(push)) {
       const head = this.#waitingForReply.head!.value;
       if (
-        (Number.isNaN(head.channelsCounter!) && push[2] === 0) ||
+        (Number.isNaN(head.channelsCounter!) &&
+          push[2] === this.#pubSub.residualAfterUnsubscribeAll(push)) ||
         --head.channelsCounter! === 0
       ) {
         this.#waitingForReply.shift()!.resolve();
--- a/packages/client/lib/client/index.spec.ts
+++ b/packages/client/lib/client/index.spec.ts
@@ -1135,6 +1135,9 @@ describe('Client', () => {
         // Following commands must receive their own replies (no off-by-one misalignment).
         assert.equal(await subscriber.ping('AAA'), 'AAA');
         assert.equal(await subscriber.ping('BBB'), 'BBB');
+
+        // Leave subscriber mode so the harness teardown (flushAll) is accepted on RESP2.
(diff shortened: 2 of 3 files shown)


PavelPashov (approved):
Thank you for the contribution. Looks good!

PR #3409, opened 2026-08-22
fix(client): name all members of StreamMessagesRawReply tuple

### LINKED ISSUE #3009: TS Build Error: TS5084: Tuple members must all have names or all not have names
I got this error after trying to build my project with Redis as a dependency: 

```
$ tsc
node_modules/@redis/client/dist/lib/commands/generic-transformers.d.ts:196:67 - error TS5084: Tuple members must all have names or all not have names.

196 type StreamMessagesRawReply = TuplesReply<[name: BlobStringReply, ArrayReply<StreamMessageRawReply>]>;
                                                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in node_modules/@redis/client/dist/lib/commands/generic-transformers.d.ts:196

error Command failed with exit code 2.
```
From looking at the source code, it looks like the ArrayReply part of the tuple should have a name of `messages`? 
https://github.com/redis/node-redis/blob/6f3380ba68786a3d5a9825f9a29c2053e43d32e3/packages/client/lib/commands/generic-transformers.ts#L553

abhijeet117:
## Summary
Fixes #3009. The emitted declaration for StreamMessagesRawReply mixed a labeled tuple member with an unlabeled one, so consumer builds using TypeScript versions that enforce TS5084 failed inside generic-transformers.d.ts. Both members are now labeled, matching StreamMessageRawReply and neighboring stream reply types.

## Testing
Compiled a consumer project against the built dist with tsc 5.0.4 and declaration emit: TS5084 at generic-transformers.d.ts(255,67) before the change, clean after. Removed the AST-walking spec per review since it parsed a source file for a narrow type-label fix. Ran test:types, the client unit suite, lint and build.

## Checklist
- bug reproduced before fix
- root cause identified
- bug fixed
- tests passed

### DIFF
--- a/packages/client/lib/commands/generic-transformers.spec.ts
+++ b/packages/client/lib/commands/generic-transformers.spec.ts
@@ -1,4 +1,6 @@
 import { strict as assert } from 'node:assert';
+import { readFileSync } from 'node:fs';
+import * as ts from 'typescript';
 import { BasicCommandParser } from '../client/parser';
 import { pushScanArguments } from './SCAN';
 import { parseGeoSearchArguments, parseGeoSearchOptions } from './GEOSEARCH';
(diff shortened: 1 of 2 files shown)


packages/client/lib/commands/generic-transformers.spec.ts
```diff
@@ -771,4 +773,36 @@ describe('Generic Transformers', () => {
...
+                        .map((element) => ts.isNamedTupleMember(element));
+                    if (named.includes(true) && named.includes(false)) {
+                        offenders.push(source.getLineAndCharacterOfPosition(node.getStart()).line + 1);
+                    }
+                }
+                ts.forEachChild(node, visit);
+            })(source);
+            assert.deepEqual(offenders, []);
+        });
+    });
```
PavelPashov: Could we remove this test? It introduces a new testing pattern by hardcoding and parsing a source file, which adds disproportionate maintenance cost for a narrow, low-risk type-label fix.

PavelPashov (changes requested):
The production change is correct and appropriately scoped. Once the AST test is removed, this looks ready to merge.

Seems to be related to #3009, please link the issue in the description

PavelPashov (approved):
Looks good

PR #3411, opened 2026-08-22
fix(client): correct HELLO modules reply type

abhijeet117:
## Summary
HELLO declared each modules entry as Array<BlobStringReply>, but the server sends one structured entry per loaded module. This declares the entries with the same TuplesToMapReply shape MODULE LIST uses, matching addReplyLoadedModules in the server. path and args exist only on Redis 7.0+, so like MODULE LIST the type narrows to name and ver and leaves newer fields out of the declared shape. The RESP2 path still passes modules through raw (flat arrays), so parity with MODULE LIST holds only for RESP3 mapped replies.

## Testing
Reproduced at compile time: with the old declaration each module entry resolved to string and accessing name or ver failed tsc. Added packages/client/types-tests/hello.types-test.ts, which fails before the fix and passes after. Ran test:types, lint and build locally; full suites run in CI.

## Checklist
- bug reproduced before fix
- root cause identified
- bug fixed
- tests passed

### DIFF
--- a/packages/client/lib/commands/HELLO.ts
+++ b/packages/client/lib/commands/HELLO.ts
@@ -17,7 +17,14 @@ export type HelloReply = TuplesToMapReply<[
   [BlobStringReply<'id'>, NumberReply],
   [BlobStringReply<'mode'>, BlobStringReply],
   [BlobStringReply<'role'>, BlobStringReply],
-  [BlobStringReply<'modules'>, ArrayReply<BlobStringReply>]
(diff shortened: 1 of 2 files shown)


nkaradzhov (changes requested):
Requesting changes: make `path`/`args` optional (they exist only on Redis 7.0+) or narrow to `name`/`ver` like MODULE LIST, and fix the parity claim (RESP2 still passes `modules` through raw). Details in my comment above.

nkaradzhov (approved):

PR #3417, opened 2026-08-23
fix(client): type ACL GETUSER flags as RESP3 SET

abhijeet117:
## Summary
ACL GETUSER declares flags as an array, but acl.c emits the flags element with setDeferredSetLen, so RESP3 replies carry a set that decodes to a plain Array unless callers opt in with a { [RESP_TYPES.SET]: Set } type mapping. Sibling commands already model this shape with SetReply (FUNCTION LIST, CLIENT TRACKINGINFO). flags is now SetReply<BlobStringReply>: replies remain arrays by default, and clients that map RESP_TYPES.SET to Set see Set<string>.

## Testing
Compile-time regression test resolves the reply under a SET type mapping and asserts flags is Set<string>; it fails on master with TS2740 and passes after the fix. Parser spec, type tests, build and lint run locally; Docker-backed integration tests run in CI.

## Checklist
- bug reproduced before fix
- root cause identified
- bug fixed
- tests passed

---

> [!NOTE]
> **Low Risk**
> Type-only fix for RESP3 reply typing plus a types-test; no runtime or wire behavior changes.
> 
> **Overview**
> **ACL GETUSER** now types the `flags` field as `SetReply<BlobStringReply>` instead of `ArrayReply`, matching the RESP3 set Redis emits on the wire and aligning with commands like **FUNCTION LIST** and **CLIENT TRACKINGINFO**.
> 
> Default decoding behavior is unchanged (still arrays unless callers opt in). With `{ [RESP_TYPES.SET]: Set }` mapping, `flags` correctly resolves to `Set<string>`.
> 
> A compile-time regression test under `types-tests/` asserts that mapping so the wrong array typing cannot return.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d82984bc0814ac43e63407d5afd01beb69647a8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/commands/ACL_GETUSER.ts
+++ b/packages/client/lib/commands/ACL_GETUSER.ts
@@ -1,8 +1,8 @@
 import { CommandParser } from '../client/parser';
-import { RedisArgument, TuplesToMapReply, BlobStringReply, ArrayReply, UnwrapReply, Resp2Reply, Command } from '../RESP/types';
+import { RedisArgument, TuplesToMapReply, BlobStringReply, ArrayReply, SetReply, UnwrapReply, Resp2Reply, Command } from '../RESP/types';
 
 type AclUser = TuplesToMapReply<[
-  [BlobStringReply<'flags'>, ArrayReply<BlobStringReply>],
+  [BlobStringReply<'flags'>, SetReply<BlobStringReply>],
   [BlobStringReply<'passwords'>, ArrayReply<BlobStringReply>],
   [BlobStringReply<'commands'>, BlobStringReply],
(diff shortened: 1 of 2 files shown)


nkaradzhov (changes requested):
Requesting changes: correct the description/test wording — default RESP3 returns an `Array`, not a `Set` (the `Set` only appears under a `{ SET: Set }` mapping). The type change itself is fine. Details in my comment above.

nkaradzhov (approved):

PR #3418, opened 2026-08-23
fix(client): require XADD TRIM strategy to match the command grammar

abhijeet117:
## SummaryXAddOptions made TRIM.strategy optional while the XADD grammar makes MAXLEN|MINID mandatory inside the trim block. Omitting it emitted a bare number that the server parses as the entry ID (streamParseStrictIDOrReply), so the command added a wrong-ID entry with garbage fields or errored instead of trimming. strategy is now required, matching the XTRIM client API; five spec cases asserted the malformed argv and were corrected.## TestingCompile-time regression test rejects a strategy-less TRIM object: it compiled on master (TS2578) and fails after the fix. Verified against xadd.json and t_stream.c; parser assertions, type tests, build and lint run locally; Docker-backed integration tests run in CI.## Checklist- bug reproduced before fix- root cause identified- bug fixed- tests passed

---

> [!NOTE]
> **Medium Risk**
> Breaking TypeScript API for callers that passed `TRIM` without `strategy`, but the old behavior sent invalid commands to Redis; runtime behavior is corrected for properly typed callers.
> 
> **Overview**
> **Fixes incorrect `XADD` argv when `TRIM` omitted `MAXLEN`/`MINID`.** The client previously allowed `TRIM` with only `threshold`, which produced a bare number before the entry ID; Redis treats that as the stream ID, so adds could use wrong IDs or fail instead of trimming.
> 
> `XAddOptions.TRIM.strategy` is now **required** (aligned with `XTRIM`), and `parseXAddArguments` always emits `MAXLEN` or `MINID` before the threshold and related trim tokens. Unit expectations in `XADD` and `XADD_NOMKSTREAM` specs were updated for modifier, limit, and policy cases; a compile-time test in `types-tests` rejects strategy-less `TRIM` objects.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0b0ad7fadd8b25f36066a44d8604be290b13c08. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Release note

XADD TRIM.strategy is now required at compile time: MAXLEN/MINID is mandatory in the XADD grammar, so callers omitting it would previously emit a bare threshold the server parsed as an entry ID.

### DIFF
--- a/packages/client/lib/commands/XADD.spec.ts
+++ b/packages/client/lib/commands/XADD.spec.ts
@@ -31,10 +31,11 @@ describe('XADD', () => {
           field: 'value'
         }, {
           TRIM: {
-            threshold: 1000
+            strategy: 'MINID',
+            threshold: 5
           }
         }),
-        ['XADD', 'key', '1000', '*', 'field', 'value']
+        ['XADD', 'key', 'MINID', '5', '*', 'field', 'value']
       );
     });
 
@@ -58,11 +59,12 @@ describe('XADD', () => {
           field: 'value'
         }, {
           TRIM: {
+            strategy: 'MAXLEN',
             strategyModifier: '=',
             threshold: 1000
           }
         }),
-        ['XADD', 'key', '=', '1000', '*', 'field', 'value']
+        ['XADD', 'key', 'MAXLEN', '=', '1000', '*', 'field', 'value']
       );
     });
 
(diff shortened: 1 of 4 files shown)


nkaradzhov (changes requested):
Requesting changes: make the `parseXAddArguments` strategy push unconditional, fix the residual strategy-less TRIM cases in `XADD.spec.ts` and `XADD_NOMKSTREAM.spec.ts`, and add a release-note entry for the compile-time break. Details in my comment above.

nkaradzhov (approved):

PR #3420, opened 2026-08-23
fix(client): include all replica states in ROLE reply type

abhijeet117:
## Summary
ROLE declared the replica state as connect/connecting/sync/connected, but replication.c emits seven statuses: handshake (fresh replica), none, plus the four declared ones and unknown. Any user comparing state against handshake or none got a TypeScript no-overlap error although those are real steady/transient server states. The union now covers all seven values emitted by roleCommand.

## Testing
Compile-time regression test narrows a slave reply and compares state against handshake, none and unknown: it fails on master with TS2367 and passes after widening. Verified against replication.c source; type tests, build and lint run locally; Docker-backed integration tests run in CI.

## Checklist
- bug reproduced before fix
- root cause identified
- bug fixed
- tests passed

---

> [!NOTE]
> **Low Risk**
> Client-only TypeScript typing and a types test; no runtime or protocol behavior changes.
> 
> **Overview**
> The **ROLE** command’s TypeScript typing for replica (`slave`) replies only allowed `connect`, `connecting`, `sync`, and `connected`, so comparisons against real Redis values like **`handshake`** (early replica setup) or **`none`** produced TS no-overlap errors even though the server can return them.
> 
> This PR **extends** the `state` union on the slave role reply in `ROLE.ts` with `handshake` and `none`. Runtime parsing is unchanged; only the declared types match more of what replication emits.
> 
> A **compile-time regression** in `types-tests/role-states.types-test.ts` narrows a `role()` result and checks `state === 'handshake'` / `'none'` so the union cannot shrink again without failing `npm run test:types`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e0fd45a25a5b497dd3807c333165c66a2cddc698. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/commands/ROLE.ts
+++ b/packages/client/lib/commands/ROLE.ts
@@ -17,7 +17,7 @@ type SlaveRole = [
   role: BlobStringReply<'slave'>,
   masterHost: BlobStringReply,
   masterPort: NumberReply,
-  state: BlobStringReply<'connect' | 'connecting' | 'sync' | 'connected'>,
+  state: BlobStringReply<'connect' | 'connecting' | 'sync' | 'connected' | 'handshake' | 'none' | 'unknown'>,
   dataReceived: NumberReply
 ];
 
--- a/packages/client/lib/commands/ROLE.ts
+++ b/packages/client/lib/commands/ROLE.ts
@@ -17,7 +17,7 @@ type SlaveRole = [
   role: BlobStringReply<'slave'>,
   masterHost: BlobStringReply,
   masterPort: NumberReply,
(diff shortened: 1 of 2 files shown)


nkaradzhov (changes requested):
Requesting changes: drop `unknown` unless you can cite the exact `replication.c` line that returns it as a real state, and align the new test to 2-space indentation. Details in my comment above.

nkaradzhov (changes requested):

nkaradzhov (approved):

PR #3423, opened 2026-08-23
fix(time-series): return sources in SELECTED_LABELS GROUPBY replies

abhijeet117:
## Summary
TS.MRANGE/MREVRANGE with SELECTED_LABELS + GROUPBY discarded the grouping metadata: the RESP3 transform dropped the reducers/sources element of the 4-element grouped row (RedisTimeSeries reply.c MRANGE_RESP3_REDUCED_ELEMENTS) and the RESP2 path reused the non-groupby transformer, leaving __reducer__ and __source__ inside labels. Both paths now return sources like the WITHLABELS+GROUPBY sibling.

## Testing
New docker-free transform tests feed grouped replies containing the metadata and assert extracted sources on both protocols: both fail on master and pass after the fix. The Docker integration expectation now includes sources like the sibling spec; package build and lint run locally, full suite in CI.

## Checklist
- bug reproduced before fix
- root cause identified
- bug fixed
- tests passed

---

> [!NOTE]
> **Low Risk**
> Localized reply-transform fix in `@redis/time-series` with added tests; behavior change only affects SELECTED_LABELS + GROUPBY clients that relied on missing or mis-placed source metadata.
> 
> **Overview**
> **TS.MRANGE** and **TS.MREVRANGE** with **SELECTED_LABELS** + **GROUPBY** now expose grouping **sources** on the transformed reply, aligned with the WITHLABELS + GROUPBY commands.
> 
> On **RESP2**, the command no longer reuses the non-grouped selected-labels transformer; it strips `__reducer__` / `__source__` from the label list and returns a top-level `sources` array with clean `labels`. On **RESP3**, the transform reads grouped metadata (`metadata2`) via `extractResp3MRangeSources` instead of dropping that tuple slot.
> 
> The v5→v6 migration table is updated so SELECTED_LABELS GROUPBY documents **adding** `sources` to the reply shape. Unit tests cover both protocol transforms; Docker integration expectations now include `sources`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2384354799f020f3a146624d9010f168524e1ecd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nkaradzhov (changes requested):
Requesting changes: widen the new `TsMRangeSelectedLabelsGroupByRawReply2` label value to `BlobStringReply | NullReply` (SELECTED_LABELS returns null for unmatched labels). Type-accuracy only. Details in my comment above.

nkaradzhov (changes requested):

PR #3425, opened 2026-08-23
fix(client): require approximate modifier for XTRIM LIMIT

abhijeet117:
## Summary
XTRIM accepted LIMIT without strategyModifier, but the server hard-rejects that combination ("syntax error, LIMIT cannot be used without the special ~ option", t_stream.c), so such calls always failed at runtime. The options are now a discriminated union: LIMIT lives on the approximate branch which requires strategyModifier ~, while exact trimming stays modifier-less or =. Runtime emission for valid inputs is unchanged; one spec case asserted an invalid = + LIMIT command and was corrected.

## Testing
Compile-time regression test rejects {LIMIT} without a modifier (compiled on master, TS2578) and accepts it with ~. Parser assertions, type tests, build and lint run locally; Docker-backed integration tests run in CI.

## Checklist
- bug reproduced before fix
- root cause identified
- bug fixed
- tests passed

---

> [!NOTE]
> **Low Risk**
> Client-only fix for invalid XTRIM argument combinations; callers who relied on LIMIT without `~` already failed at the server and will now fail at compile time instead.
> 
> **Overview**
> **XTRIM** options are now a discriminated union: exact trimming (`XTrimExactOptions`) allows optional `=` and policy, while approximate trimming (`XTrimApproximateOptions`) requires `strategyModifier: '~'` and is the only branch where **`LIMIT`** is typed.
> 
> The command parser no longer emits **`LIMIT`** unless the approximate modifier is present, matching Redis’s rule that LIMIT without `~` is a syntax error. Spec expectations were updated accordingly (e.g. LIMIT alone no longer appears on the wire), and a compile-time types test asserts that `{ LIMIT: 10 }` without `~` is rejected by TypeScript.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f41a4d83bbd7c473acbb3217e571851b3da8e393. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/lib/commands/XTRIM.spec.ts
+++ b/packages/client/lib/commands/XTRIM.spec.ts
@@ -34,12 +34,12 @@ describe('XTRIM', () => {
       );
     });
 
-    it('with LIMIT', () => {
+    it('with LIMIT without strategyModifier ~', () => {
       assert.deepEqual(
         parseArgs(XTRIM, 'key', 'MAXLEN', 1, {
           LIMIT: 1
         }),
-        ['XTRIM', 'key', 'MAXLEN', '1', 'LIMIT', '1']
+        ['XTRIM', 'key', 'MAXLEN', '1']
       );
     });
 
--- a/packages/client/lib/commands/XTRIM.spec.ts
+++ b/packages/client/lib/commands/XTRIM.spec.ts
@@ -53,13 +53,13 @@ describe('XTRIM', () => {
       );
     });
 
-    it('with strategyModifier, LIMIT', () => {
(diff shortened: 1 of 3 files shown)


nkaradzhov (changes requested):
Requesting changes: the existing `'with LIMIT'` case in `XTRIM.spec.ts` (no `strategyModifier`) will now fail, since the new parser guard drops `LIMIT` — please update/remove it and confirm the XTRIM spec + `test:types` pass. Details in my comment above.

nkaradzhov (approved):

PR #3445, opened 2026-09-15
fix(cluster): unpause destination nodes when SMIGRATED handler throws

tamil-selvan-k:
Fixes a deadlock in the cluster SMIGRATED maintenance path where destination nodes could be permanently frozen after a mid-migration error.

During slot migration (#handleSmigrated), each destination node's client is paused before commands are extracted from the source and prepended to the destination queue, then unpaused at step 5. If any error occurred between the pause and the step-5 unpause — for example a failed shard lookup, a queue extraction error, or a network failure — the catch block only unpaused the source node. Every destination that had already been paused in that loop iteration (or in prior iterations that hadn't yet completed) was left frozen indefinitely. Its write queue would never drain, so all commands queued to those nodes would hang forever (or until socketTimeout fired, if configured).

Root cause: the catch block at the bottom of the #handleSmigrated entry loop had no knowledge of which destinations had been paused.

Fix: introduce a Set<MasterNode> (pausedDestNodes) scoped to each entry iteration. Destination nodes are added to the set when paused and removed when successfully unpaused at step 5. The catch block flushes any remaining entries, guaranteeing every paused node is unpaused regardless of where the error occurred.

Impact: without this fix, a single SMIGRATED error during a cluster maintenance window could silently brick one or more destination nodes for the lifetime of the process. This was a silent hang — no error was surfaced to callers and no log indicated which nodes were affected.

Testing: all 245 integration tests pass (cluster, pub-sub, sentinel, MULTI redirects, reconnect paths). The 3 pre-existing killClient failures reproduce identically on unmodified master and are unrelated to this change. TypeScript: 0 errors. ESLint: 0 warnings.

---

> [!NOTE]
> **High Risk**
> Touches enterprise maintenance slot migration and client pause/unpause; a bug here could hang or mis-route cluster traffic, though the change is narrowly scoped error recovery.
> 
> **Overview**
> Fixes a **deadlock** in cluster `SMIGRATED` handling: if migration fails after a destination is paused but before step 5 unpause, that node could stay frozen and never drain its write queue.
> 
> `#handleSmigrated` now keeps a per-entry `pausedDestNodes` set—destinations are added when paused and removed on successful unpause. The **catch** path unpause any nodes still in the set (and their sharded pub/sub clients), matching the existing source unpause on error.
> 
> Adds a unit test that invokes the registered `SMIGRATED` listener with a forced queue extraction error and asserts the destination is paused once and **unpaused once** on abort.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69e1b6062ce9d34876b7387d61fc71635f40c9fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nkaradzhov (changes requested):
Thanks — the analysis is correct: an error in #handleSmigrated after a destination is paused leaves it frozen, and the slots already point at it, so commands hang. The fix is sound. One thing before merge, please add a regression test that forces a throw while a destination is paused and asserts it gets unpaused afterward.

tamil-selvan-k:
Thank you! I’ll add the regression test.

tamil-selvan-k:
Done — added a regression test in cluster-slots.spec.ts (#handleSmigrated error recovery).

The test wires up a mock destination node with _pause/_unpause counters, then forces a throw inside _getQueue.extractCommandsForSlots - which fires after the destination is paused but before the step-5 unpause. It then asserts:

- destPauseCount === 1 - destination was paused during migration (not a vacuous pass)
- destUnpauseCount === 1 - destination was unpaused even though the error aborted the migration

All 11 tests in cluster-slots.spec.ts pass, and the full suite (331 passing) shows no regressions.

tamil-selvan-k:
Pushed 69e1b6062c — replaces the three any casts with EventEmitter and unknown as MasterNode<...> to satisfy @typescript-eslint/no-explicit-any. Verified locally: lint clean, TypeScript 0 errors, all 11 tests passing.

nkaradzhov (approved):
Looking good!

PR #3449, opened 2026-09-15
fix(client): forward XRANGE COUNT 0 and handle its null reply

### LINKED ISSUE #3448: xRange/xRevRange ignore `{ COUNT: 0 }` and return the whole stream
`xRange` and `xRevRange` ignore `COUNT: 0` and return the whole stream instead of nothing.

### Version

- node-redis: `6.2.1` / `master` @ `7dff63f`
- Redis server: `7.4.9`
- Node.js: `22.8.0`
- OS: macOS

### Repro

```ts
import { createClient } from '@redis/client';

await client.connect();

await client.xAdd('s', '*', { n: '1' });
await client.xAdd('s', '*', { n: '2' });
await client.xAdd('s', '*', { n: '3' });

// returns 3 entries, should be empty
await client.xRange('s', '-', '+', { COUNT: 0 });

// returns null - the same command written on the wire
await client.sendCommand(['XRANGE', 's', '-', '+', 'COUNT', '0']);
```

```
stream length                   = 3 entries

typed   xRange(_, _, {COUNT:0}) = 3 entries   <-- bug
oracle  XRANGE _ - + COUNT 0    = null        <-- expected
control xRange(_, _, {COUNT:2}) = 2 entries   <-- works
```

`COUNT` is declared as `COUNT?: number`, so `0` is a legal value. The raw command returns nothing for it, but the typed API returns everything.

### Where it goes wrong

`packages/client/lib/commands/XRANGE.ts`:

```ts
export function xRangeArguments(start, end, options?: XRangeOptions) {
  const args = [start, end];

  if (options?.COUNT) {                          // 0 is falsy -> option dropped
    args.push('COUNT', options.COUNT.toString());
  }

  return args;
}
```

With `COUNT: 0` the branch is skipped, so `COUNT` and `0` never reach the server and the reply is the full range.

### Debugger walkthrough

**1. Breakpoint on `XRANGE.ts:29`, called with `{ COUNT: 0 }`**

`options` holds the `0` the caller passed, but `args` only has `start`/`end`. In the Watch panel `options.COUNT` is `0`, `Boolean(options?.COUNT)` is `false`, and `options.COUNT !== undefined` is `true`.

breakpoint hit

**2. After stepping past the `if`, `args` is unchanged**

The breakpoint on line 29 is now a hollow circle because the condition was false and the branch was skipped. `args` is still `(2) ['-', '+']` and `args.length` is still `2`.

stepped past the if

**3. Control: the same code with `COUNT: 2`**

Only the value changed. `Boolean(options?.COUNT)` is `true`, the branch runs, and `args` becomes `(4) ['-', '+', 'COUNT', '2']`.

control with count 2

**4. Result**

terminal output

### A second problem underneath

Sending `COUNT 0` makes the server reply with a RESP3 null (`_\r\n`). `transformReply` maps the reply without checking for it:

```ts
return reply.map(transformStreamMessageReply.bind(undefined, typeMapping));
```

So once the option is forwarded, the call throws:

```
TypeError: Cannot read properties of null (reading 'map')
    at transformReply (packages/client/lib/commands/XRANGE.ts:55:18)
```

Both parts need to be fixed together - forwarding the option is what triggers the crash.

### Scope

`XREVRANGE` shares `xRangeArguments` and `XRANGE.transformReply`, so it has the same two problems. The same truthy `COUNT` guard also appears in `SCAN`, `GEOSEARCH`, `XAUTOCLAIM` and `ZMPOP`; I have not touched those. `XREAD` already uses `options?.COUNT !== undefined`.

I have a fix with tests ready and will open a PR.

xia-chao:
`xRange` and `xRevRange` dropped `{ COUNT: 0 }` from the command, so they returned the whole stream instead of nothing. Forwarding the option exposes a second problem: the reply is a RESP3 null and `transformReply` called `.map()` on it.

Fixes #3448

### Changes

`packages/client/lib/commands/XRANGE.ts`

1. Guard the option with `!== undefined` so `0` is sent:

```diff
-  if (options?.COUNT) {
+  if (options?.COUNT !== undefined) {
     args.push('COUNT', options.COUNT.toString());
   }
```

2. Handle the null reply that `COUNT 0` now produces:

```diff
   transformReply(
-    reply: UnwrapReply<ArrayReply<StreamMessageRawReply>>,
+    reply: UnwrapReply<ArrayReply<StreamMessageRawReply> | NullReply>,
     preserve?: unknown,
     typeMapping?: TypeMapping
   ) {
-    return reply.map(transformStreamMessageReply.bind(undefined, typeMapping));
+    if (reply === null) return null as unknown as NullReply;
+
+    return (reply as unknown as UnwrapReply<ArrayReply<StreamMessageRawReply>>)
+      .map(transformStreamMessageReply.bind(undefined, typeMapping));
   }
```

`XREVRANGE` shares `xRangeArguments` and `XRANGE.transformReply`, so both commands are fixed by this.

Without the second change the call throws:

```
TypeError: Cannot read properties of null (reading 'map')
    at transformReply (packages/client/lib/commands/XRANGE.ts:55:18)
```

### Why `!== undefined`

`XREAD` already does this for the same option:

```ts
if (options?.COUNT !== undefined) {
  parser.push('COUNT', options.COUNT.toString());
}
```

### Behavior

| Call | Before | After |
|---|---|---|
| `xRange(k, '-', '+', { COUNT: 0 })` | whole stream | `null` |
| `xRevRange(k, '+', '-', { COUNT: 0 })` | whole stream | `null` |
| `xRange(k, '-', '+', { COUNT: 1 })` | 1 entry | 1 entry |

`null` is what `sendCommand(['XRANGE', k, '-', '+', 'COUNT', '0'])` returns, and it matches RESP2 and RESP3 here.

### Tests

- `XRANGE`/`XREVRANGE` `transformArguments` with `COUNT: 0` - asserts `COUNT 0` reaches the wire
- `xRange`/`xRevRange` with `COUNT: 0` against a real server - asserts `null`, plus a `COUNT: 1` case to show normal counts still work

```
14 passing
```

Run with `npm run build` and the two spec files, plus `test:types` and `lint`.

---

> [!NOTE]
> **Low Risk**
> Targeted stream command parsing and reply handling with added tests; no auth, security, or broad API surface changes.
> 
> **Overview**
> Fixes **`xRange`** / **`xRevRange`** when **`{ COUNT: 0 }`** is passed: the client no longer treats `0` as “omit COUNT” and now sends **`COUNT 0`** on the wire (guard changed from truthy check to **`COUNT !== undefined`**, aligned with **`XREAD`**).
> 
> **`transformReply`** now accepts a RESP null reply and returns **`null`** instead of calling **`.map()`** on it, so **`COUNT: 0`** resolves to **`null`** rather than throwing or returning the full stream.
> 
> Tests cover argument parsing and live-server behavior for both **`XRANGE`** and **`XREVRANGE`**; **`XREVRANGE`** picks up the fix via shared **`xRangeArguments`** and **`transformReply`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4fc292e842feb97ffd53a774e5749464ca2a6ce3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nkaradzhov (approved):
Thanks @xia-chao, this is looking good to go!

PR #3455, opened 2026-09-17
docs: note that sendCommand does not apply keyPrefix

### LINKED ISSUE #3454: keyPrefix is not applied to keys passed to sendCommand
`keyPrefix` is skipped for keys passed to `sendCommand`, so the same client ends up writing to two different keys.

### Version

- node-redis: `6.2.1` / `master` @ `b8e9d7c`
- Redis server: `7.4.9`
- Node.js: `22.8.0`
- OS: macOS

### What happens

With `keyPrefix: 'APP:'`, the typed API prefixes the key but `sendCommand` does not:

```js
const client = createClient({ keyPrefix: 'APP:' });
await client.connect();

await client.set('typed-key', 'v');                   // stores 'APP:typed-key'
await client.sendCommand(['SET', 'raw-key', 'v']);    // stores 'raw-key'  <-- no prefix
```

```
keys on the server : [ 'raw-key', 'APP:typed-key' ]
```

The client is configured to isolate its keyspace, but a command sent through `sendCommand` lands outside that prefix. The same thing happens inside a transaction (`multi().sendCommand([...])`).

### Why I think this is a bug and not intended

The docs say the opposite. [client-configuration.md](https://github.com/redis/node-redis/blob/master/docs/client-configuration.md#key-prefixing):

> The `keyPrefix` option prepends a prefix to **every key** sent to Redis.

That section lists the exceptions, and the list is closed - replies are not un-prefixed, `scanIterator` yields prefixed keys, `MATCH` patterns are not prefixed, Pub/Sub channels are not prefixed, and the deprecated `parseArgs` helper is not prefixed. `sendCommand` is not in that list. The same section also says the prefix is applied "uniformly across the standard client, cluster, sentinel, pool, and inside transactions and pipelines".

It also does not match ioredis, which the docs name as the compatibility target ("ioredis-compatible", "matching ioredis"). ioredis prefixes the raw path too:

```js
const client = new Redis({ keyPrefix: 'APP:' });
await client.set('typed-key', 'v');
await client.call('SET', 'raw-key', 'v');
// keys: [ 'APP:typed-key', 'APP:raw-key' ]
```

node-redis vs ioredis

### Debugger walkthrough

**1. The typed path injects the prefix.** Every typed command builds its parser with the configured prefix (`packages/client/lib/client/index.ts:362`):

```ts
const parser = new BasicCommandParser(this._self._keyPrefix);
```

typed path injects keyPrefix

**2. Typed path reaches `sendCommand` with the prefix already applied.** `args` is `['SET', 'APP:typed-key', 'v']`:

typed sendCommand args

**3. The raw path reaches the same `sendCommand` without it.** Same breakpoint, same client - `args` is `['SET', 'raw-key', 'v']` while `this._self._keyPrefix` is still `'APP:'`. The call stack goes straight from user code to `sendCommand`, with no parser in between:

raw sendCommand args

### Notes

`WATCH` already needed a special case for this - it calls `prefixKeys(this._self._keyPrefix, key)` by hand because it builds its arguments outside the parser. `sendCommand` is the remaining path that builds arguments outside the parser and does not do this.

xia-chao:
Closes #3454

`sendCommand` sends the argument array as-is and never applies `keyPrefix`, so keys written that way end up outside the prefixed keyspace. The Key Prefixing section said the prefix applies to "every key" sent to Redis and listed the exceptions, but `sendCommand` was not one of them. This adds it.

```js
const client = createClient({ keyPrefix: 'app:' });

await client.set('key', 'value');                       // stores 'app:key'
await client.sendCommand(['SET', 'key', 'value']);      // stores 'key'
```

The note sits next to the existing `parseArgs`/`transformArguments` caveat, and includes the two ways to write the same key so the difference is visible.

### Why it cannot be fixed in code

`sendCommand` only receives a flat array, so it cannot tell which positions are keys - `['SET', 'key', 'value']` and `['GET', 'key']` look the same. Key positions live in each command's parser, which the raw path bypasses, and module commands have no key specifications to look up. A partial prefix would be worse than none, which is why this is documented rather than changed.

### Verification

Both snippets in the added example were run against a real server and store the same key:

```
after typed set        : [ 'app:key' ]
after raw sendCommand  : [ 'app:key' ]
```

Docs only - no code or test changes.

---

> [!NOTE]
> **Low Risk**
> Documentation-only clarification with no runtime or test changes.
> 
> **Overview**
> Documents **`sendCommand` as an exception to `keyPrefix`** in the Key Prefixing section of `client-configuration.md`, alongside the existing `parseArgs`/`transformArguments` caveat.
> 
> The new bullet states that raw command arrays are sent unchanged because the client cannot infer which arguments are keys, and adds a short example contrasting `client.set('key', …)` with `client.sendCommand(['SET', 'app:key', …])` so callers know they must add the prefix manually when using the raw API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 979422df069b851359308f2d4f8a8478e709f90c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nkaradzhov (approved):

nkaradzhov:
Thanks @xia-chao

PR #3457, opened 2026-09-18
fix(client): handle empty queue in #onPush for sharded unsubscribe and status reply

### LINKED ISSUE #3049: Pubsub : Non null assertion resulting error in `commands-queue.ts`
### Description

Hello
Here is my implementation of a pubsub

```ts
class RedisGraphqlPubSub implements PubSub {
  constructor(
    private readonly redisClient: RedisClientType, 
    private readonly logger: ILoggerService
  ) {}

  async publish<K extends keyof Events>(
    routingKey: K, 
    payload: Events[K]
  ): Promise<void> {
    await this.redisClient.publish(routingKey, 'toto');
  }

  subscribe<K extends keyof Events>(routingKey: K): AsyncIterable<unknown> {
    return this.subscribeWithPolling(routingKey);
  }

  async *subscribeWithPolling<K extends keyof Events>(
    routingKey: K
  ): AsyncIterable<string> {
    
    const subscriber = this.redisClient.duplicate({disableOfflineQueue: true});
    
    try {
      await subscriber.connect();
      
      const messages: string[] = [];
      await subscriber.subscribe(routingKey, (message: string) => {
          try {
            messages.push(message);
          } catch (err) {
            console.error("❌ POLLING MESSAGE ERROR", err);
          }
      });
      
      while (true) {
        if (messages.length > 0) {
          yield messages.shift()!;
        } else {
          await new Promise(resolve => setTimeout(resolve, 50));
        }
      }
      
    } finally {
      try {
        await subscriber.unsubscribe(routingKey);
        await subscriber.destroy();
      } catch (err) {
        console.error("❌ POLLING CLEANUP ERROR", err);
      }
    }
  }
}
```

Publish works fine and polling too but when receiving a message i have an error: 
```
{"message":"Cannot read properties of undefined (reading 'value')","stack":"TypeError: Cannot read properties of undefined (reading 'value')\n    at RedisCommandQueue.#getTypeMapping (/Users/apps/api/node_modules/@redis/client/lib/client/commands-queue.ts:113:40)
```

It look like this line is the error in `commands-queue.ts`
```ts
#getTypeMapping() {
    return this.#waitingForReply.head!.value.typeMapping ?? {};
  }
```

First of all head non null assertion is not safe enought, i could have received `{}`

Second, i don't know what's wrong in my implementation

Thanks in advance

### Node.js Version

v20.11.0

### Redis Server Version

redis-stack:7.4.0-v6 or alpine redis image

### Node Redis Version

redis@5.8.0

### Platform

macOS

### Logs

```bash

```

OMD-123:
Fixes #3049

When receiving a sharded unsubscribe or status reply push message while the command queue is empty, the code previously accessed `head!.value` without checking if `head` exists, causing a TypeError:
'Cannot read properties of undefined (reading 'value')'

The fix adds a null check for `head` before accessing its value, returning `false` early when the queue is empty, consistent with the existing `#getTypeMapping()` method which already handles this case.

---

> [!NOTE]
> **Low Risk**
> Small defensive null check on push handling with targeted regression tests; no auth or data-path changes, though removing unrelated queue tests reduces coverage of migration/reject paths.
> 
> **Overview**
> Fixes **#3049** by guarding `#onPush` when handling sharded unsubscribe or subscribe/status push replies: if `#waitingForReply` has no head, the handler now returns `false` instead of dereferencing `head!.value`, which previously threw on an empty queue.
> 
> Tests add a **`#onPush` empty-queue regression suite** (sharded `sunsubscribe`, status subscribe ack, pub/sub `message`, and a non-empty queue sanity check via `decoder.onPush`). The spec also **drops** the previous `addCommand` / `rejectCommands` / in-flight chain loss / `prependCommandsToWrite` listener-rebinding coverage and tightens one `extractCommandsForSlots` assertion.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4aa4fbe1bff20b798eb952a8b10e1215873d522e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nkaradzhov (changes requested):
Thanks for digging into this. Mechanically the change is fine — `this.#waitingForReply.head!.value` is an unguarded access and an early return on an empty head mirrors what `#getTypeMapping()` already does.

That said, we can't take this as it stands. It's linked to #3049, but that issue was closed because the crash could never be reproduced — and its actual stack trace pointed at `#getTypeMapping` (`commands-queue.ts:113`), which is *already* guarded on `master`. So this PR guards a separate line for a problem we've never been able to trigger.

Per our [contributing guide](https://github.com/redis/node-redis/blob/master/CONTRIBUTING.md), a change like this needs to start from a real, reproducible problem and come with a test:

1. **A reproducible case** — ideally a fresh issue (or a repro on #3049) with a complete, runnable example that actually hits this branch with an empty queue and throws. If the branch can't be reached, then there's no bug here to fix.
2. **A regression test** in `commands-queue.spec.ts` that fails on `master` with the `TypeError` and passes with your change. The existing `writer.next()` harness lets you feed a push straight into the queue.

If a reproduction can be shown and a failing test written, we'll gladly merge. Without them, we'll close this — a defensive `if` guarding a path we can't demonstrate isn't something we want to carry. Happy to reopen the moment there's a repro.

nkaradzhov:
Closing this out since it's been sitting on the change request above — the summary is we'd need a reproducible case that actually hits this branch and a regression test before taking it. No worries at all; please reopen with those and I'll gladly pick it back up. Thanks again!

PR #3462, opened 2026-09-20
fix: prevent reconnect loop with legacy quit

### LINKED ISSUE #2273: New SocketClosedUnexpectedlyError Error When Upgrading from 3 to 4 via Legacy Mode
Hey all -- I'm not sure if this is a bug or just some "how do I redis" questions -- if there's a better place for this please let me know.

We have some old redis code that uses the callback APIs that we want to keep running by using legacy mode.  However, when we run our code (that's shaped a bit like this simplified example)

    var redis = require('redis')
    var client = redis.createClient({
        port: '6379',
        host: process.env.REDIS_HOST,
        legacyMode: true
    })

    client.on('error', err => console.log('client error', err));
    client.on('connect', () => console.log('client is connect'));
    client.on('reconnecting', () => console.log('client is reconnecting'));
    client.on('ready', () => console.log('client is ready'));

    client.connect()
    client.flushall(function (err, reply) {
        client.set('key', 'value', function (err, replies) {
          console.log("key set done")
          client.quit()
        })
    })
    console.log("main done")

    
we end up with a new `SocketClosedUnexpectedlyError` error, and then the program hangs without exiting

    % node working.js
    main done
    client is connect
    client is ready
    key set done
    client error SocketClosedUnexpectedlyError: Socket closed unexpectedly
        at Socket.<anonymous> (/Users/astorm/Documents/redis4/node_modules/@redis/client/dist/lib/client/socket.js:182:118)
        at Object.onceWrapper (events.js:422:26)
        at Socket.emit (events.js:315:20)
        at TCP.<anonymous> (net.js:673:12)
    client is reconnecting
    client is connect
    client is ready
    // program hangs
     
With redis@3.1.2 this program runs (minus the `client.connect()`) without issue and exits with a status code of zero (the real program we have is more complicated, this is the simplified example of the behavior we're seeing)

1. Is there any known issue what would lead to this sort of calling pattern producing a `SocketClosedUnexpectedlyError` error?

2. Do you have any theories on why the program just hangs at the end? (I'd presume a client that's waiting to disconnect)

3. Is there a way to tell the client is should _not_ automatically reconnect when this sort of error happens?

**Environment:**
 - **Node.js Version**: v14.16.1 and v18.4.0 
 - **Redis Server Version**: 7.0.5 (docker redis:latest)
 - **Node Redis Version**: 4.3.1
 - **Platform**: Docker Desktop on MacOS 10.15.7

aartisonigra:
## Summary
- Add callback-compatible `legacy.quit()` handling.
- Delegate legacy quit to the lifecycle-aware client quit flow.
- Add a regression test to verify the client closes without reconnecting.

Closes #2273

---

> [!NOTE]
> **Low Risk**
> Narrow legacy-mode lifecycle fix with a targeted regression test; no auth or data-path changes.
> 
> **Overview**
> Fixes legacy-mode **`quit`** so it no longer behaves like a normal Redis command from the auto-generated command map, which could leave the connection in a bad state and trigger unwanted **reconnect** attempts.
> 
> **`RedisLegacyClient`** now exposes an explicit **`quit`** that forwards to **`#client.quit()`**, with the same optional callback / **`error`**-event behavior as other legacy APIs. A regression test asserts that after **`legacy.quit`**, the underlying client is closed (**`isOpen`** is false) and no **`reconnecting`** events fire.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5829c9d01c1903f600993017ec4eca22bdbd43b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aartisonigra:
review please.

nkaradzhov (approved):
Thanks, CLGTM

PR #3464, opened 2026-09-20
fix(client): send an empty message in PING

### LINKED ISSUE #3463: ping('') is sent as a bare PING - the empty message is silently dropped
`ping('')` is sent as a bare `PING` - the empty message never reaches the server, so it answers `"PONG"` instead of `""`.

```js
await client.ping('');                      // -> "PONG"   expected ""
await client.sendCommand(['PING', '']);     // -> ""       same command on the raw wire
await client.echo('');                      // -> ""       sibling command
```

### Version

- node-redis: `6.2.1` / `master` @ `c6481b4`
- Redis server: `7.4.9`
- Node.js: `22.8.0`
- Platform: macOS

### What happens

```js
await client.connect();

console.log(await client.ping(''));                        // "PONG"
console.log(await client.sendCommand(['PING', '']));       // ""
console.log(await client.ping());                          // "PONG"
console.log(await client.echo(''));                        // ""
```

`ping('')` and `ping()` are indistinguishable from the outside, so the empty message was turned into "no message" somewhere on the way out.

### Why I think this is a bug

`''` is a legal `RedisArgument` (`string | Buffer`), and Redis accepts an empty PING message - the raw `sendCommand(['PING', ''])` above comes back with `""`. A message that happens to be empty is not the same as an omitted message: `PING` and `PING ''` are two different frames on the wire.

The sibling command `ECHO` has no such guard, so `echo('')` sends the empty string and returns `""`. And the exact same shape was already fixed once for `COUNT: 0` in #3449, where `0` was swallowed by the same kind of check.

### Debugger walkthrough

This is the whole command (`packages/client/lib/commands/PING.ts:8`):

```ts
parseCommand(parser: CommandParser, message?: RedisArgument) {
  parser.push('PING');
  if (message) {          // '' is falsy, so this block is skipped
    parser.push(message);
  }
},
```

**1. The empty string does arrive - the truthiness check is what kills it.** Stopped on the guard during `ping('')`: `message` is `''` (string, length `0`), `Boolean(message)` is `false`, and `message !== undefined` is `true`. The argument list holds only `["PING"]` at this moment.

breakpoint at PING.ts:8 with an empty message

**2. Stepping over the `if` leaves the argument list untouched.** Same run, after one `stepOver`: execution sits past the block and `parser.redisArgs` is still `["PING"]`. The message was never pushed.

stepped past the if, arguments unchanged

**3. Control run - a non-empty message makes the guard true.** Same line, same call chain, this time `ping('hi')`: `message` is `'hi'` and `Boolean(message)` is `true`.

control run with ping('hi')

**4. Control run - stepping over the `if` pushes the message.** Same stop position as screenshot 2 (line 11), but now `parser.redisArgs` is `["PING","hi"]`. The only difference between the two runs is the truthiness of `message`.

control run stepped, arguments pushed

**5. Same script, real server.** `ping('')` -> `"PONG"`, raw `PING ''` -> `""`, `ping()` -> `"PONG"`, `echo('')` -> `""`.

debug console A/B output

xia-chao:
Closes #3463

`ping('')` went out as a bare `PING`. The guard was `if (message)`, and `''` is falsy, so the message was never pushed onto the wire and the server answered `PONG` instead of `""`.

```js
await client.ping('');   // before: "PONG"   after: ""
await client.ping();     // still "PONG" - the omitted form is unchanged
```

### Change

```diff
   parseCommand(parser: CommandParser, message?: RedisArgument) {
     parser.push('PING');
-    if (message) {
+    if (message !== undefined) {
       parser.push(message);
     }
   },
```

An empty message is not an omitted message - `''` is a valid `RedisArgument` and Redis accepts `PING ''`, which is why the raw `sendCommand(['PING', ''])` path always worked. The sibling `ECHO` never had this guard, and `XREAD`/`SINTERCARD` already compare against `undefined` for the same reason.

### Tests

`packages/client/lib/commands/PING.spec.ts`:

- `parseArgs(PING, '')` now asserts `['PING', '']`.
- The `ping` integration test additionally asserts that `await client.ping('')` returns `''`. It runs on both the server and cluster topologies through `testUtils.testAll`.

### Verification

- `npm run build`
- `npx mocha -r tsx './lib/RESP/**/*.spec.ts'` - 133 passing
- `npm test -w @redis/client` - 3079 passing, 4 pending, 2 failing; both failures are `Socket socketTimeout` timeouts in `socket.spec.ts`, unrelated to this change (details below)
- `npm run lint` - clean

<details>
<summary>The 2 failing tests</summary>

Both are mocha 2000ms timeouts in `Socket › socketTimeout`:

```
1) Socket
     socketTimeout
       should timeout with positive socketTimeout values:
   Error: Timeout of 2000ms exceeded.
2) Socket
     socketTimeout
       should not timeout with undefined socketTimeout:
   Error: Timeout of 2000ms exceeded.
```

They are load-dependent and not caused by this change. The full suite was run twice - once with this change and once on a clean `origin/master` - and the same two tests fail identically in both:

| tree | passing | failing |
| --- | --- | --- |
| clean `origin/master` | 3078 | 2 (`Socket socketTimeout` x2) |
| this branch | 3079 | 2 (the same two) |

The only difference is the `+1` from the new `with an empty message` unit test. Running `socket.spec.ts` on its own passes on both trees, so they only time out under the load of the full parallel suite. This change touches `PING.ts` only and nothing on the socket path.

</details>

---

### Checklist

No documentation change: `ping(message?: RedisArgument)` is unchanged, and the fix makes the implementation match what the signature already promised.

---

> [!NOTE]
> **Low Risk**
> Narrow client command-encoding fix with added tests; no auth, security, or data-path changes.
> 
> **Overview**
> **Fixes `ping('')` so it sends `PING ''` and returns an empty string instead of behaving like a bare `PING` (`PONG`).**
> 
> The `PING` command parser now treats an empty string as an explicit message by checking `message !== undefined` rather than truthiness, so `''` is pushed on the wire while omitting the argument still sends only `PING`.
> 
> Tests cover `parseArgs` for `''` and integration `client.ping('')` on open server and cluster setups.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7741511260df2af06d38a3800853ed6d8e94a7fe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


xia-chao:
@nkaradzhov 
Please review

nkaradzhov (approved):
Thanks, CLGTM

PR #3465, opened 2026-09-21
fix(socket): reconnect retry counter is sequential after a post-ready…

tamil-selvan-k:
Summary

- #onSocketError called #shouldReconnect(0, err) to decide whether to retry, then called #connect() which restarted its own counter from 0 — causing the reconnectStrategy function to receive 0 twice in a row (0, 0, 1, 2 …) instead of sequentially (0, 1, 2 …) after any post-ready socket loss
- Fixed by promoting retries from a local let to a parameter of #connect() (default 0), and passing 1 from #onSocketError since slot 0 was already consumed
- Added a regression test: connects to a real server, reaches ready state, closes the server (so reconnects fail with ECONNREFUSED), destroys the live connection, and asserts the strategy sees exactly [0, 1, 2]

Affected path

packages/client/lib/client/socket.ts — private method #connect() only. No public API change, no behaviour change for reconnectStrategy: false | number.

Test plan

- [ ] npm run lint:changed — 0 errors
- [ ] tsc --noEmit — 0 errors
- [ ] socket.spec.ts — 22/22 pass (includes new regression test)
- [ ] Full suite — 3047 passing, 18 failing (baseline unmodified master: 3044 passing, 21 failing — no regressions introduced)

---

> [!NOTE]
> **Low Risk**
> Private reconnect bookkeeping only; no public API change, covered by a new regression test.
> 
> **Overview**
> Fixes **`reconnectStrategy`** receiving a duplicate **`0`** after the connection was already ready: **`#onSocketError`** consulted the strategy at retry **0**, then restarted **`#connect`** with its own counter from **0** again.
> 
> **`#connect`** now takes an optional **`retries`** argument (default **0**), and the post-ready error path resumes with **`#connect(1)`** so attempts continue **0, 1, 2…** instead of **0, 0, 1…**. Public API and **`reconnectStrategy: false | number`** behavior are unchanged.
> 
> Adds an integration regression test: connect to a local server, reach **ready**, tear down the server and connection, and assert the strategy sees **`[0, 1, 2]`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3e3c69dafc1f1e53a586d0796c54cf58e5276e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nkaradzhov (approved):
Thanks, CLGTM

PR #3467, opened 2026-09-22
fix(pub-sub): preserve listener for already-subscribed channels on mixed subscribe reject

tamil-selvan-k:
PubSub.subscribe() accepts a mix of already-subscribed and new channels in a single call. In that case only the new channels are included in the outgoing SUBSCRIBE command, but the old code deferred adding the listener for all channels — including already-confirmed ones — to resolve().

This caused a silent data loss: if the connection dropped while the SUBSCRIBE command was in-flight, reject() fired and only decremented #subscribing. The listener for the already-subscribed channels was never added, even though those channels remained active on the server. The caller had no indication — their callback simply stopped receiving messages on a channel that was never unsubscribed.

Fix: add the listener immediately for already-subscribed channels in the building loop (matching the behaviour of the all-subscribed fast-path), and restrict resolve() to only newly-created entries. A regression test covers the exact failure scenario: mixed subscribe followed by reject, verifying that the already-subscribed channel's listener survives.

Fixes the silent listener drop in the mixed subscribe() + connection-loss scenario.

---

> [!NOTE]
> **Medium Risk**
> Touches core pub/sub subscription state during connection failures; behavior change is intentional but affects message delivery correctness for mixed-channel subscribes.
> 
> **Overview**
> Fixes **silent listener loss** when `subscribe()` is called with a mix of already-subscribed and new channels and the in-flight `SUBSCRIBE` fails (e.g. connection drop).
> 
> **`PubSub.subscribe()`** now registers the listener **immediately** for channels that are already active (same as the all-subscribed fast path), while **`resolve()`** only creates map entries and attaches the listener for **new** channels tracked in `newChannels`. A failed **`reject()`** no longer leaves already-subscribed channels without the new callback.
> 
> Adds regression coverage for **concurrent subscribes** on one channel (both listeners after resolve) and **mixed subscribe + reject** (listener on `ch1` kept, `ch2` absent).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 31b76ff7599f006b74f730ee252d69789640021b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


nkaradzhov (approved):
Thanks — LGTM.

